PR #35, opened 2012-06-16
Use tabs, not spaces

fprochazka:
How did this even happen? You guys really think you represent the **majority** of the internet?


ionas:
PSR2 is really broken for in practice recommending a software feature (2 or 4 spaces soft-tabs) over a standard ASCII character (tabs, which can be set to whatever) as well as its inconsistency in regards to parentheses.

Both should be fixed in a PSR-2 update (like PSR2-Revised).

PR #168, opened 2013-08-28
turkish translation for PSR-0, PSR-1, PSR-2, PSR-3, voting protocol and psr naming conventions

hkulekci:
I will update translation day by day.


hkulekci:
Hi @philsturgeon 

I translated to Turkish language about two years ago. What do you think about that? Is there any problem about translation or structure of the documents? What can I do to be accepted? I want to continue with PSR-4 and updated parts of the PSR-0,1,2,3.

philsturgeon:
We accepted a few, then people decided it was a bit weird, so they stopped accepting new ones. 

Raise it on the mailing list to get more eyes on it.

PR #169, opened 2013-08-28
[DRAFT, WIP] PSR-5: PHPDoc Standard

mvriel:
This PR represents the current state of affairs with the PHPDoc Standard. Until this description and subject is altered this PR is considered a Work in Progress and not ready for a final review.

All feedback is welcome, please direct this to the PHP-FIG mailing list to prevent noise in this PR. This PSR covers so many subjects that unless the discussions are separated by topic that it will probably be quite impossible to follow in this PR.

I would also request that people open a thread per issue they would like to discuss and not combine several issues in a single mail; this will help to keep noise-levels down in discussions and for me to loook back or filter important information.

TODO items are tracked in the issue tracker for this PSR: https://github.com/phpDocumentor/fig-standards/issues


proposed/phpdoc.md
```diff
@@ -0,0 +1,2213 @@
...
+/**
+ * @license MIT
+ *
+ * @license GPL-2.0+
+ *
+ * @license http://www.spdx.org/licenses/MIT MIT License
+ */
+```
+
+### 8.10. @link [deprecated]
```
EspadaV8: This isn't listed as deprecated in the TOC

Mezzle:
Something that we've turned to this proposal to try and figure out, but haven't gained anthing from is, what to do when the docblock would exceed line length limitations.

```php
/**
 * setExistingFields
 *
 * @param \Doctrine\Common\Collections\Collection|\Protec\Phase4\Contract\ProductType\OutcomeFieldInterface[] $fields
 *
 * @return $this
 */
```

Where should this be split?

Mezzle:
Actually, reading a bit further, this should be

```php
/**
 * setExistingFields
 *
 * @param \Doctrine\Common\Collections\Collection<\Protec\Phase4\Contract\ProductType\OutcomeFieldInterface> $fields
 *
 * @return $this
 */

**Comment by cebe** [comment]
@Mezzle you can solve this by using `use`

```php
use Doctrine\Common\Collections\Collection;
use Protec\Phase4\Contract\ProductType\OutcomeFieldInterface;

...

/**
 * setExistingFields
 *
 * @param Collection<OutcomeFieldInterface> $fields
 *
 * @return $this
 */
```

Mezzle:
@cebe however, this PSR states

> It is RECOMMENDED for applications to expand any local name into a FQCN
> for easier processing and comparisons.

mvriel:
@Mezzle thanks for pointing that out, I have just removed it as it is a style recommendation and should not be in PSR-5. With PSR-5 we explicitly avoid formatting and coding style recommendations as this provides a definition of PHPDoc and not how to apply it.

milesj:
It's best to use the FQCN as it integrates easier with IDEs, like PHPStorm.

Mezzle:
@milesj I agree, however, I'm confused as to how best to split that line :)

milesj:
My only suggestion would be to span multiple lines, with anything occuring after the FQCN being placed on the next line.

```
/**
 * setExistingFields
 *
 * @param \Doctrine\Common\Collections\Collection<\Protec\Phase4\Contract\ProductType\OutcomeFieldInterface> 
 *        $fields This is a description
 *
 * @return $this
 */
```

proposed/phpdoc.md
```diff
@@ -0,0 +1,2215 @@
...
+
+## Appendix A. Types
+
+### ABNF
+
+A Type has the following [ABNF][RFC5234] definition:
+
+    type-expression  = type *("|" type)
+    type             = class-name / keyword / array
+    array            = (type / array-expression) "[]" / generic
```
gharlan: As `type` can be an array, this rule is recursive: e. g. `SomeClass[][]`.
Is this intended? If yes, it should be described in the "Arrays" section below.

EC-Joe:
Is there any thoughts on having a  [@mixin phpdoc](http://code4x.com/php/decorator-pattern-in-php-with-autocomplete-in-phpstorm/) (or similar). With PHPStorm IDE this is a hugely helpful addition and provides a means for avoiding adding (and maintaining) lots of `@method` docs to a class that leverages the `__call()` function for things like decorator patterns and such.

milesj:
+1 to @mixin.

KorvinSzanto:
@mixin would be invaluable to us at concrete5.

photodude:
:+1: to @mixin.

MeyerJL:
FYI it looks like `@mixin` is already [on the board](https://github.com/phpDocumentor/fig-standards/issues/35). I must have overlooked the link to the issues prior to my posting the suggestion.

proposed/phpdoc.md
```diff
@@ -0,0 +1,2214 @@
...
+an option array with 2 elements: 'required' and 'label'.
+
+```php
+/**
+ * Initializes this class with the given options.
+ *
+ * @param array $options {
+ *     @var bool   $required Whether this element is required
+ *     @var string $label    The display name for this element
+ * }
```
dantleech: How should a tuple be documented?

````php
return array('one', 'two', 3);
````

Comment (milesj) [inline proposed/phpdoc.md]: Well a tuple isn't a built in construct, so it shouldn't be.

Comment (dantleech) [inline proposed/phpdoc.md]: In that case I don't understand why an assocative array should have a representation in PHPDoc and not a tuple (i.e. a non-associative array), e.g.

````
/**
 * @param array $something {
 *   @var bool This is the first element in the tuple
 *   @var integer This is the second
 * }
 */
````

Comment (krzysiekpiasecki) [inline proposed/phpdoc.md]: +1 to @milesj

Comment (cebe) [inline proposed/phpdoc.md]: :+1:

File: proposed/phpdoc.md [hunk 4]
```diff
@@ -0,0 +1,2214 @@
...
+  - [8.5.  @deprecated](#85-deprecated)
+  - [8.6.  @example](#86-example)
+  - [8.7.  @global](#87-global)
+  - [8.8.  @internal](#88-internal)
+  - [8.9.  @license](#89-license)
+  - [8.10. @link [deprecated]](#810-link-deprecated)
+  - [8.11. @method](#811-method)
+  - [8.12. @package](#812-package)
+  - [8.13. @param](#813-param)
+  - [8.14. @property](#814-property)
```
olvlvl: What about `@property-read` and `@property-write`? Some of us are using these. Example:

```php
<?php

/*
 * @property-read mixed $a
 * @property-write mixed $b
 */
class A
{
    private $a;
    private $b;

    public function __get($property)
    {
        if ($property == 'a')
        {
            return $this->a;
        }
        else if ($property == 'b')
        {
            throw new ProppertyNotReadable($property);
        }

        throw new PropertyNotDefined($property);
    }

    public function __set($property, $value)
    {
        if ($property == 'a')
        {
            throw new ProppertyNotWritable($property);
        }
        else if ($property == 'b')
        {
            $this->b = $value;
        }

        throw new PropertyNotDefined($property);
    }
}
```

<https://github.com/ICanBoogie/Accessor>

cebe: I'd vote to add these too, there is a similar implementation in Yii framework: http://www.yiiframework.com/doc-2.0/guide-concept-properties.html https://github.com/yiisoft/yii2/blob/078b922114e5ba3481487a80f28769e69713ea33/framework/base/Object.php#L130-L163

Swader:
I would like to suggest adding `@prose` as per [this](https://github.com/ApiGen/ApiGen/issues/633).

Usage example:

```php
namespace foo;

use bar\baz as bam;

/**
* This class is used to do something.
*
* @prose intros/someclass_basics.rst
*/
class SomeClass {

    /**
    * @prose tuts/someMethod.md
    * @param array $params Configuration values
    */
    public function someMethod(array $params) {

    }
}
```

Alternatively `@prose` could be a URL or another annotation could be added: `@proseURL`. In that case it could either serve as a plain `@see`, or in certain documentation generators, it might pull in the content of `<article>` at that remote location, though this might be a bit too much.

proposed/phpdoc.md
mvriel: Those should be added and were omitted because of an error

mvriel:
Adding it as a new tag seems a bit much to me, mainly because it is more or less a specialisation for `@link` when purely considered from a reader's perspective. Perhaps it would better be served as a specialisation tag such as `@link:prose tuts/someMethod.md` although I must add that I would not recommend such usage.

A DocBlock's strength is that it sits close to the code that it documents and by referring to, for example, a tutorial you are possibly negating that strength since it is no longer purely about the method. This is a bit of a unit test versus integration test situation where the DocBlock itself is equal to a unit test in scope and a tutorial more of a integration test.

Also: by referencing an external location, especially a relative one, one will run a higher risk of referring to non-existing documentation

proposed/phpdoc.md
olvlvl: :+1:

proposed/phpdoc.md
dantleech: But I still do not understand why documenting an associative array is supported and a non-associative array is not.

proposed/phpdoc.md
```diff
@@ -0,0 +1,2214 @@
...
+## 4. Basic Principles
+
+* A PHPDoc MUST always be contained in a "DocComment"; the combination of these
+  two is called a "DocBlock".
+
+* A DocBlock MUST directly precede a "Structural Element"
+
+  > An exception to this principle is the File-level DocBlock which MUST be
+  > placed at the top of a PHP source code file as the first DocBlock in a 
+  > file.
```
shadowhand: Is this implying that *every single file in a project* must have a file-level docblock?

cebe: Maybe add an `, if present`:

> An exception to this principle is the File-level DocBlock which, if present, MUST be
> placed at the top of a PHP source code file as the first DocBlock in a file.

mlocati:
For non-static methods (implemented via `__call()`) we have
`@method [return type] [name]([type] [parameter], [...]) [description]`

What about static methods (implemented via `__callStatic()`)?
My suggestion:
`@method [static] [return type] [name]([type] [parameter], [...]) [description]`

If otherwise `static` may already be part of `return type`, what about explaining in the docs that `@method` is useful to describe methods implemented via `__call` __and via `__callStatic`__?

olvlvl:
@mlocati +1, I'm already using this with PhpStorm: https://github.com/ICanBoogie/ActiveRecord/blob/2.3/lib/Helpers.php#L17

Just a thought about @method, starting PHP7 the return type of a function may be defined as follows:

```php
function foo(): array {
    return [];
}
```

Maybe `@method` could follow this, e.g. `@method foo(): array`. What do you think?

mlocati:
> Maybe @method could follow this, e.g. @method foo(): array. What do you think?

@olvlvl I'd prefer to keep the current approach, so that all the `@<something>` follow the same rule (`type name`)...

olvlvl:
@mlocati Good point.

SpacePossum:
Is it possible to escape a tag so it can be part of another (or the description of something)?

Gert-dev:
Regarding docblock inheritance for methods: are docblocks inherited from trait methods? For example, if class `C` uses trait `T`, where `T` defines method `foo` (with docblock) and `C` also defines `foo` (without docblock), does the `foo` from `C` inherit the docblock from `T`? In a certain way, traits are code that is injected into the class, so in essence the original trait method "disappears" from the equation as soon as it's redefined in the class itself, so would docblocks not be inherited in this case?

Similarly, if method `foo` in trait `T` does not specify a docblock and `C` now extends parent class `P`, which also defines a `foo`, does the `foo` from the trait also inherit documentation from the `foo` in `P` (and hence, can it use things such as `{@inheritDoc}` for this purpose)? Using the same logic as above, that would also be the case.

cebe:
@Gert-dev yes, traits should be considered in the way you described imo.

Gert-dev:
@cebe Thanks, that seems to make the most sense. @mvriel Perhaps it would be interesting to take this up in the standard?

I have another item up for debate, after recently discussing this with a colleague. Take the following example:

```php
/**
 * @param string $bar
 */
function foo($bar) { ... }

foo(5);
```

In this example, I'm not passing a string as parameter. Is this in violation of the standard? Should I be explicitly casting `5` to a string? The documentation states that I should be passing a string and that is what the function's internals will be expecting, but I'm actually passing an integer. Due to PHP's type coercion, if I use the passed in number as a string in the function body, things will probably work out fine. The same applies here:

```php
declare(strict_types=0);

/**
 * @param string $bar
 */
function foo(string $bar) { ... }

foo(5); // Just fine, int will be coerced to string.
```

However, things are not so fine and dandy when I'm using strict typing:

```php
declare(strict_types=1);

/**
 * @param string $bar
 */
function foo(string $bar) { ... }

foo(5); // Oops!
```

For PHP < 7, I've always seen type declarations such as `@param` as a way to strictly specify the type of a parameter in a way that wasn't possible in the language itself. Since PHP 7, we do however have scalar type hinting for this and it naturally seems to complement PSR-5.

If PSR-5 takes the stance that loose coercion is simply allowed in the cases above and the first example is correct, does this then also apply to the following case?

```php
/**
 * @param string $bar
 */
function foo($bar) { ... }

foo(null);
```

I did not indicate that `$bar` is nullable, but `null` in a sense coerces to a `string` in the sense that you can use it in situations where strings are expected (such as concatenations), so is the `string|null` indication now obsolete? One could argue that no, that is not the case, since with PHP 7 that is also not the case:

```php
/**
 * @param string $bar
 */
function foo(string $bar) { ... }

foo(null); // Error in both strict and non-strict mode!
````

So this makes me think that docblocks should be strict about the type and explicitly state the types they allow, in this case `string|null`. But now we are in violation of our first conclusion: we were passing an `int` that loosely coerces to a string, which is wrong, since `int` is not an explicitly allowed type in our docblock, i.e. it says `string|null`, not `string|int|null` or even `mixed`.

This brings me to the final question: how strict are docblocks regarding typing? Are they interpreted in the same way in strict (PHP 7) and non-strict typing mode? If not, will we have to update all our docblocks if we transition from one mode to another in a file? Do they perhaps follow the strict typing mode of the file?

The most important reason for asking this question is with regards to tools and IDE's: If we leave this open for interpretation by tools doing static analysis for types, we will get varying results between them, and if we get varying results and the types are still open to interpretation, what good is specifying them at all?

**Comment by mlocati** [comment]
I think that we should adopt the most restrictive approach.

Think for instance at this example:
```php
/**
 * @param string $name
 */
function printName($name)
{
    if ($name === '') {
        echo 'You belong to the Anonymous team ;)';
    } else {
        echo "Your name is $name";
    }
}
```

michaelcullum:
This should be merged anyway as proposed PSRs live in the main repo.

PR #179, opened 2013-09-11
[PSR-0] Added translation for portuguese brazil

ghost:
Hello!
Here is a translation of the PSR-0, Please rate!


henriquemoody:
@philsturgeon, can you close this issue looks like PSR-0 was already translated to pt_BR

PR #253, opened 2014-02-04
[PSR-3] Fixed interpolate() to pass tests

dunglas:
The current implementation of the `interpolate()` provided in the PSR-3 RFC breaks tests of https://github.com/php-fig/log/blob/master/Psr/Log/Test/LoggerInterfaceTest.php and is not compliant with the RFC itself:
> Every method accepts an array as context data. This is meant to hold any extraneous information that does not fit well in a string. The array can contain anything. Implementors MUST ensure they treat context data with as much lenience as possible. A given value in the context MUST NOT throw an exception nor raise any php error, warning or notice.

This PR fixes that.

### DIFF
--- a/accepted/PSR-3-logger-interface.md
+++ b/accepted/PSR-3-logger-interface.md
@@ -73,7 +73,10 @@ Users of loggers are refered to as `user`.
       // build a replacement array with braces around the context keys
       $replace = array();
       foreach ($context as $key => $val) {
-          $replace['{' . $key . '}'] = $val;
+          // check that the value can be casted to string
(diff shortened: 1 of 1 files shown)


michaelcullum:
Sorry to resurrect this but I think this is a bug fix to the example. @Seldaek Do you want to post a ML topic asking for objections to this being merged/errata or shall we close this?

Seldaek:
If anyone cares enough to send a ML I'll happily support it, but at the moment I am swamped and I know even if I promised to do it I am not going to spend cycles on it. So I rather not promise anything :)

michaelcullum:
@dunglas?

PR #279, opened 2014-05-12
Add Affiliated Projects definition

Crell:
Define a class of projects that may be co-represented by a single Voting Representative.


michaelcullum:
What happened with this?

Crell:
The general response from the rest of the list was ¯\_(ツ)_/¯.  I expect it will bite us again later, at which point maybe we can bring it up again.

michaelcullum:
Closing for now

PR #282, opened 2014-05-20
LogLevel constant values do not comply with RFC 5424

sun:
Discussion: https://groups.google.com/forum/#!topic/php-fig/Rc5YDhNdGz4

Corresponding `Psr/Log` package change: php-fig/log#21

Origin: php-fig/log#8

---

The only precedent for Errata appears to be in the PSR-2 meta document (detached from the actual spec).  PSR-3 does not have a meta document.

### DIFF
--- a/accepted/PSR-3-logger-interface.md
+++ b/accepted/PSR-3-logger-interface.md
@@ -127,14 +127,17 @@ Users of loggers are refered to as `user`.
 - The `Psr\Log\LoggerAwareTrait` trait can be used to implement the equivalent
   interface easily in any class. It gives you access to `$this->logger`.
 
-- The `Psr\Log\LogLevel` class holds constants for the eight log levels.
+- The `Psr\Log\LogLevel` class holds constants for the eight [RFC 5424] log
+  message severity levels.
 
 2. Package
 ----------
 
 The interfaces and classes described as well as relevant exception classes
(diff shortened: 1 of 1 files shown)


texdc:
Assuming that the current implementation is intended to replicate an `enum`, I think what we're looking for is a value object similar to this:

```php
/**
 * @method static LogLevel EMERGENCY()
 * ...
 */
final class LogLevel
{
    private static $levels = [
        'emergency',
        // ...
    ];

    private $level;
    private $label;

    private function __construct(int $aLevel, string $aLabel)
    {
        $this->level = $aLevel;
        $this->label = $aLabel;
    }

    public static function __callStatic(string $aLevel, array $arguments = []) : LogLevel
    {
        $strVal = strtolower($aLevel);
        foreach (static::$levels as $level => $label) {
            if ($label === $strVal) {
                return new static($level, $label);
            }
        }
        throw new InvalidLogLevelException($aLevel);
    }

    public function asInt() : int
    {
        return $this->level;
    }

    public function __toString() : string
    {
        return $this->label;
    }
    
    public function equals($aLevel) : bool
    {
        if (is_int($aLevel)) {
            return $this->level === $aLevel;
        }
        return $this->label === strtolower($aLevel);
    }
}
```

Sorry for the necro post.

PR #328, opened 2014-09-19
[PSR-3] Fixed first codeblock

ChristianGaertner:
The indentation caused an overwrite of the php syntax styling and just used the plain highlight.

This allows follows the convention used in the rest of the document.

Cheers,

Christian


michaelcullum:
This appears to have been fixed in GH's rendering engine so I see no need to change it now (unless I'm mistaken)

PR #341, opened 2014-10-08
Deprecate PSR-0

philsturgeon:
This is going to a vote on the mailing list.

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -82,12 +82,12 @@ message can be very small or extremely large. Attempting to represent the body
 of a message as a string can easily consume more memory than intended because
 the body must be stored completely in memory. Attempting to store the body of a
 request or response in memory would preclude the use of that implementation from
-being able to work with large message bodies. The `StreamInterface` is used in
-order to hide the implementation details of where a stream of data is read from
+being able to work with large message bodies. The `StreamableInterface` is used in
+order to hide the implementation details when a stream of data is read from
 or written to.
 
-`StreamInterface` exposes several methods that enable streams to be read
-  from, written to, and traversed effectively.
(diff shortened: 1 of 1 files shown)


jslegers:
WHY? WHY, I ask you?

I love PSR-0! The ability to use the `_` in class names as an equivalent to directory separators in paths, PSR-0 allows you to create multiple libraries for the same namespace that can reference each other without having to use those nasty `use ...` statements or having to reference the full namespace.

By removing the special meaning of the `_`  character, PSR-4 both removes that very useful ability forces us to support two different standards if we still want to use older libraries in our projects. That adds needless complexity and loading time to zero-config autoloaders.

I've always considered PSR-4 a very bad standard, but deprecation of PSR-0 is even worse, as it literally FORCES me to wrote needlessly convoluted code.

See also https://github.com/php-fig/fig-standards/issues/448.

philsturgeon:
Because having two standards doing the same thing is moronic. I would have deprecated PSR-0 the second PSR-4 came out if we'd had a deprecation process at the time.

The reasons for why have been discussed heavily on the mailing list, so you should go there if you're looking for discussion. I also [wrote a blog](https://philsturgeon.uk/php/2014/07/19/send-psr0-to-the-standards-farm-in-the-sky/).

philsturgeon:
It also sounds like your problem is more with actual namespaces than PSR-4. If you're not a fan of use statements then at least they get a little [easier in PHP 7](https://wiki.php.net/rfc/group_use_declarations).

Either way, PSR-0 was old and broken and you should be happy with PSR-4 as namespaces are how this stuff is meant to work, instead of just hacking around with poor mans namespaces to avoid having to use `use` statements for their intended purpose.

PR #361, opened 2014-10-23
[PSR-6] Renamed "pool" to "cache" for better usability, made naming consistent

webmozart:
Hi there!

Thank you @tedivm for explaining the cache approaches on your blog. I like the proposal.

As an outsider of the cache working group, I would like to bring in some unbiased feedback:

1. The concept of a cache "pool" is great for discussion of various approaches. The end users of the PSR, however, don't need to know what a "pool" is. For them, it is IMO sufficient and easier if the "pool" is simply called "cache".

2. The item has `getKey()`, but `get()`. This is a bit weird - IMO the code becomes easier to read if `get()` and `set()` are called `getValue()` and `setValue()`.

3. The pool has `getItem()` and `getItems()`, but `deleteItems()` only. I added `deleteItem()` for the sake of completeness.

I have some more suggestions which require more discussion however. I'll submit them in a separate PR.

Thanks for your hard work! :)

Cheers,
Bernhard


Crell:
The naming here was discussed at obscene length in the past.  The idea of keeping the "pool" in the name was to make it clear that we were talking about a specific pool, bin, etc, not about the cache driver itself. That was a point of confusion before.

(I personally preferred "bin" to "pool", but the consensus on list was to go with pool.)

dragoonis:
I am happy with the term Pool. If we thought about changing it to Cache I would have no objections.

PR #364, opened 2014-10-24
Add PSR-1 and PSR-4 Chinese Simplified language.

tangrucheng:
Add PSR-1 and PSR-4 Chinese Simplified language.

### DIFF
--- a/accepted/zh_CN/PSR-0.md
+++ b/accepted/zh_CN/PSR-0.md
@@ -1,0 +1,59 @@
+下文描述了若要使用一个通用的`自动加载器(autoloader)`，你所需要遵守的规范：
+
+规范
+---------
+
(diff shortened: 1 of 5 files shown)


gaomd:
Finally, we have an Chinese translation (and was the first of all Chinese translation PRs) #105 merged. The overall quality is acceptable except some formatting conventions and typos. we could improve them later however.

The bright side is that we have a base to work on, I think it's a best option to the current chaos situation of Chinese translations.

As for this PR, I see no reason to use vanilla MUST, SHOULD, etc words in place of translations. And after comparing to #105 (which was merged 9 days ago), **looks like this PR was based on #105**, with some words refining and typo fixes, but this PR changed all RFC 2119 words back to English.

I suggest we focus on typo fixes and article formatting conventions for the merged zh_CN translation for now.

@samdark I suggest this PR to be closed (at least this PR does not respect copyright, according to the pull request date and commit message, correct me if I was wrong).

PR #382, opened 2014-11-20
[PSR-7] Rename getHeaderAsArray() to getHeaderLines()

weierophinney:
Per discussion on the mailing list, `getHeaderAsArray()` was easily confused as
retrieving all headers; additionally, using the "as array" verbiage smacks of
hungarian notation.

This patch renames the method to `getHeaderLines($header)`, clarifying intent.

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -85,7 +85,7 @@ convenience of working with headers as strings, headers can be retrieved from
 an instance of a ``MessageInterface`` as an array or string. Use the
 `getHeader()` method to retrieve a header value as a string containing all
 header values of a header by name concatenated with a comma.
-Use `getHeaderAsArray()` to retrieve an array of all the header values for a
+Use `getHeaderLines()` to retrieve an array of all the header values for a
 particular header by name.
 
 ```php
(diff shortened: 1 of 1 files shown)


weierophinney:
@dragoonis This one is ready for merge.

proposed/http-message.md
```diff
@@ -85,7 +85,7 @@ convenience of working with headers as strings, headers can be retrieved from
 an instance of a ``MessageInterface`` as an array or string. Use the
 `getHeader()` method to retrieve a header value as a string containing all
 header values of a header by name concatenated with a comma.
-Use `getHeaderAsArray()` to retrieve an array of all the header values for a
+Use `getHeaderLines()` to retrieve an array of all the header values for a
```
samdark: Was `getHeaderValues` already proposed?

weierophinney: Yes, and rejected as it could be easily confused with `getHeaders()`.

samdark: I don't like `getHeaderLines` since it's about getting multiple lines of the same header (probably a single one?)

weierophinney: The signature is `getHeaderLines($headerName)`; it should raise an exception if no header name is presented.

Essentially, I'm not sure there's a good name here. We're talking about getting plural lines/values for a single header, versus a set of headers. There is no name that will be completely disambiguous unless we choose to return to using hungarian notation, which also sucks.

samdark: Yeah, that's quite hard one. More alternatives:

- `getValuesForHeader()`
- `getValuesOfHeader()`

evert:
Could we get this one merged? :)

weierophinney:
Ping @dragoonis ... Ready to merge.

weierophinney:
Merged with #405

PR #383, opened 2014-11-20
[PSR-7] Allow mutable query string, body, and cookie params

weierophinney:
Following discussion and research, this pull request makes query string
arguments, body parameters, and cookie parameters mutable.

The rationale is as follows:

- Making them immutable, when demonstrated use cases for mutability exist, will
  lead to a) lots of boilerplate code when mutability is needed (best case
  scenario), and/or b) refusal to use the interfaces due to lack of flexibility
  (worst case scenario).
- PHP already treats the equivalent superglobals (`$_GET`, `$_POST`, and
  `$_COOKIE`) as mutable.
- Other similar HTTP implementations in other languages (e.g., Python WSGI, Ruby
  Rack, Node's `http.IncomingMessage`) treat these as mutable.

Note: server and file params are NOT being made mutable at this time. These are
products of the server that cannot be calculated from any other content, and
thus should remain immutable.

(Query string params can be recalculated from the `QUERY_STRING` sever param;
body parameters can be recalculated from the message body; cookie parameters can
be recalculated from the `Cookie` header(s), meaning you can always get back to
the original values if desired.)

### DIFF
--- a/proposed/http-message-meta.md
+++ b/proposed/http-message-meta.md
@@ -126,15 +126,16 @@ can lead to messages entering into an invalid or inconsistent state.
 
 #### Mutability of messages
 
-The proposal models "context-specific" mutability. This means that a message is mutable based on its context. For example:
+The proposal models "context-specific" mutability. This means that a message is
+mutable based on its context. For example:
 
 - Client-side requests are mutable, as they will be iterably built before being
   sent.
 - Client-side responses are immutable, as they represent the result of a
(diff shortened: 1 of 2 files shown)


weierophinney:
Closing; will be reverting to the originally proposed `MessageInterface`/`RequestInterface`/`ResponseInterface`/`ServerRequestInterface`; the changes listed in this PR will be incorporated in the new PR.

PR #384, opened 2014-11-20
[PSR-7] URL clarifications

weierophinney:
Per discussion on the mailing list and further research, this patch introduces the following changes:

- The request URL inside an `OutgoingRequestInterface` MUST be an absolute URL.
- The request URL inside an `IncomingRequestInterface` MUST be an origin-form URL (path info + query string).

See the changes listed in the meta document for rationale.


weierophinney:
Closing. Per mailing list discussion, I'm reverting to the original interface structure, and will be addressing the URI issues in those changes.

PR #393, opened 2014-12-23
[PSR-9][WIP] split off PSR-10

lsmith77:


proposed/security-disclosure.md
```diff
@@ -22,3 +22,99 @@ interpreted as described in [RFC 2119][].
...
+for different major version number. In this case one URL SHOULD be provided
+for each major version. In the case a major version is no longer receiving
+security fixes, instead of an URL a project MAY opt to instead simply
+note that the version is no longer receiving security fixes.
+
+## Security Disclosure Process
+
+Every project MUST provide an email address in their security disclosure
+process description. If not specified otherwise, this email address is
+``security@[project domain]``. Projects SHALL NOT use contact forms.
```
mbeccati: "is exepcted to be"?

proposed/security-disclosure.md
```diff
@@ -22,3 +22,99 @@ interpreted as described in [RFC 2119][].
...
+
+* name (name of the product)
+* cve (unique CVE ID)
+* cwe (unique CWE ID)
+* severity (low, medium high)
+* affected (version(s) using composer syntax [4])
+* status (open, in progress, disputed, completed)
+* remediation (textual description for how to fix an affected system)
+* remediationType (workaround, mitigation, vendor fix, none available, will not fix)
+* remediationLink (URL to give additional information for remediation)
```
mbeccati: a vulnerability might have multiple remediations, e.g.
* upgrade to the latest version
* delete a certain file (most of the times it is xmlrpc.php ;) ) if the above is not possible

proposed/security-disclosure-meta.md
```diff
@@ -65,6 +65,29 @@ outside authority for something as sensitive as security related topics.
...
+* http://www.yiiframework.com/security/
+* https://www.drupal.org/security
+* http://codex.wordpress.org/FAQ_Security
+* http://www.sugarcrm.com/page/sugarcrm-security-policy/en
+* http://typo3.org/teams/security/
+* http://cakephp.org/development
+* http://www.concrete5.org/developers/security/
+* http://developer.joomla.org/security.html
+* http://wiki.horde.org/SecurityManagement
+* http://www.revive-adserver.com/support/bugs/
```
mbeccati: http://www.revive-adserver.com/security/ is the main one, although we also repeat the same text in the bugs page.

lsmith77: also feel free to send a PR to my fork for obvious changes.

proposed/security-disclosure.md
quantumpacket: As long as the contact form is hosted on a separate system than the one the disclosure pertains to, I don't see why we would not allow it. That way the form can at least be protected via HTTPS so any sensitive details about the disclosure are protected (for the most part). Where as an email address has no way to ensure safe transmission. What was the reason behind forbidding contact forms and enforcing email use?

mbeccati: I don't know... We receive plenty of unwanted support request to the 
security@ email address, just because it's mentioned on the website. 
Even if its purpose is clearly written.

A contact form is going to generate even more noise and scare away 
security researchers who most likely prefer to contact the authors 
directly, possibly using PGP/GPG signed emails.

quantumpacket: Should the PSR recommend that a public key be posted as well then?

lsmith77: Good point about the public key, we can make this a SHOULD .. can you send a PR?

As for the forms, this wasn't discussed much on the list and I just added it. The main point was that its harder for the security reporter to be assured what platform is being used for the contact form and if it is therefore also compromised. Also even if the contact form uses a different set of scripts, its still possible that it is infected with the same (say for example there is a vulnerability in a popular email lib) or another vulnerability. As such it just seems easier to require email ..

PR #400, opened 2015-01-04
[PSR-7] Revert to original(-ish) interfaces

### LINKED ISSUE #296: [HTTP Message] What should Request::getUrl() return when no Host header is present?
This is the current documentation for `Request::getUrl()`:

    /**
     * Gets the absolute request URL.
     *
     * @return string Returns the URL as a string.
     */
    public function getUrl();

The problem is, to construct a full URL we need the contents of the `Host` header, which although mandatory in HTTP 1.1, is optional in HTTP 1.0.

So the implementer is left in the dark about what to do when no `Host` is present, such as when parsing the following valid request:

    GET /resource HTTP/1.0

I can see a few options:

- Throw an exception
- Return an invalid URL with no host: `http:///resource`
- Return a valid URL with a default host: `http://localhost/resource`

I have no personal preference, but I think it's important to document what to do in that case!

### LINKED ISSUE #394: [PSR-7] HTTP message interfaces
Would it be easier to understand if the interfaces were called simply

```
ClientRequestInterface
ClientResponseInterface
ServerRequestInterface
ServerResponseInterface
```

It's confusing to try and figure out which is which without reading the information first.

weierophinney:
This patch removes the client/server interface segregation, and re-instates full
mutability (with two notable exceptions). The rationale was discussed on the
mailing list; essentially, in practice, mutability will be simpler for consumers
and allow more varied architectures.

The interfaces presented now are:

- `StreamableInterface`
- `MessageInterface`
- `RequestInterface`
- `ServerRequestInterface` (was `IncomingRequestInterface`)
- `ResponseInterface`

The request interface also now has an "absolute URI" member. The "URL" member is only
used to store the path and query string, while the absolute URI is intended to store the full URI of the request - which is minimally scheme and host. If either scheme or host are not present, the getter will return null, and the setter will raise an exception.

Fixes #296
Fixes #394

#### Updates

- 2015-01-07: "base URL" was changed to "absolute URI", per discussions on #296 and in the mailing list.

### DIFF
--- a/proposed/http-message-meta.md
+++ b/proposed/http-message-meta.md
@@ -5,10 +5,11 @@ HTTP Message Meta Document
 ----------
 
 The purpose of this proposal is to provide a set of common interfaces for HTTP
-messages as described in [RFC 7230] and [RFC 7231].
+messages as described in [RFC 7230](http://tools.ietf.org/html/rfc7230) and
+[RFC 7231](http://tools.ietf.org/html/rfc7231).
 
-[RFC 7230]: http://www.ietf.org/rfc/rfc7230.txt
-[RFC 7231]: http://www.ietf.org/rfc/rfc7231.txt
+- RFC 7230: http://www.ietf.org/rfc/rfc7230.txt
+- RFC 7231: http://www.ietf.org/rfc/rfc7231.txt
 
 All HTTP messages consist of the HTTP protocol version being used, headers, and
 a message body. A _Request_ builds on the message to include the HTTP method
@@ -77,6 +78,7 @@ messages.
 
 * Provide the interfaces needed for describing HTTP messages.
 * Keep the interfaces as minimal as possible.
+* Focus on practical applications and usability.
 * Ensure that the API does not impose arbitrary limits on HTTP messages. For
   example, some HTTP message bodies can be too large to store in memory, so we
   must account for this.
@@ -109,7 +111,10 @@ messages, whether they are for requests or responses. These elements include:
 More specific interfaces are used to describe requests and responses, and more
 specifically the context of each (client- vs. server-side). These divisions are
 partly inspired by existing PHP usage, but also by other languages such as
-Ruby, Python, Go, Node, etc.
+Ruby's [Rack](https://rack.github.io),
+Python's [WSGI](https://www.python.org/dev/peps/pep-0333/),
+Go's [http package](http://golang.org/pkg/net/http/),
+Node's [http module](http://nodejs.org/api/http.html), etc.
 
 #### Why are there header methods on messages rather than in a header bag?
 
@@ -126,18 +131,7 @@ can lead to messages entering into an invalid or inconsistent state.
 
 #### Mutability of messages
 
-The proposal models "context-specific" mutability. This means that a message is mutable based on its context. For example:
-
-- Client-side requests are mutable, as they will be iterably built before being
-  sent.
-- Client-side responses are immutable, as they represent the result of a
-  request that has been made; changing the result would lead to inconsistent
-  state when multiple processes are passed the message.
-- Server-side requests are immutable, as they represent the state of the
-  current incoming request; changes to the message would lead to inconsistent
-  state when multiple processes are passed the message.
-- Server-side responses are mutable, as they are built iterably before being
-  returned to the client.
+The proposal models mutable messages.
 
 Real-world usage in clients requires mutable requests. As an example, most HTTP
 clients allow you to modify a request pre-flight in order to implement custom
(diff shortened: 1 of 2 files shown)


proposed/http-message-meta.md
```diff
@@ -156,34 +150,58 @@ This is not just a popular pattern in the PHP community:
...
+- PHP itself models the `$_GET`, `$_POST`, and `$_COOKIE` superglobals as mutable.
+- `$_POST` only models form-encoded data sent via POST. While this is arguably
+  still the most common scenario for web-submitted data, it fails to address the
+  very common needs of APIs, which may use methods such as PUT, PATCH, and DELETE
+  to submit data, and which are likely submitting JSON or XML. Since content type
+  can only be determined by inspecting the request itself, deserialization of
+  such body content can only happen within the application layer - suggesting
+  that body content parameters must be mutable.
+- Cookie encryption, while an arguable practice, is very common in modern
+  frameworks. For the practice to work, the cookie data must be mutable.
```
evert: * Streams, per definition, cannot be immutable.

(minor suggestion)

proposed/http-message.md
```diff
@@ -232,15 +236,177 @@ interface MessageInterface
...
+    /**
+     * Sets the body of the message.
+     *
+     * The body MUST be a StreamableInterface object.
+     *
+     * @param StreamableInterface $body Body.
+     * @return void
+     * @throws \InvalidArgumentException When the body is not valid.
+     */
+    public function setBody(StreamableInterface $body);
```
evert: If there is a case where getBody() may return `null`, should setBody() also allow `null` to be set?

proposed/http-message.md
```diff
@@ -332,9 +505,11 @@ interface IncomingRequestInterface extends MessageInterface
      * purposes of how duplicate query parameters are handled, and how nested
      * sets are handled.
      *
-     * @return array
+     * @param array $query Array of query string arguments, typically from
+     *     $_GET.
+     * @return void
      */
-    public function getQueryParams();
+    public function setQueryParams(array $query);
```
evert: This one is new and wasn't in a previous iteration of the full mutable spec. If query parameters gets a setter, I think it should be documented how this affects the result of getUri(), and whether setUri also updates the query params.

evert:
There's some issues I have with the state of this PR, but I think it would be good to merge regardless. The issues are minor and can be addressed later.

Hywan:
@evert :+1:

proposed/http-message-meta.md
weierophinney: Updated to add this.

proposed/http-message.md
weierophinney: I'm not sure we should allow `null`, actually, as it makes the return variable; I'm going to remove that as a valid return value from the getter.

evert: That's an even better solution!

proposed/http-message.md
weierophinney: We've had a few revisions. In the original `IncomingRequestInterface` implementation, it was mutable. Then we went with the "selective mutability", and it wasn't. Then I made an argument for essentially mutability of all input sources other than `$_SERVER` and `$_FILES`, with the assumption that you can rebuild the original state of the input sources from those and the content body.

Interestingly, `$_GET` is mutable. My assertion is that query parameters should reflect what PHP does here -- and thus also be mutable.

Altering `$_GET` does not alter either `$_SERVER['REQUEST_URI']` or `$_SERVER['QUERY_STRING']`, so my take is that changes to the query parameters in the request object should not update the request URI either. I'll document that.

evert: I think in real use-cases this is a bit problematic though.

Think about this: what if this is used in the context of a client. I set setQueryParams to pass a few variables. Would it alter the URI that the client ends up sending?

Another example, a middleware wants know whether a certain query parameter was set. Should it be forced to check both the uri and the query parameters?

evert: For what it's worth, I hold the opinion that 'query parameters' is a convenience that's not needed at all in the base spec, because the same information can be extracted from getUri().

If I can't have that, I would suggest to only provide a getter that *always* reflects what's in the uri.

If a setter is a must, I think the result of getUri() and getQueryParams() *should* always be consistent.

weierophinney: The problem with omitting query parameters from the interface is that we'll get people asking, "How do I get at `$_GET['foo']`?" They want to be able to get at that information without parsing. As such, we need access to it.

While I understand your point of view regarding having query parameters consistent with the URL, I also know that (a) as noted, PHP doesn't do this currently anyways, and (b) most times that you manipulate `$_GET`, you're not concerned about the implications to the URI; you're just stuffing stuff in there such as defaults that you'll want to access later.

Finally, it's a consistency issue: if body and cookie parameters are mutable, query parameters should be as well. At the same time, we're not requiring (and, I argue should not!) that changing these parameters would change the relevant sources (body stream and cookie header, respectively). The point is that the setters are there primarily to allow pre/post processing of the input sources.

evert: The body parameter is a bit different though, because in cases where `$_POST` is populated, the body tends to be empty. I actually felt that the cookie-related stuff should never have made it in there, best to just let those people parse it out :)

One thing I didn't realize though, is that QueryParams is only a part of the ServerRequest, and not the generic request. That makes this not as bad, and more logical imho so I'll happily drop this.

weierophinney: Correct -- this all part of that "Logical/Usage" hybrid @Crell posted on the ML about; it's only in the `ServerRequest`. I agree -- if this had been part of the regular request interface, I'd want 2-way binding!

One last point: an implementation can always choose immutability. That's one reason I think mutability is important here -- it allows the most common scenarios, and those who want immutability can always have the setters throw exceptions in their implementations or decorators. :)

weierophinney:
@evert Note that I've now incorporated the ideas from #296 into this PR (i.e., changed `BaseUrl` to `AbsoluteUri` and documented that it needs to be in sync with the value in `Url`).

proposed/http-message.md
```diff
@@ -232,15 +236,184 @@ interface MessageInterface
...
+     *
+     * When setting the URL, the absolute URI (see getAbsoluteUri() and
+     * setAbsoluteUri()) MUST be updated.
+     *
+     * @link http://tools.ietf.org/html/rfc7230#section-5.3
+     * @param string $url Request URL, with path and optionally query string.
+     * @return void
+     * @throws \InvalidArgumentException If the URL is invalid.
+     */
+    public function setUrl($url);
```
mtdowling: I can see this method name being confusing to users: it's called `setUrl()`, but it expects that the provided value be something like "/foo/bar?baz". Have you considered calling this `setRequestTarget()` like what's used in the RFC? Or maybe `setUriOriginForm()`? It seems strange to me to also mix the terms URL and URI.

proposed/http-message.md
```diff
@@ -232,15 +236,184 @@ interface MessageInterface
...
+    /**
+     * Retrieves the request URL.
+     *
+     * The request URL is the path and query string ONLY.
+     *
+     * @link http://tools.ietf.org/html/rfc7230#section-5.3
+     * @return string Returns the URL as a string. The URL MUST be an
+     *     origin-form (path + query string), per RFC 7230 section 5.3
+     */
+    public function getUrl();
```
mtdowling: Perhaps this method should be renamed? RFC 7230 does not use the term URL, so maybe we shouldn't either.

proposed/http-message.md
weierophinney: The problem is that "request target" has 4 potential forms, ranging from least information (origin-form) to most (fully qualified URI that includes all of scheme, authentication, host, port, path, and query).

The "absolute URI" member essentially address the "most" target, while the "URL" member addresses the least. @evert has made the case that URL is a well-known subset of URI, and is used to specify origin-form -- but, as you note, it could easily be confused with URI, leading to odd error conditions or expectation problems.

`setUriOriginForm()` would be a head-scratcher for anybody who has not read the spec -- which would be a majority of users. As such, I think it's not a good candidate.

Perhaps we should do absolute URI and path only? My only issue with that is if the path changes, the query string likely need to change as well, and then we're back to needing getters/setters for every URI part, or at least a plethora of them.

mtdowling: > The problem is that "request target" has 4 potential forms, ranging from least information (origin-form) to most (fully qualified URI that includes all of scheme, authentication, host, port, path, and query).

Yep. Allowing for the full range of request-target as described in section 5.3 would make this proposal support proxy URLs, OPTIONS requests, etc. What you might lose from supporting this is the ability to derive an absolute URI (where the request is going/from) from this data.

Maybe including absolute URIs isn't even needed in the PSR. Perhaps you could use just get/setRequestTarget(), and it would be what is sent over the wire, verbatim, in the start-line of a request. This would support "/foo/bar?baz", and proxy URLs (e.g., "http://xyz.com/foo/bar?baz"). Determining an absolute URI is really a sort of convenience mechanism for combining a bunch of parts of the URL together. The only thing that would be missing from constructing an absolute URI would be the scheme, but perhaps that's a concern of a client and not a message.

evert: > @evert has made the case that URL is a well-known subset of URI

That wasn't me! URL is deprecated, and URI is the current term. The spec should pick one of URL and URI and use it consistently.

weierophinney:
Closing in favor of #405.

PR #405, opened 2015-01-18
[PSR-7] Immutability + URI interface

weierophinney:
This pull request obsoletes #400 and provides the following:

- Messages are now immutable. See the [mailing list discussion](https://groups.google.com/d/msg/php-fig/9gK8vX8iYZ8/5PZ9rx8UvXYJ) for the why and how. Primarily this means that setters were changed to mutator methods using the verbiage `with` and `without`, and now return a **new** instance that contains the requested updates.
- Addition of a new interface, `UriTargetInterface`. This interface models the request-target of a request message, including those formats that represent URIs per RFC 3986. `RequestInterface` now defines `getUri()` and `withUri(UriTargetInterface $uri)` instead of the `(get|set)Url()` equivalents from previous drafts.
- `setAttributes()` was removed. Per a comment from @Crell, this method suffered conceptually in the same way that `setHeaders()` and `addHeaders()` did. A new method, `withoutAttribute()` was added, however, to allow removing an attribute entirely.

I also spent a lot of time making sure the various docblocks fully described the interfaces and methods and the stipulations on behavior. I made similar updates to the proposal and metadata documents -- and particularly the latter, to ensure that the various design decisions have backing arguments.

### DIFF
--- a/proposed/http-message-meta.md
+++ b/proposed/http-message-meta.md
@@ -6,10 +6,12 @@ HTTP Message Meta Document
 
 The purpose of this proposal is to provide a set of common interfaces for HTTP
 messages as described in [RFC 7230](http://tools.ietf.org/html/rfc7230) and
-[RFC 7231](http://tools.ietf.org/html/rfc7231).
+[RFC 7231](http://tools.ietf.org/html/rfc7231), and URIs as described in
+[RFC 3986](http://tools.ietf.org/html/rfc3986) (in the context of HTTP messages).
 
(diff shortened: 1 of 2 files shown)


weierophinney:
Ping @dragoonis -- please merge this and php-fig/http-message#16.

weierophinney:
Or maybe ping @pmjones to merge this and php-fig/http-message#16 ? I'm not sure who has rights to merge now that @philsturgeon has left...

PR #408, opened 2015-01-22
[PSR-7] Fix verbiage around authority/asterisk forms

weierophinney:
Per comments from @evert, I've updated the verbiage around the `isAuthority()` and `isAsterisk()` methods to remove references to "URI" and instead reference "request-target".


weierophinney:
Ping @pmjones

evert:
:+1:

PR #412, opened 2015-01-28
[PSR-7] Model URIs and move request-target details to the Request

weierophinney:
Per @evert on the mailing list, this patch accomplishes a few things:

- `UriTargetInterface` was renamed to `UriInterface`, and modified to
  remove the various references to request-targets (particularly the
  `is*()` methods). The URI will always be necessary for making
  requests, and is generally of use on server-side requests as well,
  regardless of its representation in the actual request message.
- `RequestInterface` was updated to reference `UriInterface` instead of
  `UriTargetInterface`.
- `RequestInterface` was updated to add the methods `getRequestLine()`
  and `withRequestLine($requestLine)`. The first will return the request
  line as calculated or as set with `withRequestLine()`; this allows
  users to specify the alternate request-target forms when desired,
  while keeping those details out of the URI implementation.
  Additionally, that change will ensure that clients always have the
  full URI, which will be necessary when establishing an actual
  connection, regardless of the request-target form.

### DIFF
--- a/proposed/http-message-meta.md
+++ b/proposed/http-message-meta.md
@@ -205,22 +205,28 @@ provides convenience in client applications by allowing users to create new
 instances of a base URI instance with just the segments that change (e.g.,
 updating the path only).
 
-### Why are URIs and request-targets merged in a single interface?
+### Why does the request interface have methods for dealing with the request line as whole, AND compose a URI?
 
 RFC 7230 details the request line as containing a "request-target". Of the four
-forms of request-target, only one is a URI compliant with RFC 3986. However,
-since all forms are valid for purposes of requests, the proposal must
-accommodate each.
+forms of request-target, only one is a URI compliant with RFC 3986; the most
+common form used, however, is origin-form, which represents the URI without the
+scheme or authority information. Moreover, since all forms are valid for
+purposes of requests, the proposal must accommodate each.
 
-The most common use case for requests, however, will be specifying either
-absolute URIs (absolute-form) or relative URIs (origin-form) -- which are also
-the most commonly known request-target types. To keep verbiage familiar for
-developers, `RequestInterface` only references the term "URI".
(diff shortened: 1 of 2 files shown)


weierophinney:
Ping @evert — please review, and, on approval, I'll have @pmjones merge.

weierophinney:
@pmjones You were too fast -- I was waiting for @evert to review! That said, it's almost verbatim from his suggestions on the mailing list; my guess is we'll have tweaks at best. :)

evert:
Hey @weierophinney , I was a bit tired yesterday.. and now I'm looking at the patch, I think I would make one change.

Change requestLine to requestTarget, and remove the method and http version.

Both the method and http version are already represented, so this would make it a bit simpler. I carelessly copied the 'RequestLine' terminology, without further thinking that it it's not the exact same thing as 'RequestTarget'.

I'm very sorry I didn't see this immediately :( I feel you're putting so much work in and I'm not contributing enough for the amount of comments that I have...

weierophinney:
@evert I'm actually happy with the RequestLine terminology (just finished implementing it in phly/http). If we keep it, I'm actually of the mind that we should have similar on the ResponseInterface. 

I can see your point, though. Mulling it over, and will let you know if I prepare another patch.

evert:
Yea, think about it :) Because there's not really anything extra you can do with "RequestLine" that you could not do with a combination of method + request-target + http-version, and it does increase redundancy a bit.

It also introduced this syntax:

    $request
       ->withRequestLine('OPTIONS / HTTP/1.1')
       ->withMethod('GET');

Which might make one wonder which method will ultimately be used.

weierophinney:
@evert Good point; wouldn't want those to go out-of-sync. Making a new PR now.

weierophinney:
@evert See #414.

PR #414, opened 2015-01-28
[PSR-7] Only work with the request target, not the line

weierophinney:
As @evert notes, by allowing the ability to specify the entire request
line, we add the potential for the request line to be out of sync with
the HTTP method and protocol version composed — which would represent
both inconsistent state and an invalid message.

This patch renames `*RequestLine()` to `*RequestTarget()`.


weierophinney:
Waiting for feedback from @evert.

evert:
Changes look great!

weierophinney:
@pmjones Please merge! :)

PR #415, opened 2015-01-28
Added notes about case sensitivity of headers

evert:
Also made some unrelated formatting tweaks.

Please review @weierophinney

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -1,4 +1,4 @@
-﻿HTTP message interfaces
+HTTP message interfaces
 =======================
 
 This document describes common interfaces for representing HTTP messages as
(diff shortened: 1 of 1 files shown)


proposed/http-message.md
```diff
@@ -70,7 +73,7 @@ Both `Psr\Http\Message\RequestInterface` and `Psr\Http\Message\ResponseInterface
 implemented directly, implementors SHOULD implement
 `Psr\Http\Message\RequestInterface` and `Psr\Http\Message\ResponseInterface`.
 
-From here forward, the namespace `Psr\Http\Message` will be omitted when
+From here forward, the namespace `Psr\Http\MessageInterface` will be omitted when
```
weierophinney: This is incorrect -- the namespace **is** `Psr\Http\Message`.

proposed/http-message.md
```diff
@@ -97,6 +100,14 @@ echo $message->getHeader('foo');
 // Outputs: baz
 ```
 
+Despite that headers may be retrieved case-insenstively, the original case
+MUST be preserved by the implementation, in particular when retrieved with
+`getHeaders()`.
+
+Non-conforming HTTP applications may depend on a certain case, so it is useful
+for a user to be able to dictate the case of the HTTP headers when creating a 
+request or response.
```
weierophinney: I'd argue that this information should also be added to the interface definition itself -- under `withHeader()`, `withAddedHeader()`, and `getHeaders()`.

proposed/http-message.md
evert: Misread this :S

proposed/http-message.md
evert: Will do

evert:
How does this look?

weierophinney:
:+1: !

@pmjones Please merge this when merging #414.

weierophinney:
@evert Can you also do a PR against php-fig/http-message with the interface changes, please?

evert:
Yup

PR #423, opened 2015-02-01
[PSR-7] Correct example in headers with multiple values

geekwright:
Switch example to use current method names


weierophinney:
@pmjones merge this one or #424. I don't care. They're both the same, and correct. :) When you merge one, close the other.

weierophinney:
@geekwright Sorry -- saw the one from Hari come through, but not this one!

PR #424, opened 2015-02-02
PSR7 : no more set and add methods

harikt:


harikt:
@weierophinney may be you could verify this and merge ?

weierophinney:
@pmjones Changes are good; please merge. :-)

xabbuh:
By the way, this is the same as #423.

harikt:
@xabbuh nice. I didn't find it. I will close the PR, for credits should be given for FIFO ;) .

PR #427, opened 2015-02-09
[PSR-6] Consistent CacheItemInterface & CacheItemPoolInterface usage

matthiasmullie:


AndrewCarterUK:
I've just submitted a very similar pull request: https://github.com/php-fig/fig-standards/pull/520

This request misses an additional error on line 271/273 and doesn't make any changes to similar mistakes in the cache-meta.md file. I would have commented these changes here but I've only just found this!

Crell:
Thanks!

PR #429, opened 2015-02-11
[PSR-7] Reference the "parsed body", not "body params"

weierophinney:
Body parameters *are* the parsed body.

Per conversations with @Crell, @ircmaxell, @pmjones, and several others, the body parameters are always the parsed body content; calling them out as such, and without the "parameters" verbiage, implies that parsing may create non-array values, which is exceptionally likely when working with API and/or non-form-encoded payloads.

Moreover, the content will always be specific to the incoming request, so supporting one method representing `$_POST` and another representing parsed content is redundant and potentially confusing.

Finally, considering the fact that the parsed body is an ambiguous domain, forcing an ubiquitous signature is an unnecessary and limiting constraint.

However, the interface *does* specify that IF the request method is POST, AND the `Content-Type` is `application/x-www-form-urlencoded`, THEN the parsed body MUST be equivalent to `$_POST`.

### DIFF
--- a/proposed/http-message-meta.md
+++ b/proposed/http-message-meta.md
@@ -394,10 +394,11 @@ incoming requests:
   via the `$_SERVER` superglobal; these are part of the PHP Server API (SAPI)).
 - Access to the query string arguments (usually encapsulated in PHP via the
   `$_GET` superglobal).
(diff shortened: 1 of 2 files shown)


weierophinney:
@pmjones Please merge, following the various discussions at #ssp15 . :)

PR #443, opened 2015-02-23
[PSR-7] Include multipart/form-data in parsed body specified behaviors

geekwright:
This matches the expressed intentions of "parameters submitted via HTTP POST."

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -717,8 +717,9 @@ interface ServerRequestInterface extends RequestInterface
     /**
      * Retrieve any parameters provided in the request body.
      *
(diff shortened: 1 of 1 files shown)


Crell:
Seems reasonable to me, and covers a legitimate gap in the original. @weierophinney, any objection?

weierophinney:
@philsturgeon @pmjones or @dragoonis -- This is a good merge; it doesn't affect the specification in terms of behavior currently, but tightens up a definition. (I'd forgotten this use case, even though I specifically addressed it in Apigility last year!)

Thanks, @geekwright !

PR #444, opened 2015-02-27
simplification of \Example\Psr4AutoloaderClass

AntonLee:


michaelcullum:
I'm not really sure I see a huge amount of benefit to these changes and as the spec is finalised, unless there is a good reason to change it, I'd rather leave it as-is.

PR #446, opened 2015-03-01
[PSR-7] Detail Host header sources

weierophinney:
Per [the mailing
list](https://groups.google.com/forum/#!msg/php-fig/yZGMFW_fWEk/uuZlhmtIN_8J),
this patch details that the Host header should be pulled from the composed
`UriInterface`'s host segment if:

- no value is present for the Host header, AND
- the composed URI has a host segment.

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -501,12 +501,30 @@ namespace Psr\Http\Message;
 interface RequestInterface extends MessageInterface
 {
     /**
+     * Extends MessageInterface::getHeaders() to provide request-specific
+     * behavior.
+     *
+     * Retrieves all message headers.
+     *
+     * This method acts exactly like MessageInterface::getHeaders(), with one
+     * behavioral change: if the Host header has not been previously set, the
+     * method MUST attempt to pull the host segment of the composed URI, if
+     * present.
+     *
+     * @see MessageInterface::getHeaders()
+     * @see UriInterface::getHost()
+     * @return array Returns an associative array of the message's headers. Each
+     *     key MUST be a header name, and each value MUST be an array of strings.
+     */
+    public function getHeaders();
+
+    /**
      * Extends MessageInterface::getHeader() to provide request-specific
      * behavior.
      *
      * This method acts exactly like MessageInterface::getHeader(), with
      * one behavioral change: if the Host header is requested, but has
-     * not been previously set, the method SHOULD attempt to pull the host
+     * not been previously set, the method MUST attempt to pull the host
      * segment of the composed URI, if present.
      *
      * @see MessageInterface::getHeader()
@@ -515,6 +533,24 @@ interface RequestInterface extends MessageInterface
      * @return string
      */
     public function getHeader($name);
+
+    /**
+     * Extends MessageInterface::getHeaderLines() to provide request-specific
+     * behavior.
+     *
+     * Retrieves a header by the given case-insensitive name as an array of strings.
+     *
+     * This method acts exactly like MessageInterface::getHeaderLines(), with
+     * one behavioral change: if the Host header is requested, but has
+     * not been previously set, the method MUST attempt to pull the host
+     * segment of the composed URI, if present.
+     *
+     * @see MessageInterface::getHeaderLines()
+     * @see UriInterface::getHost()
+     * @param string $name Case-insensitive header field name.
+     * @return string[]
+     */
+    public function getHeaderLines($name);
 
     /**
      * Retrieves the message's request target.
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -135,6 +135,17 @@ Note: Not all header values can be concatenated using a comma (e.g.,
 `MessageInterface`-based classes SHOULD rely on the `getHeaderLines()` method
 for retrieving such multi-valued headers.
 
+##### Host header
+
+In requests, the Host header typically mirrors the host segment of the URI, as
+well as the host used when establishing the TCP connection. However, the HTTP
+specification allows the Host header to differ from each of the two.
+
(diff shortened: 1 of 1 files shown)


proposed/http-message.md
```diff
@@ -490,6 +501,22 @@ namespace Psr\Http\Message;
 interface RequestInterface extends MessageInterface
 {
     /**
+     * Extends MessageInterface::getHeader() to provide request-specific
+     * behavior.
+     *
+     * This method acts exactly like MessageInterface::getHeader(), with
+     * one behavioral change: if the Host header is requested, but has
+     * not been previously set, the method SHOULD attempt to pull the host
+     * segment of the composed URI, if present.
```
xabbuh: "off" instead of "of"?

evert:
Great, :+1:

mtdowling:
LGTM

proposed/http-message.md
```diff
@@ -135,6 +135,17 @@ Note: Not all header values can be concatenated using a comma (e.g.,
...
 
+##### Host header
+
+In requests, the Host header typically mirrors the host segment of the URI, as
+well as the host used when establishing the TCP connection. However, the HTTP
+specification allows the Host header to differ from each of the two.
+
+The `RequestInterface` overrides the `MessageInterface::getHeader()` method to
+indicate that if no Host header is present, but a host segment is present in the
+composed `UriInterface`, the value from the URI should be used. If a Host header
```
Tobion: the value from the URI __must__ be used

proposed/http-message.md
```diff
@@ -490,6 +501,58 @@ namespace Psr\Http\Message;
...
 {
     /**
+     * Extends MessageInterface::getHeaders() to provide request-specific
+     * behavior.
+     *
+     * Retrieves all message headers.
+     *
+     * This method acts exactly like MessageInterface::getHeaders(), with one
+     * behavioral change: if the Host header has not been previously set, the
+     * method MUST attempt to pull the host segment of the composed URI, if
```
Tobion: I think we should also add that this host header SHOULD then be returned as first element of the headers array.

As specified https://tools.ietf.org/html/rfc7230#section-5.4

> Since the Host field-value is critical information for handling a
   request, a user agent SHOULD generate Host as the first header field
   following the request-line.

evert: Disagree, :-1: there's no significance to the order in the array.

It's up to a client to correctly produce the http request and put the host first.

Tobion:
I think we should indeed add this logic :+1:
because RFC 7230 even says

> A server MUST respond with a 400 (Bad Request) status code to any
   HTTP/1.1 request message that lacks a Host header field

So when using RequestInterface for making a request as a client, this seems necessary.

evert:
@Tobion it's possible to make HTTP/1.0 requests and responses which don't require it.

Tobion:
Then why do you think we should dictate that behavior?

Also just returning pulling getUri()->getHost() does not seem enough. From what I can see, the `Host` header should also contain the port (if it's non-standard).

evert:
@Tobion this is actually a good point, I think you're right. But instead of checking if it's non-standard or standard (it can be complicated with http/https) it should just *always* use the port from the URI object if it's set, and don't if it's empty.

Tobion:
So to me, all this logic is task of the code that actually transmits the requests, e.g. using phps `header()` function. This class would need to implement exactly that (and usually only for HTTP/1.1 requests).

weierophinney:
So, to sum up, the only change needed is:

- indicate that the port should be included in the Host header if it is non-standard for the scheme

Does this sound correct? Pinging @evert and @Tobion

evert:
@weierophinney , no, only include the port if the port has been set, but don't make any assumptions about it being standard/non-standard. So if getPort() returns 80, add `:80`.

Normalization is a task for outside of this interface. In most normal cases getPort() should not return anything.

evert:
I'd be happy to make the change for you as well, but tomorrow at the earliest.

Tobion:
So I would suggest a different appraoch:

When a new URI is set, the Host header gets overwritten with the new value (host:port) from the Uri unless there was no host header before.
- This allows HTTP/1.0 without Host header
- And it keeps the Host header in sync with the Uri. "Most users will just want to set setUri() and you don't want to ask hem to explicitly also have to set the Host header every time. "
- Having a different Host header than uri->host would only be possible when explicitly setting the host header.

evert:
This approach was actually discussed in detail on the mailing list and in irc.

We've decided that we were going to make the behavior of the Host header consistent with the request target. That is, use the information from the URI object _unless_ it's explicitly overridden. This covers 99% of use-cases.

We decided to not care about the HTTP/1.0 case. Even though HTTP/1.0 does not require it, it's not harmful either. The situation where someone wants to explicitly test a HTTP/1.0 without a host header is the only real use-case for it (as far as I can deduce). This use-case is a bit too small to make add this additional behavior for. HTTP/1.0 is rare enough.

I think we reached concensus on this behavior, but if not... go to the mailing list for the discussion and pick it up there. Not all the interested parties see github issues.

Tobion:
I've read the discussion on the mailing list and even quoted parts of it.

As I said above, I'm convinced now that we should not add this magic at all. There is nothing we need to do here.
It's the task of the transmitting code to send a Host header based on getUri(), if there was no Host header given and HTTP/1.1 is used.
The current spec just adds magic and slowness (because you have to do it on every getHeader() call) and only forbids use-cases without actually solving anything.

proposed/http-message.md
Tobion: Then it's also up to the client to add the Host header in the first place.

evert:
@Tobion if you disagree, please argue on the list. This is not the appropriate place.

PR #449, opened 2015-03-03
[PSR-7] Added language to withPath() regarding percent encoding

weierophinney:
Per RFC 3986, section2, reserved characters MUST be percent encoded; however,
the implementation MUST NOT double-encode.

See the relevant discussion on the mailing list for more details:

- https://groups.google.com/d/msgid/php-fig/d5102d77-dab0-48cd-8e79-2999128ed948%40googlegroups.com

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -1287,6 +1287,10 @@ interface UriInterface
      * Additionally, the query string SHOULD be parseable by parse_str() in
      * order to be valid.
      *
+     * The implementation MUST percent encode reserved characters as
+     * specified in RFC 3986, Section 2, but MUST NOT double-encode any
+     * characters.
+     *
      * An empty query string value is equivalent to removing the query string.
      *
      * @param string $query The query string to use with the new instance.
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -1265,7 +1265,7 @@ interface UriInterface
      * The path MUST be prefixed with "/"; if not, the implementation MAY
      * provide the prefix itself.
      *
-     * The implementation MUST percent encode reserved characters as
+     * The implementation MUST percent-encode reserved characters as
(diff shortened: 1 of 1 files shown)


evert:
:+1:

mtdowling:
Looks good, but can this same blurb be used in the `withQuery` method as well?

weierophinney:
@mtdowling absolutely; got caught up on the `withPath()` detail, and spaced out on that part. Will update in a minute.

mtdowling:
LGTM

proposed/http-message.md
```diff
@@ -1265,6 +1265,10 @@ interface UriInterface
      * The path MUST be prefixed with "/"; if not, the implementation MAY
      * provide the prefix itself.
      *
+     * The implementation MUST percent encode reserved characters as
```
Tobion: "percent-encode" according to https://tools.ietf.org/html/rfc3986#section-2.1

proposed/http-message.md
```diff
@@ -1265,6 +1265,10 @@ interface UriInterface
      * The path MUST be prefixed with "/"; if not, the implementation MAY
      * provide the prefix itself.
      *
+     * The implementation MUST percent-encode reserved characters as
```
Tobion: - getPath should say that reserved characters MUST be percent-encoded
- withPath should only say that it MUST NOT double-encode characters.

This is because withPath must not encode immediately. That's an implementation detail.

Tobion:
Actually in it's current form this is wrong. If we say `withPath` encodes reserved characters, then all these chars would be encoded

> reserved    = gen-delims / sub-delims
      gen-delims  = ":" / "/" / "?" / "#" / "[" / "]" / "@"
      sub-delims  = "!" / "$" / "&" / "'" / "(" / ")"
                  / "*" / "+" / "," / ";" / "="

But this is of course wrong because we don't want to encode them, e.g. `/`.

PR #450, opened 2015-03-03
 Document return values for non-existant headers

evert:
Allow `getHeader()` to return `null` if the header did not exist.

@weierophinney does this make sense? I noticed that phly/http returns an empty string, but there's a difference between an empty header, and no header at all.


weierophinney:
:+1: 

@pmjones Let's merge this one as well. 

@evert Can you make the same PR against the http-message repo, please?

PR #451, opened 2015-03-04
[PSR-7] Use the term URI instead of URL consistently in the phpdoc

Tobion:
1. one incorrect casing of URL
2. in the phpdoc we should use URI since it's also UriInterface and getUri().


simensen:
:+1:

weierophinney:
:+1: Pinging @pmjones and/or @philsturgeon ...

Also, @Tobion — can you please submit the related changes on the php-fig/http-message repo?

PR #456, opened 2015-03-08
change PSR-7 status to "R" at Numerical Index section

sasezaki:


weierophinney:
The plan is to go to voting this week, so we may actually want to make that "V". :)

sasezaki:
ok, this PR is no longer correctly from yesterday. :-)

PR #459, opened 2015-03-18
[PSR-7] Use `#`-prefixed headers instead of separator-style

### LINKED ISSUE #453: PSR-7: Markdown rendering issue
Hi, the [References section of PSR-7](https://github.com/php-fig/fig-standards/blob/master/proposed/http-message.md#references) got GitHub's Markdown rendering engine confused, notice these 4 unordered list items were being separated into two groups: 3 items a group then left a single item. It's possible due to next item starts with `1.`.

screen-shot-2015-03-06-at-11 23 03

I've just reported this bug to GitHub and will update their replies in this issue (just incase somebody discovers the exact same issue).

weierophinney:
Per #453, using separator-style headers (those with `===` and `---` filled lines
following the header) can cause rendering issues if the header starts with a
`{number}.` prefix. All other proposals use `#`-prefixed headers, likely to
address this very issue, so this patch alters headers to do the same both for
consistency and to correct rendering issues.

Fixes #453

### DIFF
--- a/proposed/http-message-meta.md
+++ b/proposed/http-message-meta.md
@@ -1,8 +1,6 @@
-HTTP Message Meta Document
-==========================
+# HTTP Message Meta Document
 
-1. Summary
-----------
+## 1. Summary
 
 The purpose of this proposal is to provide a set of common interfaces for HTTP
 messages as described in [RFC 7230](http://tools.ietf.org/html/rfc7230) and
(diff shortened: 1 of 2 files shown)


proposed/http-message-meta.md
```diff
@@ -166,8 +161,7 @@ messages.
   there will be a certain amount of invention needed to describe HTTP message
   interfaces in PHP.
 
-5. Design Decisions
--------------------
+## 5. Design Decisions
 
 ### Message design
```
xabbuh: Would it make sense to add numbers to all headlines (e.g. `5.1` here)?

samdark: Probably yes. Could be easier to refer to specific paragraphs. Would you like making a pull request?

weierophinney: @samdark Yes, please. :-)

xabbuh: Sure, please have a look at #470.

PR #465, opened 2015-03-19
[PSR-7] Added note on stream mutability to section 1.3

weierophinney:
Largely re-iterates content currently in the metadocument, but @simensen and I
agree this should be an official part of the spec for clarity.

tl;dr: Streams are mutable, due to inability to make PHP resources immutable; behave accordingly.


weierophinney:
Ping @pmjones and/or @philsturgeon and/or @dragoonis — this is a clarification for PSR-7.

PR #466, opened 2015-03-19
[PSR-7] Examples of common stream implementations

weierophinney:
Common complaints/confusions about using streams as the basis for message bodies
include:

- returning files
- emitting output directly
- using iterators/generators for large data sets

This patch includes additions to the meta document to describe how these might
be achieved, linking to examples.


weierophinney:
I've had @simensen review these already; pinging @pmjones @philsturgeon and/or @dragoonis to merge!

ZOMG! I think we're about ready for a vote!

philsturgeon:
image

PR #467, opened 2015-03-20
PSR-7 typo

cebe:
I'm a bit late to the party but this should be fixed :)

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -39,7 +39,7 @@ This is the response body
 ```
 
 The first line is the "status line", and contains, in order, the HTTP protocol
-version, the HTTP status code, and a "reason phrase," a human-readable
+version, the HTTP status code, and a "reason phrase", a human-readable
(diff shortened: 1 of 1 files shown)


joshdifabio:
The original text is correct; punctuation _should_ go inside quotation marks[1].

[1] http://www.grammarbook.com/punctuation/quotes.asp (see 3a)

samdark:
Doesn't it depend on if it's British or American English? It is valid as is for American variant for sure...

joshdifabio:
The existing text is correct according to British convention. I can't speak for Americans, but according to the link I posted it's standard both in US and British English for punctuation to go within quotation marks.

samdark:
Indeed. Wikipedia seems to mislead about it. Both BBC and Guardian guides are suggesting comma inside of quotation marks. @cebe it seems original is correct.

cebe:
weird, did not know about this rule and looks strange to me. but okay if it is correct, leave it as is.

PR #468, opened 2015-03-21
[PSR-7] TCP abbreviation should be uppercase

samdark:
This is very minor change. `tcp` in text was changed to `TCP` since it's an abbreviation.


weierophinney:
@pmjones :+1:

PR #469, opened 2015-03-21
[PSR-7] Better wording about getting multiple values as string

samdark:
Modified explanation of "Headers with multiple values" and corresponding interface so it's now impossible to assume headers are always concatenated with comma even if reader isn't giving enough attention and skipping notes.

It's not affecting technical side of the PSR at all but provides a bit less confusing wording.

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -111,8 +111,8 @@ request or response.
 In order to accommodate headers with multiple values yet still provide the
 convenience of working with headers as strings, headers can be retrieved from
 an instance of a `MessageInterface` as an array or a string. Use the
-`getHeader()` method to retrieve a header value as a string containing all
-header values of a case-insensitive header by name concatenated with a comma.
+`getHeader()` method to retrieve a string containing concatenated
+header values of a case-insensitive header by name.
 Use `getHeaderLines()` to retrieve an array of all the header values for a
 particular case-insensitive header by name.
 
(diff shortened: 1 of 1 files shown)


proposed/http-message.md
```diff
@@ -111,8 +111,8 @@ request or response.
 In order to accommodate headers with multiple values yet still provide the
 convenience of working with headers as strings, headers can be retrieved from
 an instance of a `MessageInterface` as an array or a string. Use the
-`getHeader()` method to retrieve a header value as a string containing all
-header values of a case-insensitive header by name concatenated with a comma.
+`getHeader()` method to retrieve a string containing concatenated
+header values of a case-insensitive header by name.
```
weierophinney: The original verbiage indicated that the value WILL be returned as a comma-concatenated string. This was done to enforce a common expectation and signature from what is returned by the method. The implication is that if you know a given header CAN NOT be concatenated in this way, you will use the `getHeaderLines()` method and concatenate them manually. This also ensures that implementations do not have to be aware of every single header type, particularly as RFC 7230 allows for custom header types, which may have their own rules about concatenation.

weierophinney:
:-1:

mtdowling:
:-1:

proposed/http-message.md
```diff
@@ -385,12 +385,11 @@ interface MessageInterface
...
      *
-     * This method returns all of the header values of the given
-     * case-insensitive header name as a string concatenated together using
-     * a comma.
+     * This method returns a string containing concatenated header values of the given
+     * case-insensitive header name.
      *
-     * NOTE: Not all header values may be appropriately represented using
-     * comma concatenation. For such headers, use getHeaderLines() instead
+     * NOTE: Most header values are concatenated using a comma but some aren't because they could not be represented
```
GrahamCampbell: this line is too long

proposed/http-message.md
GrahamCampbell: the word header will fit on the previous line now

weierophinney:
@pmjones and/or @simensen — Please close; we resolved this with the most recent set of pull requests initiated this week.

PR #470, opened 2015-03-21
prefix all headlines with numbers for consistency

xabbuh:


proposed/http-message-meta.md
```diff
@@ -163,7 +163,7 @@ messages.
 
 ## 5. Design Decisions
 
-### Message design
+### 5.1 Message design
```
gaomd: Shouldn't we ending headline numbers with `.` just like `5. Design Decisions` did?

xabbuh: Well, it's also `4.1` and `4.2`,for example.

michaelcullum:
I'm not really sure I see a huge amount of benefit to these changes and as the spec is finalised, unless there is a good reason to change it, I'd rather leave it as-is.

PR #471, opened 2015-03-24
[PSR-7] Fixed a typo in an example

weierophinney:
`s/withRequestForm/withRequestTarget/`; thanks to @webmozart for reporting the issue.

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -230,7 +230,7 @@ For example, a user may want to make an asterisk-form request to a server:
 ```php
 $request = $request
     ->withMethod('OPTIONS')
-    ->withRequestForm('*')
+    ->withRequestTarget('*')
     ->withUri(new Uri('https://example.org/'));
 ```
 


simensen:
I'm of the mindset that this change should be allowed, even while in voting, as it was incorrect before. :+1:

@pmjones do you agree?

pmjones:
No changes while voting. Cannot have even the *appearance* of impropriety.

simensen:
@pmjones Can you tell me where in the bylaws it shows that? I tried to read it earlier and didn't see anything at all about "no changes while in voting."

I'd very much hate to have to pull this back to review for something like this. Would we be able to make the change after the fact? I'm not sure where it says what is appropriate to do where after the vote, either. /cc @weierophinney

pmjones:
It doesn't say it anywhere, but it doesn't have to. Precedent alone is enough.

Making minor corrections after the fact strikes me as being no problem, seeing as we have precedent for that as well.

weierophinney:
We can merge this after the vote is complete, as it's definitely a typo, and not a substantive change. (I've seen a number of similar PRs on the other proposals.)

simensen:
@webmozart We will attempt to merge this after voting has been completed if this PSR is accepted. We will not be able to make a change like this during the voting phase.

/cc @pmjones @weierophinney

weierophinney:
@pmjones Since we're back in review mode... this can be merged now. /cc @simensen

PR #473, opened 2015-03-26
[PSR-10][WIP] initial draft for a security disclosure PSR

lsmith77:
as discussed on the list we decided it would make more sense to split the security disclosure aspect from the security reporting aspect (https://github.com/php-fig/fig-standards/pull/393/files).

### DIFF
--- a/proposed/security-disclosure-publication.md
+++ b/proposed/security-disclosure-publication.md
@@ -1,0 +1,85 @@
+## Introduction
+
+There are two aspects with dealing with security issues: One is process
+by which security issues are reported and fixed in projects, the other
+is how the general public is informed about the issues and any remedies
+available. While PSR-9 addresses the former, this PSR, ie. PSR-10, deals with
+the later. That is how security issues are disclosed and what format such
+disclosures should follow. Especially today where PHP developers are sharing
+code across projects more than ever, this PSR aims to ease the challenges in
+keeping an overview of security issues in all dependencies and the steps 
+required to address them.
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in [RFC 2119][].
+
+[RFC 2119]: http://tools.ietf.org/html/rfc2119
+
+## Goal
+
+The goal of this PSR is to give project leads a clearly defined approach
+to allowing end users to discover security disclosures including a clearly
+defined structured format for disclosures.
+
+## Disclosure Discovery
+
+Every project MUST provide a link to its security vulnerability database in
+an obvious place. Ideally this should be on the main domain of the given
+project. This MAY be a sub-domain in case it is a sub-project of a larger
+initiative. If the project has a dedicated page for its disclosure process
+discovery then this is also considered a good place for this link.
+The link MAY use the custom link relation ``php-vuln-disclosures``,
+ie. for example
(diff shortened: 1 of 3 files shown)


zerocrates:
There are some inconsistencies with the meta doc and schema (ignoring problems with validity, since the older all-in-one PR admitted that the draft schema probably wasn't valid anyway) that make it a little hard to tell whether certain things are actually parts of the proposal.

These mostly feel like little typos and mismatches between the two documents, but in several cases there are a few reasonable interpretations of what the actual intent is.

* The set of mitigation types doesn't include `vendor fix`. Since it's mentioned in the meta-doc, I assume that's just an oversight
* The placement and naming of the `affected` element makes it seem like it's supposed to be the versions the *vuln* affects, but the schema has `affected` within the `remediation` element. The result is that it's not clear whether the remediations, the vuln, or both, will be able to specify applicable versions.
* There's a `summary` element defined within the remediation type that's not mentioned in the metadoc (generally, the metadoc seems to imply a flat structure where the different remediation-related elements are siblings, while the schema doc instead has a `remediation` element with children)
* As currently written with the repeatable `choice`, each remediation can contain an unlimited number of summaries, affected version declarations, and types. Since the remediations themselves repeat, it seems that one or all of those child elements is meant to instead have a maximum of 1

proposed/security-disclosure-publication.md
```diff
@@ -0,0 +1,84 @@
...
+ie. for example
+``<link rel="php-vuln-disclosures" href="http://example.org/disclosures"/>``.
+
+Note that projects MAY choose to host their disclosure files on a domain
+other than their main project page. It is RECOMMENDED to not store the
+disclosures in a VCS as this can lead to the confusions about which branch
+is the relevant branch. If a VCS is used then additional steps should be taken
+to clearly document to users which branch contains all vulnerabilities for
+all versions. If necessary projects MAY however split vulnerability disclosure
+files by major version number. In this case
```
stof: the sentence is incomplete here

lsmith77: fixed

lsmith77:
noted in the intro that this is PSR-10

mlhess:
Speaking for Drupal here.  
* severity (low, medium high)
Should be clarified some place.  (Forgive me if I missed it) 
In Drupal space, we worked to create something that calculates the risk and then gives a score, the score is from the following questions:

Access complexity: how difficult is it for the attacker to leverage the vulnerability? *
Authentication: what privilege level is required for an exploit to be successful? *
Confidentiality impact: does this vulnerability cause non-public data to be accessible? *
Integrity impact: can this exploit allow system data (or data handled by the system) to be compromised? *
Zero-day impact: does a known exploit exist? *
Target distribution: what percentage of module users are affected? *

The questions come in part from the NIST standards.

barryvdh:
As this PSR is supposed to automate things, will this also link to patches, updates etc? For example, Magento 1 releases a .sh patch to fix security issues. I can imagine that it would be useful to automate the downloading/discovery of those patches.
But you might also want to link to the releases notes, a press releases, download page etc. Should an (optional) distinction of the link type be made?

lsmith77:
@mlhess yeah severity is tricky and opens up a lot of questions, which might be beyond what we can cover in this standard. ie if we try to solve that, we might never get done. so we may need to leave it to the source to define their concept of severity.

lsmith77:
@barryvdh the idea is indeed to enable providing information about what steps to take to download fixes and how to apply them. however we also do not want to complicate things by becoming a build system of sorts. imho it makes more sense for people to start investigating into composer and post install scripts.

mlhess:
@lsmith77 can we keep a field for it in there and maybe some simple guidance on how to classify issues?

PR #477, opened 2015-03-30
[PSR-6] Split setExpiration() into two methods

Crell:
Final edits that were lost when Paul deleted the public repository.

### DIFF
--- a/proposed/cache.md
+++ b/proposed/cache.md
@@ -237,7 +237,7 @@ interface CacheItemInterface
     public function exists();
 
     /**
-     * Sets the expiration time this cache item.
+     * Sets the expiration time for this cache item.
      *
      * @param \DateTime|\DateTimeImmutable $expiration
      *   The point in time after which the item MUST be considered expired.
@@ -251,7 +251,7 @@ interface CacheItemInterface
     public function expiresAt($expiration);
 
     /**
-     * Sets the expiration time this cache item.
+     * Sets the expiration time for this cache item.
      *
      * @param int|\DateInterval $time
      *   The period of time from the present after which the item MUST be considered
--- a/proposed/cache.md
+++ b/proposed/cache.md
@@ -200,18 +200,10 @@ interface CacheItemInterface
      *
      * @param mixed $value
      *   The serializable value to be stored.
-     * @param int|\DateTime $ttl
-     *   - If an integer is passed, it is interpreted as the number of seconds
-     *     after which the item MUST be considered expired.
-     *   - If a DateTime object is passed, it is interpreted as the point in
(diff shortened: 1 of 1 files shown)


tedivm:
Please note that these methods were written unilaterally by @Crell and were not approved by anyone else on the PSR-6 group. Please hold off on merging these until a discussion can be had on the mailing list.

proposed/cache.md
```diff
@@ -227,21 +235,31 @@ interface CacheItemInterface
     public function exists();
 
     /**
-     * Sets the expiration for this cache item.
+     * Sets the expiration time this cache item.
```
neclimdul: Not sure if this is a correct sentence. It at least does not make it clear what the method is expected to do to me.

proposed/cache.md
```diff
@@ -227,21 +235,31 @@ interface CacheItemInterface
...
+     *   the value should be stored permanently or for as long as the
+     *   implementation allows.
+     *
+     * @return static
+     *   The called object.
+     */
+    public function expiresAt($expiration);
+
+    /**
+     * Sets the expiration time this cache item.
```
neclimdul: Not sure if this is a correct sentence. It at least does not make it clear what the method is expected to do to me.

proposed/cache.md
```diff
@@ -227,21 +235,31 @@ interface CacheItemInterface
...
      *
-     * @param int|\DateTime $ttl
-     *   - If an integer is passed, it is interpreted as the number of seconds
-     *     after which the item MUST be considered expired.
-     *   - If a DateTime object is passed, it is interpreted as the point in
-     *     time after which the item MUST be considered expired.
-     *   - If null is passed, a default value MAY be used. If none is set,
-     *     the value should be stored permanently or for as long as the
-     *     implementation allows.
+     * @param \DateTime|\DateTimeImmutable $expiration
```
AndrewCarterUK: Both of these classes implement \DateTimeInterface

AndrewCarterUK:
I think splitting setExpiration() into expiresAt() and expiresAfter() makes things a lot more readable. It's clear that expiresAt() requires a point in time to be reached and expiresAfter() requires a period of time to have passed.

With the existence of these methods the $ttl parameter on the set() method becomes redundant as you can do this:

```php
$item->set($value)->expiresAfter(3600);
```

Which is easier to read than this:

```php
$item->set($value, 3600);
```

proposed/cache.md
Crell: I should have added, not replaced. :-)  Fixed.

proposed/cache.md
Crell: Agreed. Fixed.

Issue #479, opened 2015-03-31
Should UriInterface::withPath percent-encode slashes?


mnpenner:
The [documentation for `UriInterface::withPath`](https://github.com/php-fig/http-message/blob/ac2460e2eaba78395423bfceeb09736b1efa34fc/src/UriInterface.php#L208-L210) says:

> The implementation MUST percent-encode reserved characters as
> specified in RFC 3986, Section 2, but MUST NOT double-encode any
> characters.

[RFC 3986](https://www.ietf.org/rfc/rfc3986.txt) says the reserved characters are:

> reserved    = gen-delims / sub-delims  
> gen-delims  = ":" / "/" / "?" / "#" / "[" / "]" / "@"
> sub-delims  = "!" / "$" / "&" / "'" / "(" / ")"
> / "*" / "+" / "," / ";" / "="

i.e., it includes `/`.  Does that mean `withPath` should percent-encode `/`? Or should it percent-encode those (and only those) characters *with the exception of* `/`?

---

Just noticed [Tobion's comment](https://github.com/php-fig/fig-standards/pull/449#issuecomment-77266583) but it wasn't addressed.

samdark:
It should encode `/` if it is a part of parameter but should not if it's separating path itself i.e. `http://example.com/test/it&path=a%2Fb%2Fc`.

mnpenner:
@samdark How is `withPath` supposed to understand your intention?

samdark:
I have no idea, just explained what's meant in RFC.

samdark:
Thinking about it a bit more I guess it can be determined by presence of `&` i.e. before `&` it's path where `/` should be left as is, after `&` it's parameters which should have `/` encoded.

mnpenner:
@samdark The entire string *is* the path. The method name is `withPath` and it accepts a path as its only argument.

samdark:
Right. Then there's basically no way to tell where it's meant to be encoded and where not.

mnpenner:
Indeed. So what does FIG have to say about this?

Perhaps the method should instead take an array as the argument, wherein each element represents a path segment?

pjdietz:
I don't think you'd want the method to encode a slash itself. Where this might come up is if you had an endpoint like `/characters/{name}` and you wanted `{name}` to extract "homer/marge", you'd need to use the URI `/characters/homer%2Fmarge`. In cases like this, I think the user will just have to pre-encode that portion.

samdark:
[...]

PR #481, opened 2015-04-01
[PSR-7] Clarification of path, query, and fragment encoding

weierophinney:
Comments on #479 and #449 indicate uncertainty about how to encode the path, particularly the `/` character; additionally, if references to encoding are made in other methods, these should be addressed.

This patch:

- Removes references to encoding from the `withPath()`, `withQuery()`, and `withFragment()` methods.
- Adds verbiage on encoding to the `getPath()`, `getQuery()`, and `getFragment()` methods, indicating they MUST be percent-encoded, but MUST NOT be double-encoded.
- Adds references to RFC 3986 Section 2 and the relevant component-specific section in each of the `getPath()`, `getQuery()` and `getFragment()` methods as the canonical source for determining what should and should not be encoded.
- Adds verbiage to `getPath()` specifically around encoding of the `/` character.

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -531,7 +531,7 @@ interface RequestInterface extends MessageInterface
      *
      * This method acts exactly like MessageInterface::getHeaders(), with one
      * behavioral change: if the Host header has not been previously set, the
-     * method MUST attempt to pull the host segment of the composed URI, if
+     * method MUST attempt to pull the host component of the composed URI, if
      * present.
      *
      * @see MessageInterface::getHeaders()
@@ -548,7 +548,7 @@ interface RequestInterface extends MessageInterface
      * This method acts exactly like MessageInterface::getHeader(), with
      * one behavioral change: if the Host header is requested, but has
      * not been previously set, the method MUST attempt to pull the host
-     * segment of the composed URI, if present.
+     * component of the composed URI, if present.
      *
      * @see MessageInterface::getHeader()
      * @see UriInterface::getHost()
@@ -570,7 +570,7 @@ interface RequestInterface extends MessageInterface
      * This method acts exactly like MessageInterface::getHeaderLines(), with
      * one behavioral change: if the Host header is requested, but has
      * not been previously set, the method MUST attempt to pull the host
-     * segment of the composed URI, if present.
+     * component of the composed URI, if present.
      *
      * @see MessageInterface::getHeaderLine()
      * @see UriInterface::getHost()
@@ -1221,17 +1221,17 @@ interface UriInterface
     public function getUserInfo();
 
     /**
-     * Retrieve the host segment of the URI.
+     * Retrieve the host component of the URI.
      *
-     * This method MUST return a string; if no host segment is present, an
+     * This method MUST return a string; if no host component is present, an
      * empty string MUST be returned.
      *
-     * @return string Host segment of the URI.
+     * @return string Host component of the URI.
      */
     public function getHost();
 
     /**
-     * Retrieve the port segment of the URI.
+     * Retrieve the port component of the URI.
      *
      * If a port is present, and it is non-standard for the current scheme,
      * this method MUST return it as an integer. If the port is the standard port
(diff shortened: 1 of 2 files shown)


samdark:
Great explanation.

weierophinney:
Ping @simensen and @Tobion for review.

proposed/http-message.md
```diff
@@ -1236,6 +1236,16 @@ interface UriInterface
      * This method MUST return a string; if no path is present it MUST return
      * the string "/".
      *
+     * The value returned MUST be returned in percent-encoded form, but MUST NOT
+     * double-encode any characters. To determine what characters to encode,
+     * please refer to RFC 3986, Sections 2 and 3.3.
+     *
+     * If the value should include a slash ("/") not intended as delimeter
+     * between path segments, that value MUST be encode (e.g., "%2F") when
```
mtdowling: must be *encoded*

proposed/http-message.md
```diff
@@ -1236,6 +1236,16 @@ interface UriInterface
      * This method MUST return a string; if no path is present it MUST return
      * the string "/".
      *
+     * The value returned MUST be returned in percent-encoded form, but MUST NOT
+     * double-encode any characters. To determine what characters to encode,
+     * please refer to RFC 3986, Sections 2 and 3.3.
+     *
+     * If the value should include a slash ("/") not intended as delimeter
```
graste: "delimeter" vs. "delimiter"

proposed/http-message.md
graste: "MUST be encode" vs. "MUST be encoded"

proposed/http-message.md
```diff
@@ -1236,6 +1236,16 @@ interface UriInterface
      * This method MUST return a string; if no path is present it MUST return
      * the string "/".
      *
+     * The value returned MUST be returned in percent-encoded form, but MUST NOT
+     * double-encode any characters. To determine what characters to encode,
+     * please refer to RFC 3986, Sections 2 and 3.3.
+     *
+     * If the value should include a slash ("/") not intended as delimiter
```
Tobion: I would try to make clear that this is just an example. The same applies for the other reserved characters as well.

weierophinney:
@Tobion I've incorporated your feedback now.

proposed/http-message.md
```diff
@@ -1260,6 +1280,15 @@ interface UriInterface
      *
      * The string returned MUST omit the leading "#" character.
      *
+     * The value returned MUST be percent-encoded, but MUST NOT double-encode
+     * any characters. To determine what characters to encode, please refer to
+     * RFC 3986, Sections 2 and 3.5.
+     *
+     * As an example, if the fragment should include a hash symbol ("#"), that
+     * value MUST be encoded (e.g., "%23") when passed to the instance.
```
Tobion: This seems wrong to me. For the fragment, a  hash is always wrong inside the fragment. So it must be encoded anyway. So it does not need to be passed encoded to `withFragment` since it should be encoded automatically anyway. So `withFragment('foo#bar')` should be equal to `withFragment('foo%23bar')`.

The examples for path and query are correct because both version exist: encoded and non-encoded. This is why the caller must decide what he means.

Tobion: So I guess for fragment we can just remove that sentence since there are no chars with special meaning in there.

weierophinney:
@Tobion I've removed the sentence now.

@pmjones and/or @simensen — This one is ready for merge.

proposed/http-message.md
```diff
@@ -1236,7 +1236,17 @@ interface UriInterface
...
+     * double-encode any characters. To determine what characters to encode,
+     * please refer to RFC 3986, Sections 2 and 3.3.
+     *
+     * As an example, if the value should include a slash ("/") not intended as
+     * delimiter between path segments, that value MUST be encoded (e.g., "%2F")
+     * when passed to the instance.
+     *
+     * @see https://tools.ietf.org/html/rfc3986#section-2
+     * @see https://tools.ietf.org/html/rfc3986#section-3.3
+     * @return string The path segment of the URI.
```
Tobion: another wrong rebase

weierophinney: @Tobion what is incorrect, exactly?

Tobion: path "component", see diff

weierophinney:
@Tobion fixed.

@pmjones and/or @simensen — ready for merge.

PR #483, opened 2015-04-01
[PSR-7] "integer" is not a valid scalar type

GrahamCampbell:
It has always been the case that integer and boolean are aliases of int and bool, despite people's resistance to this (blame java, whatever), but as of php 7.0, integer really isn't allowed. PHP's implementation of scalar type hinting only allows you to use `int`, not `integer`. I think we should respect this and comply with PHP's official scalar types.

// cc @ircmaxell

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -930,7 +930,7 @@ interface ResponseInterface extends MessageInterface
      * The Status-Code is a 3-digit integer result code of the server's attempt
      * to understand and satisfy the request.
      *
-     * @return integer Status code.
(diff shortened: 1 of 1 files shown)


Tobion:
:+1:

nyamsprod:
:+1:

mtdowling:
:+1:

simensen:
@weierophinney Are you manually synchronizing the package yourself?

weierophinney:
@simensen - yes. Periodically updating the http - message repo and/or creating PR pairs.

PR #488, opened 2015-04-03
Path "component" instead of "segment"

Tobion:
A path segment is something different, it's only one part between two slashes. But the path component is the full path. See RFC 3986

Btw, the rest of PSR-7 also speaks about "segments", e.g. host segment. But these are also named host component in RFC 3986. I don't see why we should diverge from the standard here...

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -1231,12 +1231,12 @@ interface UriInterface
     public function getPort();
 
     /**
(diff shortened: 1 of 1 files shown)


mtdowling:
:+1:

weierophinney:
:+1: — @pmjones and/or @simensen — please merge.

PR #490, opened 2015-04-07
Renamed StreamableInterface to StreamInterface

webmozart:
Since both the PSR-7 text as well as the doc blocks speak of "streams" and not of "streamable objects", the naming of the interface should be changed to match this language.

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -151,13 +151,13 @@ message can be very small or extremely large. Attempting to represent the body
 of a message as a string can easily consume more memory than intended because
 the body must be stored completely in memory. Attempting to store the body of a
 request or response in memory would preclude the use of that implementation from
-being able to work with large message bodies. `StreamableInterface` is used in
+being able to work with large message bodies. `StreamInterface` is used in
 order to hide the implementation details when a stream of data is read from
 or written to. For situations where a string would be an appropriate message
 implementation, built-in streams such as `php://memory` and `php://temp` may be
 used.
 
-`StreamableInterface` exposes several methods that enable streams to be read
+`StreamInterface` exposes several methods that enable streams to be read
 from, written to, and traversed effectively.
 
 Streams expose their capabilities using three methods: `isReadable()`,
@@ -169,10 +169,10 @@ write-only, or read-write. It can also allow arbitrary random access (seeking
 forwards or backwards to any location), or only sequential access (for
 example in the case of a socket, pipe, or callback-based stream).
 
-Finally, `StreamableInterface` defines a `__toString()` method to simplify
+Finally, `StreamInterface` defines a `__toString()` method to simplify
 retrieving or emitting the entire body contents at once.
 
-Unlike the request and response interfaces, `StreamableInterface` does not model
+Unlike the request and response interfaces, `StreamInterface` does not model
 immutability. In situations where an actual PHP stream is wrapped, immutability
 is impossible to enforce, as any code that interacts with the resource can
 potentially change its state (including cursor position, contents, and more).
@@ -466,24 +466,24 @@ interface MessageInterface
     /**
      * Gets the body of the message.
      *
-     * @return StreamableInterface Returns the body as a stream.
+     * @return StreamInterface Returns the body as a stream.
      */
     public function getBody();
 
     /**
      * Create a new instance, with the specified message body.
      *
-     * The body MUST be a StreamableInterface object.
+     * The body MUST be a StreamInterface object.
      *
      * This method MUST be implemented in such a way as to retain the
      * immutability of the message, and MUST return a new instance that has the
      * new body stream.
      *
-     * @param StreamableInterface $body Body.
+     * @param StreamInterface $body Body.
      * @return self
      * @throws \InvalidArgumentException When the body is not valid.
      */
-    public function withBody(StreamableInterface $body);
+    public function withBody(StreamInterface $body);
 }
 ```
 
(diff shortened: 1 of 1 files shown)


pmjones:
:+1:

weierophinney:
:-1: for rains stated on the mailing list: major change (change in interface name) during review period, and good arguments also exist for the current naming.

samdark:
+1 for renaming it even though it may hold getting back to voting again.

weierophinney:
Considering good arguments can be made for either `StreamInterface` or `StreamableInterface`, the past of least resistance is to keep the status quo, particularly as this was never a point of contention in any prior discussions.

@pmjones and/or @simensen, please close.

webmozart:
I would like you to consider merging this PR again. My arguments stand, we had comments in favor above and so far the only major counter argument was that we were in Review stage, which we aren't now.

weierophinney:
@webmozart - My point was that there are compelling arguments for either name, which makes sticking with what we have now just as reasonable as merging.

webmozart:
> compelling arguments

Would you mind repeating them for StreamableInterface? I couldn't find them in the meta document. 

The only argument that I can remember is that "an implementation does not have to be backed by a PHP stream", but to me this argument is irrelevant. Through the interface, we create our own definition of a "stream" that clearly distinguishes a "PSR-7 stream" from a "PHP stream". This is is important when the reader reads "stream" in the document. Right now, it's not entirely clear whether "stream" refers to a "streamable" or a "PHP stream".

weierophinney:
The principal and most compelling argument is that it's not an actual PHP stream, and cannot be used directly as a PHP stream without first wrapping it in either a `StreamWrapper` or creating a PHP stream context around it. As such, it's *streamable*, but **not** a *stream*.

webmozart:
Apparently I did not make my argument clear. Let me try again. Please reread the following quote from the PSR about streams:

> ### 1.3 Streams
>
> `StreamableInterface` is used in
order to hide the implementation details when a stream of data is read from
or written to. For situations where a string would be an appropriate message
implementation, **built-in streams [1]** such as `php://memory` and `php://temp` may be
used.
>
> `StreamableInterface` exposes several methods that enable **streams** to be read
from, written to, and traversed effectively.
>
>**Streams** expose their capabilities using three methods: `isReadable()`,
`isWritable()`, and `isSeekable()`. These methods can be used by **stream**
collaborators to determine if a stream is capable of their requirements.
>
> Each **stream instance** will have various capabilities: it can be read-only,
write-only, or read-write. It can also allow arbitrary random access (seeking
forwards or backwards to any location), or only sequential access (for
example in the case of a socket, pipe, or callback-based **stream**).
>
> Finally, `StreamableInterface` defines a `__toString()` method to simplify
retrieving or emitting the entire body contents at once.
>
> Unlike the request and response interfaces, `StreamableInterface` does not model
immutability. In situations where an actual **PHP stream [2]** is wrapped, immutability
is impossible to enforce, as any code that interacts with the resource can
potentially change its state (including cursor position, contents, and more).
Our recommendation is that implementations use **read-only streams** for
server-side requests and client-side responses. Consumers should be aware of
the fact that the **stream instance** may be mutable, and, as such, could alter
the state of the message; when in doubt, create a new **stream instance** and attach
it to a message to enforce state.

Technically, this section makes no sense, as we never defined what a "stream" is. 

Solution 1: Replace all bold occurrences by "streamable" so that the text makes sense. But then the language is unnatural. 

Solution 2: Rename the interface to `StreamInterface`. Then "stream" is clearly defined. In exceptions, clarify the type of the stream, e.g. "PHP stream" (as you're doing already, see [1] and [2])

weierophinney:
@webmozart And what of the cases where the implementation is not backed by an actual stream? Such cases are another reason why I feel Streamable makes more sense, as it inputs that the backing "resource", whatever it is, had been made compatible with streams. 

Regarding your primary concern: We don't define what a stream is because it's a language concept (as in, part of the php language).

TBH, I *do* see your point. I also see his reasons for the current naming. Since the easiest path is to *not* make a change, that's why I recommended closing this request. 

@mtdowling - do you have any thoughts to share, considering this was your contribution?

mtdowling:
When I initially proposed it, they were called StreamInterface.

I agree with @webmozart that the language in prose should be more clear to better distinguish  from PHP streams. That said, it really doesn't matter if it's called StreamInterface or StreamableInterface... 

If I had to pick one, I prefer StreamInterface. To me, an "able" suffix is usually some kind of action that can be taken. In this instance, it's actually a thing (a stream). Also of note is that HTTP/2 has streams and not streamables: http://http2.github.io/http2-spec/#StreamsLayer

m00t:
For me, stream is an abstraction which provide functionality like get next char, seek etc. Object returned from getBody() does not need to be backed by php native streams to be a stream. That is why I personally prefer `StreamInterface` name. It does not mean "this object backed by php native streams". It does not mean "php native streams can be replaced by this object" either. It means for me "this object is a stream", which sounds a 100% correct in my opinion.

weierophinney:
@simensen and/or @pmjones - phrase reopen and merge.

auroraeosrose:
Naming is an important thing - many many things are overloaded in computing
Naming this as a "stream" is adding another layer of overloading on a name that has a very specific meaning in PHP and overloading it with psr's definition of "stream" - whether you're meaning it as an abstract stream or not (do you know how problematic the filter extension naming is already? please please let's not make it worse)
This will confuse users which is bad when it's supposed to be designed to help

This will also make adding a StreamInterface in core PHP problematic
I really think this is a very very bad idea for the simple fact that if(when) PHP DOES get a streaminterface suddenly search mechanisms are going to return very convoluted values and confuse the heck out of users as we suddenly have to explain "well, a PHP native stream is this but in PSR-7 a stream is this"

PR #491, opened 2015-04-07
Made MessageInterface::getHeaders() consistent with getHeader()

webmozart:
Currently, the following consistency condition is not satisfied:

```php
$message->getHeaders()[$k] === $message->getHeader($k)
```

This PR takes the least intrusive route of fixing this consistency problem. An alternative is to change getHeader($k) to return an array, but then getHeader() == getHeaderLines(), which is probably not desired.

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -350,17 +350,18 @@ interface MessageInterface
      * Retrieves all message headers.
      *
      * The keys represent the header name as it will be sent over the wire, and
-     * each value is an array of strings associated with the header.
+     * each value contains the values for the header concatenated together using
+     * a comma.
      *
      *     // Represent the headers as a string
-     *     foreach ($message->getHeaders() as $name => $values) {
-     *         echo $name . ": " . implode(", ", $values);
+     *     foreach ($message->getHeaders() as $name => $value) {
+     *         echo $name . ": " . $value;
      *     }
      *
      *     // Emit headers iteratively:
-     *     foreach ($message->getHeaders() as $name => $values) {
-     *         foreach ($values as $value) {
-     *             header(sprintf('%s: %s', $name, $value), false);
+     *     foreach ($message->getHeaders() as $name => $value) {
+     *         foreach ($message->getHeaderLines($name) as $line) {
+     *             header(sprintf('%s: %s', $name, $line), false);
      *         }
      *     }
      *
@@ -368,7 +369,7 @@ interface MessageInterface
      * exact case in which headers were originally specified.
      *
      * @return array Returns an associative array of the message's headers. Each
-     *     key MUST be a header name, and each value MUST be an array of strings.
+     *     key MUST be a header name, and each value MUST be a string.
      */
     public function getHeaders();
 


pmjones:
:+1:

simensen:
@pmjones Would you think this change would constitute a big enough change that it would require going back to draft? I could see a case for #490 maybe just being clarification but even that one seems to be enough of a change (actually changing an interface name) that I'm not sure we can consider these without going back to draft.

pmjones:
(/me ponders)

I don't think so. This is a consistency issue, and while it modifies an expected return value, it does not modify the signature.

(I will admit to not having liked the inconsistency before, but felt like the only one. Now that there's independent review I feel it's a change that should be made.)

weierophinney:
:-1: This is a HUGE change, everyone. The reason the current method returns an array of arrays is to facilitate emission of headers. Since a number of headers cannot be represented using comma concatenation, this allows emission of multiple header line for individual header fields in a trivial manner. Changing that would require addition of a separate method for retrieving the internal structure.

We chose the current behavior of getHeader (return a string)as it's the most common expectation for retrieving a single header; we have a complementary method, getHeaderLines, to retrieve the array of values for that header. However, the most common use case for retrieving all headers at once is for emitting them at once (either a client sending them to a server, or a server emitting them back to the client).

In other words, the signatures are optimized for the most common use cases.

pmjones:
My primary issue is with the concatenation of the strings as the returned value. It should just return an array of the strings, and let the user concatenate them as the user wishes.

pmjones:
Let me put it this way. What I would expect of getHeader('X-Foo') would be:

    [
        'foo-value-1',
        'foo-value-2',
        'foo-value-3',
    ]

What I would expect of getHeaders() would be:

    [
        'X-Foo' => [
            'foo-value-1',
            'foo-value-2',
            'foo-value-3',
        ],
        'X-Bar' => [ 
            'bar-value-1',
        ],

        // even a single value is an array element
        'X-Baz' => [
            'baz-value-1',
        ],
    ]

Which version of the document (i.e., the current one, or the proposed changed one) exhibits that behavior?

pjdietz:
Neither. The current will give you those results for `getHeaderLines('X-Foo')` and `getHeaders()`. The proposed would give you:

```
"foo-value-1, foo-value-2, foo-value-3"
```

And 

```
[
    'X-Foo' => 'foo-value-1, foo-value-2, foo-value-3',
    'X-Bar' => 'bar-value-1',
    'X-Baz' => 'baz-value-1'
]
```

I think leaving everything as is makes more sense as you get all the details with `getHeaders()` and `getHeaderLines()`, and have `getHeader()` as a convenience method for when you just want the concatenated string.

pmjones:
I will admit to having a habitual expectation that a plural of a method should return a collection of the values from a singular method. To wit, I expect `getThings()` to return a collection of results from `getThing()`. I think that's a reasonable expectation. I have been quiet about this inconsistency in the API because it seemed like I was the only one with that expectation, but now that someone else has brought it up, I think it's wise to examine it again.

weierophinney:
@pmjones I'll dig up the ML threads where the decisions were made later when I have a computer in front of me. We chose explicitly not to do an "as string" variant as that's the primary use case for accessing a single header. The consensus at the time was "naming isn't great, but it's concise and addresses the most common use cases." We can reexamine, but I want us to do so in light of previous discussions.

pmjones:
Totally fair. I am willing to be (re-)convinced, especially in the interest of expediency. But the inconsistency does grate on me.

pmjones:
Pre-emptively, as an alternative, perhaps two sets of methods would be wise: `getHeaders()` and `getHeader()` return the value(s) as set, and `getHeaderLines()` and `getHeaderLine()` (a new method) return the concatenated strings with the header label. I know that's a big change this late in the game so I mention it only as a way to address the consistency issue.

pmjones:
@weierophinney @webmozart please take a look at #493 as an alternative to this PR.

webmozart:
Closing in favor of #493.

PR #492, opened 2015-04-07
Added UploadedFileInterface

webmozart:
The structure of PHP's $_FILES array is inconsistent with the structure of $_POST when the names of the form fields contain array keys.

For example, when submitting the text inputs "text[0]" and "text[1]", the contents of $_POST are:

```
array(1) { 
    ["text"]=>array(2) { 
        [0]=>string(5)"text0" 
        [1]=>string(5)"text1" 
    } 
} 
```

However, when submitting the file inputs "file[0]" and "file[1]", the contents of $_FILES are:

```
array(1) { 
    ["file"]=>array(2) { 
        ["name"]=>array(2) { 
            [0]=>string(9)"file0.txt" 
            [1]=>string(9)"file1.txt" 
        } 
        ["type"]=>array(2) { 
            [0]=>string(10)"text/plain" 
            [1]=>string(10)"text/html" 
        } 
    } 
} 
```

While you can read `$_POST['text'][1]` to access the contents of the field "text[1]", you *cannot* read `$_FILES['file'][1]` to access the contents of the field "file[1]", which is a quite big usability issue in practice. Symfony, for example, rearranges the contents of the $_FILES array for that sake.

I think PSR-7 should fix this issue. Otherwise every user of the PSR interfaces must fix this themselves, which isn't exactly interoperable.

I see two approaches to fix this:

* Describe this issue and the precise rules for rearranging the array in the PSR. This is verbose and prone to mistakes, which is why I don't like it very much.
* Enforce implementations to return `UploadedFileInterface` objects "in the same structure" as the $_FILES array (same wording as in getQueryParams()). This solution is concise and leaves little room for interpretation. The correct result in the above case would be:

```
array(1) { 
    ["file"]=>array(2) { 
        [0]=>object(UploadedFileInterface)
        [1]=>object(UploadedFileInterface)
    } 
} 
```

Hence it is possible to read `$message->getUploadedFiles()['file'][1]` to access all data related to the field "file[1]".

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -785,6 +785,11 @@ interface ServerRequestInterface extends RequestInterface
      * request. They SHOULD be injected during instantiation, such as from PHP's
      * $_FILES superglobal, but MAY be derived from other sources.
      *
+     * The implementation SHOULD ensure that is_uploaded_file() returns true for
+     * the paths of the uploaded files. However, this recommendation MAY be
+     * ignored in special cases, for example when using temporary files as
+     * uploaded files during testing.
+     *
      * @return UploadedFileInterface[] The uploaded file(s), if any.
      */
     public function getUploadedFiles();
@@ -1445,9 +1450,6 @@ interface UploadedFileInterface
      *
      * Implementations SHOULD return the value stored in the "tmp_name" key
      * of the file in the $_FILES array.
-     *
-     * Implementations MUST guarantee that is_uploaded_file() returns true
-     * for the returned path.
      *
      * @return string The absolute path to the uploaded file.
      */
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -785,11 +785,6 @@ interface ServerRequestInterface extends RequestInterface
      * request. They SHOULD be injected during instantiation, such as from PHP's
      * $_FILES superglobal, but MAY be derived from other sources.
      *
-     * The implementation SHOULD ensure that is_uploaded_file() returns true for
-     * the paths of the uploaded files. However, this recommendation MAY be
-     * ignored in special cases, for example when using temporary files as
-     * uploaded files during testing.
-     *
      * @return UploadedFileInterface[] The uploaded file(s), if any.
      */
     public function getUploadedFiles();
@@ -1450,6 +1445,12 @@ interface UploadedFileInterface
      *
      * Implementations SHOULD return the value stored in the "tmp_name" key
      * of the file in the $_FILES array.
+     *
+     * The implementation MUST guarantee that the file pointed to by the path
+     * can safely be used by the calling code. More precisely, the
+     * implementation MUST ensure that is_uploaded_file() returns true for
+     * the path if the path was provided through the $_FILES array or a
+     * similar mechanism.
      *
      * @return string The absolute path to the uploaded file.
      */
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -1446,6 +1446,9 @@ interface UploadedFileInterface
      * Implementations SHOULD return the value stored in the "tmp_name" key
      * of the file in the $_FILES array.
      *
+     * Implementations MUST guarantee that is_uploaded_file() returns true
+     * for the returned path.
+     *
      * @return string The absolute path to the uploaded file.
      */
     public function getPath();
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -776,18 +776,32 @@ interface ServerRequestInterface extends RequestInterface
     public function withQueryParams(array $query);
 
     /**
-     * Retrieve the upload file metadata.
+     * Retrieve the uploaded files.
      *
-     * This method MUST return file upload metadata in the same structure
+     * This method MUST return the file upload metadata in the same structure
      * as PHP's $_FILES superglobal.
      *
      * These values MUST remain immutable over the course of the incoming
      * request. They SHOULD be injected during instantiation, such as from PHP's
      * $_FILES superglobal, but MAY be derived from other sources.
      *
-     * @return array Upload file(s) metadata, if any.
+     * @return UploadedFileInterface[] The uploaded file(s), if any.
      */
-    public function getFileParams();
+    public function getUploadedFiles();
+
+    /**
+     * Create a new instance with the specified uploaded files.
+     *
+     * These MAY be injected during instantiation.
+     *
+     * This method MUST be implemented in such a way as to retain the
+     * immutability of the message, and MUST return a new instance that has the
+     * updated body parameters.
+     *
+     * @param UploadedFileInterface[] $uploadedFiles The uploaded files.
+     * @return self
+     */
+    public function withUploadedFiles(array $uploadedFiles);
 
     /**
      * Retrieve any parameters provided in the request body.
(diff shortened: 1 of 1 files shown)


webmozart:
The wording of the doc block possibly needs to be improved to prevent "in the same structure" being interpreted as:

```
array(1) { 
    ["file"]=>array(2) { 
        ["name"]=>array(2) { 
            [0]=>object(UploadedFileInterface)#0
            [1]=>object(UploadedFileInterface)#1
        } 
        ["type"]=>array(2) { 
            [0]=>object(UploadedFileInterface)#0
            [1]=>object(UploadedFileInterface)#1
        } 
    } 
} 
```

Please tell me whether you think this is necessary. I'd also be happy if you proposed how to word the fix.

simensen:
@webmozart I'm glad you brought this up again but I'm not sure we'll be able to do anything about it without going back to draft. I'll talk to @weierophinney and @pmjones about it.

proposed/http-message.md
```diff
@@ -776,18 +776,32 @@ interface ServerRequestInterface extends RequestInterface
...
-     * This method MUST return file upload metadata in the same structure
+     * This method MUST return the file upload metadata in the same structure
      * as PHP's $_FILES superglobal.
      *
      * These values MUST remain immutable over the course of the incoming
      * request. They SHOULD be injected during instantiation, such as from PHP's
      * $_FILES superglobal, but MAY be derived from other sources.
      *
-     * @return array Upload file(s) metadata, if any.
+     * @return UploadedFileInterface[] The uploaded file(s), if any.
```
weierophinney: This doesn't cover how to resolve "nested" file uploads per the example in your summary. For instance, if the files were in the form as `example[image]` and `example[icon]`, how would I retrieve these from the returned metadata? What you detail here is an array of upload files, but in your summary, I think this is more likely a _tree_.

weierophinney:
First, thanks for bringing the detailed write-up, @webmozart — I forgot that we do this as well in ZF1 and ZF2.

I'm of a couple of minds on this.

First, what you describe looks more like a tree, and not an array of objects. If you retrieve `$_FILES['file'][0]`, per your example in the summary, the structure would be:

```php
array(
    'file' => aray(
        0 => UploadedFileInterface,
        1 => UploadedFileInterface,
    ),
)
```

This is definitely conducive to retrieving the file information by the given name and depth within the form, but not the simple `UploadedFileInterface[]` that you specify as a return value in the `ServerRequestInterface`.

The next question I have is whether we need an interface for the file upload details, or if we can rather just specify the format of an individual file specification as an array. I think if we went with the latter (specifying the array format), we can make this a clarification to the specification, rather than a substantive addition.

However, this brings up my last point. In the case of headers, we've made the decision that we will not treat individual headers as part of the specification. This means that if you want to, for instance, examine cookies, or set cookies, or work with cache control headers, you would delegate this functionality to another library:

```php
$cookies = new CookieJar($request->getHeaderLines('Cookie'));
$value = $cookies->get('chocolate', 'fudge');
```

I wonder if we should not do similarly here, and keep the specification as it currently is. If you want to retrieve or work with the files, you pull them from the file params and pass them on to a processor:

```php
$files = new FileUploads($request->getFileParams());
if ($files->isValid('example', 'image')) { // pull example[image] file and validate
}
```

In `Zend\Filter`, `Zend\Validator`, and `Zend\File\Transfer`, we do something similar to the above, in that we pass the `$_FILES` array intact to the object, and it then munges data to do what is necessary. Developers then do not actually interact with `$_FILES` directly, nor even with the files as present in the request instance; they just pass the values along to objects that treat that as their domain.

Thoughts?

simensen:
@weierophinney My assumption was that if we could not handle this directly in the spec that we could create Uploaded File utilities like I'm trying to do w/ Cookies and like what you suggested at the bottom of your message there.

I do not believe that leaving our $_FILES representation looking exactly like $_FILES will hinder interoperability. It does mean that it is yet one more thing the frameworks will have to write (or find let commodity third-party implementations take over) but I think that is different than hindering interoperability.

That said, if we could find a way to do this that is not complicated/threatens the spec or making it easier to misunderstand/abuse (right now our files handling is exactly the same as $_FILES, I think, so much less chance to implement it "wrong"...) I would not be opposed to seeing it added. I do enjoy not having to care about these things when working with Symfony's Request. In that case I'd want the uploaded file interface to look as much like $_FILES as possible.

In all cases, if we end up introducing this to the spec I think it would require us to go back to draft.

proposed/http-message.md
webmozart: You are right that this can be a tree. The same is actually true for `getQueryParams()` and `getParsedBody()`.

Do you have any proposal how to improve the wording?

webmozart:
> First, what you describe looks more like a tree, and not an array of objects

Yes. The same is true for `getQueryParams()` and `getParsedBody()`. I don't know how far we should go into describing the tree semantics for these three methods.

> The next question I have is whether we need an interface for the file upload details

We can also specify the structure of the array. But that requires a formal, verbal description (for which I foresee a lot of discussions about the exact wording), while an interface describes a structured value without ambiguities per se.

> I wonder if we should not do similarly here, and keep the specification as it currently is.

We could. While this doesn't hinder interoperability as such, it makes the usage of file uploads in a PSR-7 request unnecessarily (IMO) complicated. The solution above, OTOH, seems fairly simple to me.

I'll leave that for you to decide.

proposed/http-message.md
weierophinney: > The same is actually true for getQueryParams() and getParsedBody().

Except that these are *always* unstructured, which is a very different situation. `$_FILES` has a well-documented (if sometimes vexing!) structure.

In this case, I think we can only have a return typehint of `array`, and would have to indicate that it returns either an array of `UploadedFileInterface` instances, or a nested tree of them, as a description for the return value annotation.

weierophinney:
> > I wonder if we should not do similarly here, and keep the specification as it currently is.
> We could. While this doesn't hinder interoperability as such

Except that it does. Any existing libraries that expect unfiltered `$_FILES` data would now need to be altered to accept *either* that structure *or* the PSR-7 structure and the interfaces it entails. I can think of several such libraries that would be affected.

Considering we do not make changes to query parameters or cookies, leaving the `$_FILES` data raw and unfiltered would be more consistent with the rest of the input parameters described in `ServerRequest`. We can document in the meta document that the expectation is to pass the value on to other processors if nested/hierarchical names are used, similarly to how we already indicate complex headers should be approached.

weierophinney:
@webmozart I've discussed this matter with @pmjones and @simensen , and at this time do not recommend merging the pull request. As noted elsewhere in the comments, other server-side input parameters are expected in the same structure in which the superglobal represents them (with the exception of body parameters, as parsing of those will depend on the content-type and HTTP method). Having `$_FILES` diverge from that removes interoperability with existing libraries that expect the raw format, and also means new libraries written targeting PSR-7 would not be able to be used with barebones PHP. 

We *do* recognize that accessing nested file upload values is not straight-forward, but feel that this is the realm for dedicated libraries to handle.

If desired, I can open a pull request to outline that decision in the meta document. Until then, @simensen and/or @pmjones : ready to close.

webmozart:
> Having $_FILES diverge from that removes interoperability with existing libraries

I don't understand this argument. How can existing "interoperability with existing libraries" be an argument in a PSR that has not even been published?

> If desired, I can open a pull request to outline that decision in the meta document.

That would be wise.

weierophinney:
> How can "interoperability with existing libraries" be an argument in a PSR that has not even been published?

Because existing libraries already target `$_FILES` and/or accept it as an argument. PSR-7 just becomes the source for that value for those libraries in the future.

weierophinney:
@webmozart I've been mulling this over overnight, and just now discussing it more with @simensen . We're thinking the following:

- Accept the PR…
- but also keep `getFileParams()` for those that want/require the raw `$_FILES` structure…
- and detail in a narrative section of the spec the "why" behind this, what the expectations are for implementors, and how developers would consume the feature (all of which you've done in the PR overview and comments, but it needs to be in the spec).

If you want, I can fork the PR and do the above; let me know how you want to proceed.

webmozart:
@weierophinney Sounds good! Sure, go ahead. :)

webmozart:
@weierophinney btw I'm also fine if `getFileParams()` simply returns a restructured array, without adding `getUploadedFiles()` and the interface. However then we need to put sufficient thought into the description of how to restructure the array, which might prove difficult.

weierophinney:
@webmozart I've opened #498, which incorporates your commits from this PR, adds the changes I indicated previously, and adds narrative for both the specification and meta document.

weierophinney:
@simensen and/or @pmjones — please close in favor of #498, which incorporates and builds on this request (it retains the original commits from @webmozart).

PR #493, opened 2015-04-07
HTTP: Alternative way to address consistency issues mentioned in #491

pmjones:
This PR suggests the following changes:

- `getHeader($name)` returns the header values as set by the user; it does not return a concatenated string.

- `getHeaders()` is now essentially the equivalent of calling `getHeader()` for every header in the message; it does not return an array of header line strings.

- `getHeaderLine($name)` is a new method that takes the place of the old `getHeader()`, where the return value is expected to be an entire header line with the values concatenated into a comma-separated string.

- `getHeaderLines()` has been removed.

This PR has no equivalent for the old `getHeaderLines($name)` behavior that would return an array of multiple lines for a single header, each with a separate value for that header. The user can mimic that behavior with something like:

    $lines = [];
    foreach ($message->getHeader($name) as $values) {
        foreach ($values as $value) {
            $lines[] = "{$name}: {$value}";
        }
    }

Finally, this further enforces the idea that `getHeader()` or `getHeaders()` should always return an array for the header values themselves.

I think this is both more flexible and more consistent than the previous version, and addresses both my own gripes and those expressed in #491.

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -398,15 +398,6 @@ interface MessageInterface
     public function getHeader($name);
 
     /**
-     * Retrieves all header lines as array of strings.
-     *
-     * If there are no headers, this method should return an empty array.
-     *
-     * @return string[]
-     */
-    public function getHeaderLines();
-
-    /**
      * Retrieves the line for a single header, with the header values as a
      * comma-separated string.
      *
@@ -550,21 +541,6 @@ interface RequestInterface extends MessageInterface
      * @return array|null
      */
     public function getHeader($name);
-
-    /**
-     * Extends MessageInterface::getHeaderLines() to provide request-specific
-     * behavior.
-     *
-     * This method acts exactly like MessageInterface::getHeaderLines(), with
-     * one behavioral change: if the Host header is requested, but has
-     * not been previously set, the method MUST attempt to pull the host
-     * segment of the composed URI, if present.
-     *
-     * @see MessageInterface::getHeaderLines()
-     * @see UriInterface::getHost()
-     * @return string[]
-     */
-    public function getHeaderLines();
 
     /**
      * Extends MessageInterface::getHeaderLines() to provide request-specific
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -350,16 +350,16 @@ interface MessageInterface
      * Retrieves all message header values.
      *
      * The keys represent the header name as it will be sent over the wire, and
-     * each value is the string, or array of strings, associated with the header.
+     * each value array of strings associated with the header.
      *
      *     // Represent the headers as a string
      *     foreach ($message->getHeaders() as $name => $values) {
-     *         echo $name . ": " . implode(", ", (array) $values);
+     *         echo $name . ": " . implode(", ", $values);
      *     }
      *
      *     // Emit headers iteratively:
      *     foreach ($message->getHeaders() as $name => $values) {
-     *         foreach ((array) $values as $value) {
+     *         foreach ($values as $value) {
      *             header(sprintf('%s: %s', $name, $value), false);
      *         }
      *     }
@@ -368,8 +368,8 @@ interface MessageInterface
      * exact case in which headers were originally specified.
      *
      * @return array Returns an associative array of the message's headers. Each
-     *     key MUST be a header name, and each value MUST the value(s) for that
-     *     header.
+     *     key MUST be a header name, and each value MUST be the array of values
+     *     for that header.
      */
     public function getHeaders();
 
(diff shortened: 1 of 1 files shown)


harikt:
Good one Paul :) .

webmozart:
Looks good.

However, I preferred if getHeader()/getHeaders() always return arrays of strings, even if there is just one header. While the flexible return type can be nice, it's also prone to errors if you forget to cover both cases (returns string vs returns string[]) in the calling code.

pmjones:
> I preferred if getHeader()/getHeaders() always return arrays of strings, even if there is just one header.

(/me nods)

I can compromise on that point if the remainder seems reasonable.

pmjones:
To further illuminate my preference for the string|array return, I was thinking of the user who calls `$message->withHeader('Content-Type', 'text/plain')` in one area, and then to inspect it later, calls `$message->getHeader('Content-Type')`. It seems a bit much to have to expect back an array when the user set a string. However, I like I said, I can compromise on that point if it's the only objection.

weierophinney:
@webmozart and @pmjones : Again, I want to point out that the current names are the subject of lengthy discussion. The most relevant thread is:

- https://groups.google.com/d/msg/php-fig/TMlKWlEiUWI/qOkpztmCz8QJ

@webmozart I'm going to be blunt here: I don't care if you _prefer_ a particular convention. What matters here are the use cases, which are outlined in that thread, and which I've outlined before. They are as follows:

- The most common use case when retrieving a single header is to receive the entire header at once as a string. This is why `getHeader()` returns a string value.
- The most common use case when retrieving all headers at once is to emit headers, either from a client or from a server. In these cases, you will want the individual values so that you can choose to concatenate them correctly based on the header, or emit the discrete values as discrete lines, separately. As such, `getHeaders()` returns the structure it currently does.
- `getHeaderLines()` exists for those headers where comma-concatenation will not work. This allows the specification to be specific in what `getHeader()` returns, while allowing consumers to operate on those headers that have non-standard concatenation.

We specify that header values are stored as an array internally, to allow for multiple values for any given header. This is an important point, as it ensures consistency of return value for `getHeaderLines()` and `getHeaders()`; you do not need to test if the value is a string OR an array; you know it's an array of strings.

If we absolutely **must** make a change, I'm more aligned with this particular proposal. However, this still presents a fairly large change, as existing libraries building against the current specification will need to change to work with the changes. I think this falls outside the realm of "review" at this point, and as such, I'm very much against making this change, when the docblocks clearly indicate what the return values are, and we can, if necessary, address the reasons why within the meta document, linking to the aforementioned mailing list thread.

pmjones:
> https://groups.google.com/d/msg/php-fig/TMlKWlEiUWI/qOkpztmCz8QJ

Yup; my dissatisfaction with the CSV representation track back at least that far. ;-)  This PR expands on that thread to allow for the same use-cases under what I think are more reasonable names, addressing the consistency issues that I and @webmozart raised.  E.g., Dowling's comment about `getHeader()` early in that thread is addressed by `getHeaderLine()` and `getHeaderLines()`.

> this still presents a fairly large change

Yes, to my own additional dismay.

> existing libraries building against the current specification will need to change to work with the changes

I get that; however, as the PSR was only "proposed" and not "accepted", that might be considered a reasonable risk for those libraries to have undertaken.

> I think this falls outside the realm of "review" at this point, and as such, I'm very much against making this change

I'm not especially thrilled about that prospect either. The cost in time would be 2 weeks of "draft" and another 2 weeks of "review". But I also think it's the right thing to do.

You (and Beau) are of course free to disregard this suggestion, and again, my apologies for the Johnny-come-lately nature of the PR.

pjdietz:
> I preferred if getHeader()/getHeaders() always return arrays of strings, even if there is just one header.
I put together PR #494 to see what this would look like.

webmozart:
Thanks @weierophinney for linking that topic. From what I see, the result of the discussion was that the names `getHeaders()`+`getHeader()`+`getHeaderAsArray()` were rejected, together with an assessment of the typical use cases.

The proposal in this PR (and #494) is quite different though, and I don't think that it conflicts with the discussion you linked.

weierophinney:
@pmjones The reason I'm preferring #494 to this one is that I think part of the confusion is over the semantics of `getHeaderLine(s)?`; the impression for those just reading the method name is that these methods would return string values, when currently, and in your own PR, they return an array.

pmjones:
#493 has been updated per this conversation.

proposed/http-message.md
```diff
@@ -542,7 +538,7 @@ interface RequestInterface extends MessageInterface
      * @see MessageInterface::getHeader()
      * @see UriInterface::getHost()
      * @param string $name Case-insensitive header field name.
-     * @return string
+     * @return array|null
```
GrahamCampbell: should this be `string[]|null`?

proposed/http-message.md
```diff
@@ -383,34 +384,30 @@ interface MessageInterface
...
-     * and supply your own delimiter when concatenating.
+     * This method returns an array of all the header values of the given
+     * case-insensitive header name.
      *
      * If the header did not appear in the message, this method should return
      * a null value.
      *
      * @param string $name Case-insensitive header field name.
-     * @return string|null
+     * @return array|null
```
GrahamCampbell: should this be `string[]|null`?

proposed/http-message.md
weierophinney: No; the idea is that it returns a consistent type (array), so you can always use the return value for iteration.

proposed/http-message.md
weierophinney: See above.

PR #494, opened 2015-04-07
Alternative way to address consistency issues mentioned in #491 and #493

pjdietz:
Another variant of PR #491 and #493 with specific return types.

This PR suggests the following changes:

- `getHeader($name)` takes the place of the current `getHeaderLines($name)` and returns an array of strings (even with only one).
- `getHeaders()` is unchanged and returns an array with arrays of strings for values.
- `getHeaderLine($name)` takes the place of the current `getHeader($name)` and returns a (concatenated) string.
- `getHeaderLines()` returns an array with strings as values.

This provides a predictable interface where the user can know the return type in advance and won't need to test whether the method returned a string or array.

It also adds consistency where both:

```
$message->getHeaders()[$k] === $message->getHeader($k)
```

And

```
$message->getHeaderLines()[$k] === $message->getHeaderLine($k)
```

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -386,7 +386,7 @@ interface MessageInterface
      * Retrieves a header by the given case-insensitive name as an array of
      * strings.
      *
-     * If the header did not appear in the message, this method should return an
+     * If the header did not appear in the message, this method MUST return an
      * empty array.
      *
      * @param string $name Case-insensitive header field name.
@@ -405,7 +405,7 @@ interface MessageInterface
      * comma concatenation. For such headers, use getHeader() instead
      * and supply your own delimiter when concatenating.
      *
-     * If the header did not appear in the message, this method should return
+     * If the header did not appear in the message, this method MUST return
      * a null value.
      *
      * @param string $name Case-insensitive header field name.
@@ -414,7 +414,8 @@ interface MessageInterface
     public function getHeaderLine($name);
 
     /**
-     * Retrieves all message headers.
+     * Retrieves all message headers as an array of header names and string
+     * values.
      *
      * The keys represent the header name as it will be sent over the wire, and
      * each value is a string of the header values concatenated together using
@@ -566,7 +567,9 @@ interface RequestInterface extends MessageInterface
      * Extends MessageInterface::getHeaderLines() to provide request-specific
      * behavior.
      *
-     * Retrieves a header by the given case-insensitive name as a string.
+     * This method returns all of the header values of the given
+     * case-insensitive header name as a string concatenated together using
+     * a comma.
      *
      * This method acts exactly like MessageInterface::getHeaderLines(), with
      * one behavioral change: if the Host header is requested, but has
@@ -584,7 +587,8 @@ interface RequestInterface extends MessageInterface
      * Extends MessageInterface::getHeaderLines() to provide request-specific
      * behavior.
      *
-     * Retrieves all message headers.
+     * Retrieves all message headers as an array of header names and string
+     * values.
      *
      * This method acts exactly like MessageInterface::getHeaderLines(), with
      * one behavioral change: if the Host header has not been previously set,
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -383,6 +383,18 @@ interface MessageInterface
     public function hasHeader($name);
 
     /**
+     * Retrieves a header by the given case-insensitive name as an array of
+     * strings.
+     *
+     * If the header did not appear in the message, this method should return an
+     * empty array.
+     *
+     * @param string $name Case-insensitive header field name.
+     * @return string[]
+     */
+    public function getHeader($name);
+
+    /**
      * Retrieve a header by the given case-insensitive name, as a string.
      *
      * This method returns all of the header values of the given
@@ -390,7 +402,7 @@ interface MessageInterface
      * a comma.
      *
      * NOTE: Not all header values may be appropriately represented using
-     * comma concatenation. For such headers, use getHeaderLines() instead
+     * comma concatenation. For such headers, use getHeader() instead
      * and supply your own delimiter when concatenating.
      *
      * If the header did not appear in the message, this method should return
@@ -399,18 +411,22 @@ interface MessageInterface
      * @param string $name Case-insensitive header field name.
      * @return string|null
      */
-    public function getHeader($name);
+    public function getHeaderLine($name);
 
     /**
-     * Retrieves a header by the given case-insensitive name as an array of strings.
+     * Retrieves all message headers.
      *
-     * If the header did not appear in the message, this method should return an
-     * empty array.
+     * The keys represent the header name as it will be sent over the wire, and
+     * each value is a string of the header values concatenated together using
+     * a comma.
      *
-     * @param string $name Case-insensitive header field name.
-     * @return string[]
+     * While header names are not case-sensitive, getHeaderLines() will preserve
+     * the exact case in which headers were originally specified.
+     *
+     * @return array Returns an associative array of the message's headers. Each
+     *     key MUST be a header name, and each value MUST be a string.
      */
-    public function getHeaderLines($name);
+    public function getHeaderLines();
 
     /**
      * Return an instance with the provided header, replacing any existing
@@ -542,7 +558,7 @@ interface RequestInterface extends MessageInterface
      * @see MessageInterface::getHeader()
      * @see UriInterface::getHost()
      * @param string $name Case-insensitive header field name.
-     * @return string
+     * @return string[]
      */
     public function getHeader($name);
 
@@ -550,7 +566,7 @@ interface RequestInterface extends MessageInterface
      * Extends MessageInterface::getHeaderLines() to provide request-specific
      * behavior.
      *
-     * Retrieves a header by the given case-insensitive name as an array of strings.
+     * Retrieves a header by the given case-insensitive name as a string.
      *
      * This method acts exactly like MessageInterface::getHeaderLines(), with
      * one behavioral change: if the Host header is requested, but has
@@ -560,9 +576,27 @@ interface RequestInterface extends MessageInterface
      * @see MessageInterface::getHeaderLines()
      * @see UriInterface::getHost()
      * @param string $name Case-insensitive header field name.
-     * @return string[]
+     * @return string
      */
-    public function getHeaderLines($name);
+    public function getHeaderLine($name);
+
+    /**
+     * Extends MessageInterface::getHeaderLines() to provide request-specific
+     * behavior.
+     *
+     * Retrieves all message headers.
+     *
+     * This method acts exactly like MessageInterface::getHeaderLines(), with
+     * one behavioral change: if the Host header has not been previously set,
+     * the method MUST attempt to pull the host segment of the composed URI, if
+     * present.
+     *
+     * @see MessageInterface::getHeaderLines()
+     * @see UriInterface::getHost()
+     * @return array Returns an associative array of the message's headers. Each
+     *     key MUST be a header name, and each value MUST be a string.
+     */
+    public function getHeaderLines();
 
     /**
      * Retrieves the message's request target.
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -386,7 +386,7 @@ interface MessageInterface
      * Retrieves a header by the given case-insensitive name as an array of
      * strings.
      *
-     * If the header did not appear in the message, this method MUST return an
+     * If the header does not appear in the message, this method MUST return an
      * empty array.
      *
      * @param string $name Case-insensitive header field name.
@@ -405,7 +405,7 @@ interface MessageInterface
      * comma concatenation. For such headers, use getHeader() instead
      * and supply your own delimiter when concatenating.
      *
-     * If the header did not appear in the message, this method MUST return
+     * If the header does not appear in the message, this method MUST return
      * a null value.
      *
      * @param string $name Case-insensitive header field name.
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -390,7 +390,7 @@ interface MessageInterface
      * empty array.
      *
      * @param string $name Case-insensitive header field name.
-     * @return string[] An array of discrete string values for the given header.
+     * @return string[] An array of string values for the given header.
      *    If the header does not appear in the message, this method MUST
      *    return an empty array.
      */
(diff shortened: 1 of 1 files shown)


webmozart:
:+1:

weierophinney:
@codeguy and @mtdowling — I think it looks good, but would like your opinion as well. I think this clarifies the method names and intentions quite well. It *does* represent a change, however, and as two others who are currently coding implementations, want your feedback.

Thanks!

codeguy:
:+1:

proposed/http-message.md
```diff
@@ -383,6 +383,18 @@ interface MessageInterface
     public function hasHeader($name);
 
     /**
+     * Retrieves a header by the given case-insensitive name as an array of
+     * strings.
+     *
+     * If the header did not appear in the message, this method should return an
```
mtdowling: If the header does not appear in the message, this method MUST return an empty array.

weierophinney:
In talking with @mtdowling just now, we want to point out that the following would not be true:

```php
$message->getHeader($k) === $message->getHeaders()[$k];
```

**unless** the header key `$k` matches the casing used to inject the header. (Not that this would be a show-stopper, but it should be noted.)

proposed/http-message.md
```diff
@@ -542,7 +558,7 @@ interface RequestInterface extends MessageInterface
      * @see MessageInterface::getHeader()
      * @see UriInterface::getHost()
      * @param string $name Case-insensitive header field name.
-     * @return string
+     * @return string[]
```
mtdowling: Can you add a description of the return type? e.g., "Returns an array of strings"

proposed/http-message.md
```diff
@@ -550,7 +566,7 @@ interface RequestInterface extends MessageInterface
      * Extends MessageInterface::getHeaderLines() to provide request-specific
      * behavior.
      *
-     * Retrieves a header by the given case-insensitive name as an array of strings.
+     * Retrieves a header by the given case-insensitive name as a string.
```
mtdowling: It might be good to repeat some of the verbiage in getHeaderLines() that describes the return value (a concatenated string, yadda, yadda...)

proposed/http-message.md
```diff
@@ -399,18 +411,22 @@ interface MessageInterface
      * @param string $name Case-insensitive header field name.
      * @return string|null
      */
-    public function getHeader($name);
+    public function getHeaderLine($name);
 
     /**
-     * Retrieves a header by the given case-insensitive name as an array of strings.
+     * Retrieves all message headers.
```
mtdowling: Maybe something more explicit to help people understand why this method exists: Retrieves all message headers as an array of header name to string values.

pjdietz:
@weierophinney Ah, good catch. Still, it's closer. Also not true if `$k` did not appear since `getHeader($k)` will be an empty array and `getHeaders()[$k]` will be an undefined index.

weierophinney:
@pjdietz Can you address the more targetted feedback from @mtdowling? I think with those changes, we can likely merge.

Thanks!

mtdowling:
This looks good.

I wonder if there's a good reason to have `getHeaderLines()` though. `getHeaderLine($k)` in this PR is useful because it concatenates values together into a string case-insensitively, allowing easy access to headers across the whole message case-insensitively. `getHeaderLines()` isn't that useful because it only concats values together of the same case-insensitive name. It's far less useful unless we complicate the method a bit... For example: returns a normalized list of header names to values (where normalized means it flattens out case-insensitivity across all of the headers). I don't know if it would be worth it at that point.

pjdietz:
@weierophinney and @mtdowling will do!

weierophinney:
@mtdowling I've been unsure about the plural form for retrieving string values, too. I'm only entertaining it because we've had three separate PRs now trying to make for symmetric methods for singular/plural forms.

proposed/http-message.md
weierophinney: Seconded. :)

pjdietz:
I've added the clarification @mtdowling suggested in a new commit. I'm not sure how useful `getHeaderLines()` is either. I liked the direction @pmjones was going with #493 but wanted to avoid having a method return strings for some headers and arrays for others.

mtdowling:
I like this PR the most out of all of the current PR's on the table. I do think that since none of us see the value in `getHeaderLines()` that it should be removed. I think that change would make this PR stand out even more as the best approach.

pmjones:
Not "none." ;-) Example forthcoming.

pmjones:
To wit:

```php
<?php
    foreach ($message->getHeaderLines() as $line) {
        header($line);
    }
```

wolfy-j:
This is incorrect example, $line will contain header value but not name, plus it will corrupt such headers as Set-Cookie and etc.

pjdietz:
@pmjones That's a more useful getHeaderLines() than in this PR. I missed what you had done for that method in #493.
```
   /**
     * Retrieves all header lines as array of strings.
     *
     * If there are no headers, this method should return an empty array.
     *
     * @return string[]
     */
    public function getHeaderLines();
```

weierophinney:
@pmjones and @pjdietz: That's not a great example. First, it should likely be this:

```php
foreach ($message->getHeaderLines() as $header => $value) {
    header(sprintf('%s: %s', $header, $value));
}
```

Second, it will break on some headers, as not all headers can be safely comma-concatenated. The specification indicates that if a header can allow multiple values, comma-concatenation or multiple headers are allowed, but does not specify that they MUST allow comma-concatenation. In particular, `Set-Cookie` and `Cookie` famously do follow that recommendation (they use `;` as a delimiter). This is why @mtdowling and I have been arguing it's a useless method.

pjdietz:
Although, yeah, cookies and such won't work.

pmjones:
@pjdietz If the primary difference between #493 and #494 is the standardization of returning arrays, that's something I can modify in #493 very easily.

pmjones:
@weierophinney Noted -- so then how does one get a list of all header names that are in the message? `$names = array_keys($message->getHeaders();` and then `foreach ($names as $name) { header(sprintf('%s: %s', $name, $message->getHeaderLine($name))); }` ?

pmjones:
And honestly I would expect implementations to do the sane thing and have `getHeaderLines()` refer back to `getHeaderLine()` internally -- but perhaps that is naively optimistic on my part.

wolfy-j:
@pmjones you can use `getHeaders()` directly to get header name and it's all lines, @weierophinney has great example in his library.

webmozart:
@weierophinney Just for the sake of completeness, could you give some examples where comma-concatenation is not valid?

pmjones:
@wolfy-j In this and in #493, `getHeaders()` returns an array of values, not a string and not lines.

weierophinney:
@webmozart Already did: `Cookie` and `Set-Cookie`.

wolfy-j:
@pmjones, i think #493 does change `getHeaders()` behaviour, this PR does not:
> getHeaders() is unchanged and returns an array with arrays of strings for values.

pmjones:
Cf. the newly modified #493; arrays are now returned consistently.

weierophinney:
> @weierophinney Noted -- so then how does one get a list of all header names that are in the message? $names = array_keys($message->getHeaders(); and then foreach ($names as $name) { header(sprintf('%s: %s', $name, $message->getHeaderLine($name))); } ?

@pmjones I'm not even sure what you're trying to do in that example. `getHeaderLines()` as proposed here returns key/value pairs, where the key is the header name, the value the concatenated string. That's why I suggested:

```php
foreach ($message->getHeaderLines() as $header => $value) {
    // ...
}
```

But yes: you would get the list of headers that are currently present using either `array_keys($message->getHeaderLines())` or `array_keys($message->getHeaders()`, and each would also return the header names in the case with which they were registered.

pmjones:
@wolfy-j : Ah so! In which case, I recommend against this PR. I'd like to see something that returns the header values as set, *and* something that returns lines.  In #493, getHeader() and getHeaders() return the array of values; getHeaderLine() returns a single header line concatenated, and getHeaderLines() returns an array of those lines.

wolfy-j:
@pmjones, so we ended with decision `getHeader()` vs `getHeaderLine()`, there is no other differences in this PRs.

pjdietz:
What's the use case for concatenation? Which headers would not work if `getHeaderLines()` returned an array of strings with an item for each header value? For example, if there are three "Set-cookie" values, the method would return:

```
[
    "Set-cookie: ...", 
    "Set-cookie: ...", 
    "Set-cookie: ...", 
    "X-foo: bar"
]
```

pmjones:
Apparently some folks like the idea of concatenated return values. #493 takes that into account, while also allowing access to the array of values as set.

weierophinney:
@pmjones See my comment on your own PR; I see it exactly the opposite. What this PR does is clarify the "Line" verbiage, which, semantically, implies strings, but is currently (with `getHeaderLines()`) returning an array. This PR changes the interface to return what you would expect based on the method name.

The only part where I see that as maybe *not* being the case is `getHeader()`, which would now return an array. However, since it is mirrored by `getHeaders()`, and returns the same structure, this makes sense.

The bigger question is whether we need both `getHeaderLine()` and `getHeaderLines()`. As @mtdowling and I have argued, the latter seems like a relatively dangerous idea, as RFC 7230 does not require that a comma be used for concatenation. We have stipulated that in our string versions for consistency and to ease the burden on implementors (as custom header values may have their own requirements, too); as such, having an array of key/string pairs is not terribly useful.

weierophinney:
@pjdietz What you just described is the current behavior of `getHeaderLines($name)`… Bringing us full-circle back to the current implementation.

wolfy-j:
@pjdietz, i can't see it specified in your original post, even more:
> `$message->getHeaderLines()[$k] === $message->getHeaderLine($k)`

pjdietz:
@weierophinney , not quite. What I'm thinking is `getHeaderLines()` with no argument returns the full "{name}: {value}" for every header. Something you could use with this:

```
    foreach ($message->getHeaderLines() as $line) {
        header($line);
    }
```

pjdietz:
But where no concatenation takes place.

pjdietz:
@wolfy-j Sorry for the confusion. That's not in this PR. Just throwing that out as an idea.

webmozart:
@pjdietz I think this method is too specific. You can easily solve this use case with `getHeaders()`.

I think that `getHeaderLines()` can safely be removed then. Apart from that, I like this PR.

wolfy-j:
@webmozart :+1:

weierophinney:
@pjdietz Eww. No. Again, not a sane solution, due to the problem of concatenation. The concatenation issue is why `getHeaders()` returns the structure it does. You can see how that is consumed here:

- https://github.com/phly/http/blob/master/src/Server.php#L205-L216

Let's go with the following:

- `getHeader($name)` should return the array of values, and an empty array if no values are present or header is missing.
- `getHeaders()` should remain unchanged.
- Add `getHeaderLine($name)`, which would provide a comma-concatenated value for the named header.
- Remove `getHeaderLines()` entirely.

At that point, the most common use cases are present, and the verbiage issues are fixed.

@pmjones Can you live with that? And are you understanding why I think `getHeaderLines()` should be removed, particularly with the link made above?

pjdietz:
I like removing `getHeaderLines()` as well as it doesn't really seem that useful and doesn't get you anything you can't get otherwise.

With that removed, this PR mostly just changes the name of the current `getHeaderLines($name)` to `getHeader($name)` and the current `getHeader($name)` to `getHeaderLine($name)`. It's a slight change, but seems a bit more consistent.

pmjones:
I can live with getting rid of getHeaderLines().

weierophinney:
@pjdietz Only thing remaining is to clarify where you have `@return string[]` on `getHeader()` to indicate "An array of discrete string values for the given header" or similar as a description of the return value. Otherwise :+1:.

pjdietz:
```
    /**
     * Retrieves a header by the given case-insensitive name as an array of
     * strings.
     *
     * If the header does not appear in the message, this method MUST return an
     * empty array.
     *
     * @return string[] An array of discrete string values for the given header.
     */
    public function getHeader($name);
```
We have language above indicating an empty array for no header. Do you want to see that in the @return line as well?

pmjones:
FWIW #493 has been updated per this conversation.

weierophinney:
> We have language above indicating an empty array for no header. Do you want to see that in the @return line as well?

Yes, please.

pjdietz:
@weierophinney All set. I added corresponding language for `getHeaderLine()` as well.

weierophinney:
:ship:

proposed/http-message.md
```diff
@@ -383,6 +383,20 @@ interface MessageInterface
...
 
     /**
+     * Retrieves a header by the given case-insensitive name as an array of
+     * strings.
+     *
+     * If the header does not appear in the message, this method MUST return an
+     * empty array.
+     *
+     * @param string $name Case-insensitive header field name.
+     * @return string[] An array of discrete string values for the given header.
```
mtdowling: I'm not sure what you mean by discrete here. I think it muddies the language unless clarified. For example, getHeader() should not parse each header in order to separate out CSV. If a user calls `setHeader('foo', ['bar, baz', 'bam'])`, then calling `getHeader('foo')` should return `['bar, baz', 'bam']`.

pjdietz:
@mtdowling Yes, `['bar, baz', 'bam']` is what I'd expect, too. I think discrete because you get an array item for each value. @weierophinney, was that your intent?

weierophinney:
@pjdietz and @mtdowling that was my intent. Perhaps the clarification should be "An array of string values as provided for the given header."

pjdietz:
Whoops, just saw the rest of your line. New one coming in a sec.

weierophinney:
:ship:

mtdowling:
:shipit: :boat:

weierophinney:
One last note (for @pjdietz and @pmjones): there are examples under section 1.2 that reference `getHeader()` and `getHeaderLines()` that need to be updated. Can one of you update those, or shall I open a separate pull request on merge?

pjdietz:
Sure, I can take care of it. Heading out now, but look for it in about an hour.

pmjones:
@pjdietz I have merged #493 as the basis for #494; please update from the new version, and then we can merge this PR on top of it.

pjdietz:
Ok, I've added a couple new commits. @pmjones, #493 was missing some of the return value descriptions worked out in the comments on this thread, so I added those back in. Also, section 1.2 should now be up to date.

weierophinney:
@pjdietz changes look good; thanks for making the changes to section 1.2!

@pmjones and/or @simensen — ready to merge!

pjdietz:
Glad to help!

PR #495, opened 2015-04-08
Add more flexibility to headers

geggleto:
Would like to see an optional separator as sometimes I don't want to use commas. Kind of nit-picking... but would be nice.

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -387,7 +387,7 @@ interface MessageInterface
      *
      * This method returns all of the header values of the given
      * case-insensitive header name as a string concatenated together using
-     * a comma.
+     * a specified seperator.
      *
      * NOTE: Not all header values may be appropriately represented using
      * comma concatenation. For such headers, use getHeaderLines() instead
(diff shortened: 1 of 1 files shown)


proposed/http-message.md
```diff
@@ -399,7 +399,7 @@ interface MessageInterface
      * @param string $name Case-insensitive header field name.
      * @return string|null
      */
-    public function getHeader($name);
+    public function getHeader($name, $separator = ',');
```
GrahamCampbell: this needs to be added to the phpdoc too

GrahamCampbell:
:-1: on the idea

weierophinney:
:-1:

We specifically chose to only allow `,` to reduce the number of possible implementation paths. If you want to implement an alternate separator, you can retrieve all values for a header and pass the value to `implode()`. Once #493/#494 are merged, this will look like:

```php
$concatenated = implode(';', $message->getHeader($name));
```

mtdowling:
:-1:

weierophinney:
@pmjones and/or @simensen — please close.

PR #496, opened 2015-04-08
[PSR-7] Added section on file uploads

weierophinney:
This adds a section on why `getFileParams()` is tied to the `$_FILES` superglobal, and the vision for how to handle file uploads by consumers.

See #492 for more information

### DIFF
--- a/proposed/http-message-meta.md
+++ b/proposed/http-message-meta.md
@@ -613,6 +613,75 @@ Examples of this practice already exist in libraries such as
 has functionality for casting the value to a string, these objects can be
 used to populate the headers of an HTTP message.
 
+### What about handling arrays of file uploads?
+
+When submitting a form with file inputs named `file[0]` and `file[1]`, the
+contents of `$_FILES` becomes:
+
+```php
+array( 
+    'file' => array(
+        'name' => array(
+            0 => 'file0.txt',
+            1 => 'file1.txt',
+        ) ,
+        'type' => array(
+            0 => 'text/plain',
+            1 => 'text/html',
+        ),
+        /* and so on */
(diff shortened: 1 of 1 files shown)


weierophinney:
Pinging @simensen to review.

proposed/http-message-meta.md
```diff
@@ -613,6 +613,75 @@ Examples of this practice already exist in libraries such as
...
+as each specific file detail (e.g., content type) is nested with those of any
+sibling uploads.
+
+We considered specifying that `getFileParams()` MUST parse the values into a
+nested structure that mimics how they were provided in the form; however, we
+feel:
+
+- It introduces inconsistency in the API. Other parameters, such as query string
+  and cookie parameters, are specified such that they MUST retain the
+  structure of their corresponding superglobals. (Body parameters are a
```
webmozart: I disagree with this statement. While the `$_POST` and `$_GET` arrays are usable by default, the `$_FILES` array isn't. Being consistent with `$_FILES` is not a desirable goal for me.

Two major PHP frameworks, i.e. ZF and Symfony, fix the `$_FILES` array in their implementations. Probably more. I think that it is irresponsible to ignore this issue in the PSR.

weierophinney: In ZF, we don't handle it at the request level, though. We handle it in our `Zend\File\Transfer ` component, and the related filters and validators. Those are fed the raw `$_FILES` structure, which the request object encapsulates. (I've actually been blissfully unaware of how we do this until investigating for this issue.)

If anything, that validates the approach suggested here.

weierophinney:
Closing in favor of #492.

PR #498, opened 2015-04-09
[PSR-7] Additional upload files functionality

weierophinney:
This pull request is based on (and incorporates commits from) #492.

In addition to the changes there, it presents the following:

- Clarifications to the `UploadedFileInterface` and `ServerRequestInterface`, including re-addition of `getFileParams()` as a canonical, non-mutable source for upload file data.
- Addition of section 1.6 to the specification, describing "Uploaded files".
- Addition of two "design decision" sections to the metadocument:
 - "Why are both getFileParams() and getUploadedFiles() present?"
 - "Why does getUploadedFiles() return objects instead of arrays?"

### DIFF
--- a/proposed/http-message-meta.md
+++ b/proposed/http-message-meta.md
@@ -606,7 +606,7 @@ represented by `$_FILES`. This data is useful in several cases:
   - https://github.com/siriusphp/upload
 - As a canonical source of file upload information that cannot be mutated.
 
-`getUploadedParams()` exists to fix a known issue with how PHP structures the
+`getUploadedFiles()` exists to fix a known issue with how PHP structures the
 `$_FILES` superglobal when array notation (e.g., `files[0]`, `files[1]`, and so
 on) is used to name file uploads. It can be _derived_ from the return value of
 `getFileParams()`, but presents a structure that is more user friendly, as you
--- a/proposed/http-message-meta.md
+++ b/proposed/http-message-meta.md
@@ -655,6 +655,8 @@ used to populate the headers of an HTTP message.
 * Michael Dowling
 * Larry Garfield
 * Evert Pot
-- Tobias Schultze
+* Tobias Schultze
+* Bernhard Schussek
+* Anton Serdyuk
 * Phil Sturgeon
 * Chris Wilkinson
--- a/proposed/http-message-meta.md
+++ b/proposed/http-message-meta.md
@@ -594,6 +594,32 @@ specific, and the results of detection can be easily injected into objects that
 need it, and/or calculated as needed using utility functions and/or classes from
 the `RequestInterface` instance itself.
 
+### Why are both getFileParams() and getUploadedFiles() present?
+
+`getFileParams()` exists to provide the raw file upload data, typically as
+represented by `$_FILES`. This data is useful in several cases:
+
+- A number of existing libraries exist that already can process the `$_FILES`
+  array; having a method that returns that raw value allows them to interoperate
+  with this specification.
+  - https://github.com/codeguy/Upload
+  - https://github.com/siriusphp/upload
+- As a canonical source of file upload information that cannot be mutated.
+
+`getUploadedParams()` exists to fix a known issue with how PHP structures the
+`$_FILES` superglobal when array notation (e.g., `files[0]`, `files[1]`, and so
+on) is used to name file uploads. It can be _derived_ from the return value of
+`getFileParams()`, but presents a structure that is more user friendly, as you
+can access the various uploads using the same tree structure in which the files
+were submitted.
+
+### Why does getUploadedFiles() return objects instead of arrays?
+
+`getUploadedFiles()` returns a tree of `Psr\Http\Message\UploadedFileInterface`
+instances. This is done primarily to simplify specification: instead of
+requiring paragraphs of implementation specification for an array, we can
+specify an interface.
+
 ### What about "special" header values?
 
 A number of header values contain unique representation requirements which can
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -1639,6 +1639,16 @@ interface UploadedFileInterface
     public function getTemporaryPath();
     
     /**
+     * Retrieve the file size.
+     *
+     * Implementations SHOULD return the value stored in the "size" key of
+     * the file in the $_FILES array, OR the value returned by filesize().
+     *
+     * @return int|null The file size in bytes or null if none was provided.
+     */
+    public function getSize();
+    
+    /**
      * Retrieve the error associated with the uploaded file.
      *
      * The return value MUST be one of PHP's UPLOAD_ERR_XXX constants.
@@ -1668,21 +1678,6 @@ interface UploadedFileInterface
      *     was provided.
      */
     public function getClientFilename();
-    
-    /**
-     * Retrieve the size sent by the client.
-     *
-     * Do not trust the value returned by this method. A client could send
-     * a wrong or malicious size with the intention to corrupt or hack your
-     * application.
-     *
-     * Implementations SHOULD return the value stored in the "size" key of
-     * the file in the $_FILES array.
-     *
-     * @return int|null The size sent by the client in bytes or null if none
-     *     was provided.
-     */
-    public function getClientSize();
     
     /**
      * Retrieve the mime type sent by the client.
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -816,11 +816,6 @@ interface ServerRequestInterface extends RequestInterface
      * request. They SHOULD be injected during instantiation, such as from PHP's
      * $_FILES superglobal, but MAY be derived from other sources.
      *
-     * The implementation SHOULD ensure that is_uploaded_file() returns true for
-     * the paths of the uploaded files. However, this recommendation MAY be
-     * ignored in special cases, for example when using temporary files as
-     * uploaded files during testing.
-     *
      * @return UploadedFileInterface[] The uploaded file(s), if any.
      */
     public function getUploadedFiles();
@@ -1531,6 +1526,12 @@ interface UploadedFileInterface
      *
      * Implementations SHOULD return the value stored in the "tmp_name" key
      * of the file in the $_FILES array.
+     *
+     * The implementation MUST guarantee that the file pointed to by the path
+     * can safely be used by the calling code. More precisely, the
+     * implementation MUST ensure that is_uploaded_file() returns true for
+     * the path if the path was provided through the $_FILES array or a
+     * similar mechanism.
      *
      * @return string The absolute path to the uploaded file.
      */
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -1527,6 +1527,9 @@ interface UploadedFileInterface
      * Implementations SHOULD return the value stored in the "tmp_name" key
      * of the file in the $_FILES array.
      *
+     * Implementations MUST guarantee that is_uploaded_file() returns true
+     * for the returned path.
+     *
      * @return string The absolute path to the uploaded file.
      */
     public function getPath();
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -296,6 +296,85 @@ application-specific rules (such as path matching, scheme matching, host
 matching, etc.). As such, the server request can also provide messaging between
 multiple request consumers.
 
+### 1.6 Uploaded files
+
+`ServerRequestInterface` specifies two methods for retrieving uploaded file
+metadata. `getFileParams()` is intended as a canonical source for the raw data
+normally present in PHP's `$_FILES` superglobal.
+
+The `$_FILES` superglobal has some well-known problems when dealing with arrays
+of file inputs. As an example, if you have a form that submits an array of files
+— e.g., the input name "files", submitting `files[0]` and `files[1]` — PHP will
+represent this as:
+
+```php
+array(
+    'files' => array(
+        'name' => array(
+            0 => 'file0.txt',
+            1 => 'file1.html',
+        ),
+        'type' => array(
+            0 => 'text/plain',
+            1 => 'text/html',
+        ),
+        /* etc. */
+    ),
+)
+```
+
+instead of the expected:
+
+```php
+array(
+    'files' => array(
+        0 => array(
+            'name' => 'file0.txt',
+            'type' => 'text/plain',
+            /* etc. */
+        ),
+        1 => array(
+            'name' => 'file1.html',
+            'type' => 'text/html',
+            /* etc. */
+        ),
+    ),
+)
+```
+
+The result is that consumers need to know this language implementation detail,
+and write code for gathering the data for a given upload.
+
+As such, `ServerRequestInterface` defines an additional method,
+`getUploadedFiles()`, intended for providing a normalized structure.
+Implementations are expected to:
+
+- Aggregate all information for a given file upload, and use it to populate a
+  `Psr\Http\Message\UploadedFileInterface` instance.
+- Re-create the submitted tree structure, with each leaf being the appropriate
+  `Psr\Http\Message\UploadedFileInterface` instance for the given location in
+  the tree.
+
+Because the uploaded files data is derivative (derived from `$_FILES` or the
+data returned by `getFileParams()`), a mutator method, `withUploadedFiles()`, is
+also present in the interface, allowing delegation of the normalization to
+another process.
+
+In the case of the above examples, consumption resembles the following:
+
+```php
+$file0 = $request->getUploadedFiles()['files'][0];
+$file1 = $request->getUploadedFiles()['files'][1];
+
+printf(
+    "Received the files %s and %s",
+    $file0->getClientFilename(),
+    $file1->getClientFilename()
+);
+
+// Received the files file0.txt and file1.html
+```
+
 ## 2. Package
 
 The interfaces and classes described are provided as part of the
@@ -807,7 +886,7 @@ interface ServerRequestInterface extends RequestInterface
     public function withQueryParams(array $query);
 
     /**
-     * Retrieve the uploaded files.
+     * Retrieve the uploaded files in raw $_FILES structure.
      *
      * This method MUST return the file upload metadata in the same structure
      * as PHP's $_FILES superglobal.
@@ -816,7 +895,23 @@ interface ServerRequestInterface extends RequestInterface
      * request. They SHOULD be injected during instantiation, such as from PHP's
      * $_FILES superglobal, but MAY be derived from other sources.
      *
-     * @return UploadedFileInterface[] The uploaded file(s), if any.
+     * @return array Upload file(s) metadata, if any; an empty array MUST be
+     *     returned if no data is present.
+     */
+    public function getFileParams();
+
+    /**
+     * Retrieve normalized file upload data.
+     *
+     * This method returns upload metadata in a normalized tree, with each leaf
+     * an instance of Psr\Http\Message\UploadedFileInterface.
+     *
+     * These values MAY be prepared on-demand from the composed file parameters
+     * as exposed by getFileParams(), MAY be injected during instantiation, or
+     * MAY be injected with withUploadedFiles().
+     *
+     * @return array An array tree of UploadedFileInterface instances; an empty
+     *     array MUST be returned if no data is present.
      */
     public function getUploadedFiles();
 
@@ -826,10 +921,10 @@ interface ServerRequestInterface extends RequestInterface
      * These MAY be injected during instantiation.
      *
      * This method MUST be implemented in such a way as to retain the
-     * immutability of the message, and MUST return a new instance that has the
+     * immutability of the message, and MUST return an instance that has the
      * updated body parameters.
      *
-     * @param UploadedFileInterface[] $uploadedFiles The uploaded files.
+     * @param array An array tree of UploadedFileInterface instances.
      * @return self
      */
     public function withUploadedFiles(array $uploadedFiles);
@@ -1514,7 +1609,7 @@ namespace Psr\Http\Message;
  *
  * Instances of this interface are considered immutable; all methods that
  * might change state MUST be implemented such that they retain the internal
- * state of the current instance and return a new instance that contains the
+ * state of the current instance and return an instance that contains the
  * changed state.
  */
 interface UploadedFileInterface
@@ -1548,7 +1643,8 @@ interface UploadedFileInterface
      * Implementations SHOULD return the value stored in the "error" key of
      * the file in the $_FILES array.
      *
-     * @return int One of the UPLOAD_ERR_XXX constants.
+     * @see http://php.net/manual/en/features.file-upload.errors.php
+     * @return int One of PHP's UPLOAD_ERR_XXX constants.
      */
     public function getError();
     
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -1611,6 +1611,14 @@ namespace Psr\Http\Message;
  * might change state MUST be implemented such that they retain the internal
  * state of the current instance and return an instance that contains the
  * changed state.
+ *
+ * When used in an SAPI environment where $_FILES is populated, files referenced
+ * by instances SHOULD be moved using is_uploaded_file() and
+ * move_uploaded_file() to ensure permissions and upload status are verified
+ * correctly.
+ *
+ * @see http://php.net/is_uploaded_file
+ * @see http://php.net/move_uploaded_file
  */
 interface UploadedFileInterface
 {
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -1615,22 +1615,20 @@ namespace Psr\Http\Message;
 interface UploadedFileInterface
 {
     /**
-     * Retrieve the path to the uploaded file.
+     * Retrieve the (temporary) path to the uploaded file.
      *
-     * This method MUST return an absolute file path.
+     * This method MUST return an absolute file path; the path is assumed
+     * to be temporary, per the typical PHP upload mechisms.
      *
      * Implementations SHOULD return the value stored in the "tmp_name" key
      * of the file in the $_FILES array.
      *
      * The implementation MUST guarantee that the file pointed to by the path
-     * can safely be used by the calling code. More precisely, the
-     * implementation MUST ensure that is_uploaded_file() returns true for
-     * the path if the path was provided through the $_FILES array or a
-     * similar mechanism.
+     * can safely be used by the calling code.
      *
      * @return string The absolute path to the uploaded file.
      */
-    public function getPath();
+    public function getTemporaryPath();
     
     /**
      * Retrieve the error associated with the uploaded file.
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -396,7 +396,7 @@ operations will work regardless of environment. In particular:
 - `move($path)` is provided as a safe and recommended alternative to calling
   `move_uploaded_file()` directly on the temporary upload file. Implementations
   will detect the correct operation to use based on environment.
-- `getStream()` will return a `StreamableInterface` instance. In non-SAPI
+- `getStream()` will return a `StreamInterface` instance. In non-SAPI
   environments, one proposed possibility is to parse individual upload files
   into `php://temp` streams instead of directly to files; in such cases, no
   upload file is present. `getStream()` is therefore guaranteed to work
@@ -406,7 +406,6 @@ As examples:
 
 ```
 // Move a file to an upload directory
-$uuid = create_a_uuid();
 $filename = sprintf(
     '%s.%s',
     create_uuid(),
@@ -415,7 +414,9 @@ $filename = sprintf(
 $file0->move(DATA_DIR . '/' . $filename);
 
 // Stream a file to Amazon S3.
-// Assume $s3wrapper is a PHP stream that will write to S3.
+// Assume $s3wrapper is a PHP stream that will write to S3, and that
+// Psr7StreamWrapper is a class that will decorate a StreamInterface as a PHP
+// StreamWrapper.
 $stream = new Psr7StreamWrapper($file1->getStream());
 stream_copy_to_stream($stream, $s3wrapper);
 ```
@@ -1647,7 +1648,7 @@ interface UploadedFileInterface
     /**
      * Retrieve a stream representing the uploaded file.
      *
-     * This method MUST return a StreamableInterface instance, representing the
+     * This method MUST return a StreamInterface instance, representing the
      * uploaded file. The purpose of this method is to allow utilizing native PHP
      * stream functionality to manipulate the file upload, such as
      * stream_copy_to_stream() (though the result will need to be decorated in a
@@ -1656,7 +1657,7 @@ interface UploadedFileInterface
      * If the move() method has been called previously, this method MUST raise
      * an exception.
      *
-     * @return StreamableInterface Stream representation of the uploaded file.
+     * @return StreamInterface Stream representation of the uploaded file.
      * @throws \RuntimeException in cases when no stream is available or can be
      *     created.
      */
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -355,10 +355,10 @@ Implementations are expected to:
   `Psr\Http\Message\UploadedFileInterface` instance for the given location in
   the tree.
 
-Because the uploaded files data is derivative (derived from `$_FILES` or the
-data returned by `getFileParams()`), a mutator method, `withUploadedFiles()`, is
-also present in the interface, allowing delegation of the normalization to
-another process.
+Because the uploaded files data is derivative (derived from `$_FILES`, the data
+returned by `getFileParams()`, or the request body), a mutator method,
+`withUploadedFiles()`, is also present in the interface, allowing delegation of
+the normalization to another process.
 
 In the case of the above examples, consumption resembles the following:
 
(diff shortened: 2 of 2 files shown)


weierophinney:
Pinging @webmozart, @simensen, and @pmjones to review.

proposed/http-message-meta.md
```diff
@@ -594,6 +594,32 @@ specific, and the results of detection can be easily injected into objects that
...
+represented by `$_FILES`. This data is useful in several cases:
+
+- A number of existing libraries exist that already can process the `$_FILES`
+  array; having a method that returns that raw value allows them to interoperate
+  with this specification.
+  - https://github.com/codeguy/Upload
+  - https://github.com/siriusphp/upload
+- As a canonical source of file upload information that cannot be mutated.
+
+`getUploadedParams()` exists to fix a known issue with how PHP structures the
```
pmjones: getUploadedParams() should be getUploadedFiles() ?

weierophinney:
Thanks for the catch, @pmjones — fixed!

pmjones:
I think I'd prefer it if the UploadedFileInterface method names reflected more directly the original array key names; e.g., getTmpFile(), getError(), etc.  I get that we have a chance here to give what we think are "better" or "more clear" names, but the least-contentious names I can think of are the original key names.

weierophinney:
@pmjones The names threw me as well. @webmozart — you okay with that change?

webmozart:
@pmjones @weierophinney I'm uneasy about this. I have to look up the original names in the manual everytime I use them ("tmp_name" vs. "name"?), since I don't use them very often - and I guess nobody does. Using clearer method names seems to have a big benefit here.

You also chose different names for the request getters than the original PHP array names (e.g. `getQueryParams()` instead of `getGetParams()`) for the same reason. I'd like you to consider this here as well.

At last, there is a security aspect to the modified names. Calling the methods `getClientName()`, `getClientSize()` etc. lessens the chance of using them without sanitization by accident, or at least it becomes easier to find such vulnerable code snippets. If they're simply called `getName()` et al., such vulnerabilities are easily overlooked.

webmozart:
More review tomorrow, it's late here. Thanks for your work @weierophinney!

weierophinney:
> I have to look up the original names in the manual everytime I use them ("tmp_name" vs. "name"?), since I don't use them very often - and I guess nobody does. Using clearer method names seems to have a big benefit here.

The irony here is that I actually have memorized the `$_FILES` structure, which is why the proposed names confused me.

However:

> At last, there is a security aspect to the modified names. Calling the methods getClientName(), getClientSize() etc. lessens the chance of using them without sanitization by accident, or at least it becomes easier to find such vulnerable code snippets.

The above makes a lot of sense to me, and I'm happy to keep the names as-is based on that. We may want to indicate in the spec or the metadocument the rationale for the new names, however.

simensen:
@pmjones @weierophinney I think that the name changes make sense and as @webmozart mentioned we changed the names of other things. I'd like to take a look at the names again tomorrow to see if there are other changes that might be able to be made. In general I'm :+1: for leaving them as-is, though, as I don't see a need to try and replicate the somewhat obscure snake_case names like tmp_name in a camelCaseWorld.

pmjones:
If we're changing them, then at least let's make them even more clear. Something like getRaw*() or getUnsafe*() seem better than getCient*() if we want the name to be a warning.

pmjones:
And regarding the name getQuery() vs getGet(), that doesn't strike me as an appropriate comparison; the repetition alone was a reason to change it.

pmjones:
Hell, on further thought, if we are going to rename things to say what came from the client so it's obviously regarded as unsafe, then we need to rename getPost() etc. to getClientPost() etc. for consistency. Which I am against doing. ;-)

m00t:
Please, review some thoughts against this PR's changes described in that issue: https://github.com/php-fig/fig-standards/issues/501

wolfy-j:
@pmjones, Request and file upload names seems unrelated (btw, there is no `getPost()` method, there is `getParsedBody()` one, unless i'm missing something). `getClient*` in this particular case is not about "data sent by client", but about value (name, mimetype and size) filled with data from client computer, so it seems reasonable to separate them from `getPath`, `getError` which are generated by PHP. 
I would only recommend to add `getSize` method related to real upload size (fetched by PHP) as this one of the most often use cases.

pmjones:
> is not about "data sent by client", but about value (name, mimetype and size) filled with data from client computer,

The query params, body params, etc. also came from the client computer.

wolfy-j:
> The query params, body params, etc. also came from the client computer.
Yes, entirelly. File upload contain both generated and client values in one object.

pmjones:
Here's the short version of my argument: If we are going to note *here* in the method name that the data came from the client, we should mention in *every* method name that the data came from client *across the entire interface*. If we're not going to do that everywhere, there's no reason to do it here. Obviously I don't think we need to do it everywhere, and I don't think we need to do it here either.

wolfy-j:
But you still OK to rename methods, like proposed by you `getUnsafe*`? Is it just about "client" word or about renaming itself?

pmjones:
As you can see from later posts, I'd really rather leave them closer to their original key names.

simensen:
If I'm getting data from POST (first name, last name, whatever) I expect that to be sent from the client and can trust it accordingly (read: I shouldn't). They can quite literally send whatever they want.

If I'm looking at a class called "Uploaded File" and it has "getFileSize" it might not be immediately be clear that it was the file size that the client sent vs the actual file size that the implementing class verified to be true.

From the method names, as given, it is more clear to me that the path and the error are trustworthy sources of information ( getPath, getError ) as that information will be provided by PHP. On the other hand the other fields ( getClientFilename, getClientSize, getClientMimeType ) was information provided by the client that PHP has no way to verify as to be true.

I think the distinction is important but I do not know if that means the method names themselves should indicate this difference. This could be a documentation issue ( on getSize, getMimeType, etc ) and normalize the names.

As far as getPath, I think that keeping things closer to the original (along the lines as @pmjones suggested) might be a good idea because it might be easy to assume that getPath is safe to rely on *after* the fact when in reality the path is to a temporary file. Something like getTemporaryPath might be more appropriate in that case.

weierophinney:
I've renamed `getPath()` to `getTemporaryPath()` per the suggestion of @simensen. Additionally, I've removed the verbiage about `is_uploaded_file()` as a validation for the temporary path, per discussion with @m00t on #501.

wolfy-j:
What about adding `getSize()`?

webmozart:
@weierophinney I think `getTemporaryPath()` is a good name. It indicates that the file should be moved somewhere else.

I was coming from the point that someone might want to add a method like `move()` to an implementation, which would move the file somewhere else and update the path. But that would violate immutability anyway and hence is bogus (this consideration could be added to the meta doc).

@wolfy-j I'm not sure whether we should any other methods (`getSize()`, ...) to the interface. The current methods allow to access all information that you need. If we start adding convenience methods, where do we stop?

I still think `getClient*()` are the best available method names. The big difference to POST data is that POST data is sent by the client per definition. The `UploadedFileInterface`, OTOH, mixes server data and client data. A clear distinction should be made for security reasons.

wolfy-j:
Sure, i was expecting such feedback as at may require to add `getMimetype` and etc. Agreed.

webmozart:
A last point I forgot is `is_uploaded_file()`: I think it is important that we cover this, so that users of the interface can use simple file functions instead of `move_uploaded_file()` (see #501). For example:

```php
// works for "real" uploaded files, fails for mock files
move_uploaded_file($uploadedFile->getTemporaryPath(), $dest);

// works for both
// is_uploaded_file() should be called for real uploaded files
move($uploadedFile->getTemporaryPath(), $dest);
```

I think that the implementation (real vs. mock) should decide whether it is possible to call `is_uploaded_file()`. For security reasons, however, the interface should dictate that this method must be called if possible.

pmjones:
ALternatively, rename getPath() to getSafePath(); by implication, everything else is unsafe.

weierophinney:
@pmjones : 

> ALternatively, rename getPath() to getSafePath(); by implication, everything else is unsafe.

Except that it's not a "safe" path when working in an SAPI environment; if the file is not moved out of that path, it will be GC'd eventually. As such, "temporary" is likely the best verbiage here, and mirrors `tmp_name` adequately..

m00t:
@webmozart according to this comment http://stackoverflow.com/a/16277045 move_uploaded_file() does not provide any additional security now. Is it true?

weierophinney:
@m00t Are you and I reading the same comment? He specifically notes:

> Technically unnecessary might still be a good idea: we are talking security, why not be extra safe?

and recommends still using it today when possible.

weierophinney:
@webmozart Added some verbiage to the `UploadedFileInterface` class-level docblock; please review.

m00t:
@weierophinney I dont think that
> why not be extra safe
is a good argument. I can not imagine any attack to be possible when I use move() instead of move_uploaded_file(). Does it exist?

weierophinney:
@m00t Read the various comments to http://php.net/move_uploaded_file

Among them are notes in differences of the user/group used and/or whether the directory or file permissions are considered when performing `move_uploaded_file()` vs `rename()`. These are important security considerations, and are a reasonable rationale for using the former vs the latter *when* *possible*.

pmjones:
All right, last comment then I'll leave it alone. If you're all dead set on having the "client" notation in the method name (which I am still against) at least make the method names symmetrical: it's not "Temporary" in that case, it's "Server" (as vs "Client" in the others).

weierophinney:
@pmjones The path is *not* symmetrical with the other metadata, and doesn't need to be treated as such in the naming.

The rationale for the "client" verbiage is to _signal_ to the _user_ about the origin of the value (the implication being that it's _untrusted_ input that was provided by the _client_). In other words, the naming is signaling _intent_.

When we look at the path, we have to consider what PHP is signaling. The corresponding segment in `$_FILES` is `tmp_name`, and it indicates where PHP is _temporarily_ storing the file. The fact that it's on the server is an environment detail; the key detail is the intent: it's temporary, and you cannot rely on the file remaining in that location between requests (as it may be removed via garbage collection or other processes).

Naming the method `getTemporaryPath()` provides a _signal_ to the _user_ about the information they are consuming: it's a path, and not a permanent one, and thus they'll need to decide if they want to keep the file (which will involve moving it out of that path).

While the names do not have a 1:1 correspondence with the components of `$_FILES`, I'd argue that they show _intent_ better.

m00t:
@weierophinney forgive me for providing such a short answer for your long comment about abstractions in #501. I read it several times in order to ensure I understand you properly. And I do not agree with you here. 

**In short**: this abstraction works fine when you have control over both ServerRequestInterface implementation and client code that uses this interface. But then you do not need a standard. When you do not have control over one of this two parts, abstraction does not work unless you put in standard phrase "client code MUST NOT use neither move_uploaded_file() nor is_uploaded_file()".

Please, give me a couple of days, I will shape my thoughts in a clear manner about this issue.

weierophinney:
@m00t What do you mean by "client code" here? This interface is specifically the `ServerRequestInterface`, and is only intended for handling an incoming request, server-side. Perhaps this is the source of the difference of opinion?

pmjones:
@weierophinney I'm not convinced, but I'm also clearly in the minority. I think all the arguments thus far have been wrong, but I'm not the decider.

m00t:
@weierophinney When you have interface, you actually have two sides: implementation of this interface and client code which is using objects which implement this interface. In my opinion, the main purpose of standard interfaces like PSR-7 is to separate client code from interface implementation, so that they can be written by developers who are completely unaware of each other's existence.

m00t:
@pmjones This is a bit subjective opinion, but I agree with @weierophinney about method names. For me, `getClientMimeType()` and `getClientFilename()` sounds just perfect: "get filename which client provides to me". I am not sure about `getClientSize()`, btw. Is it the size that browser really provides to the server, or this is calculated by PHP itself? May be this should be just `getSize()`?

weierophinney:
@m00t I'm still not understanding your point. Let's assume that I know the purpose of interfaces. The point of `getFileParams()` is to provide a convenience wrapper around `$_FILES`, which PHP developers are already aware of. They don't need to know how that gets populated, only that it is.

In most cases, you'll not be marshaling the `ServerRequestInterface` instance yourself; you'll write code that receives it. You can then consume it however you need. When dealing with file uploads, you have three ways to do that:

- `getFileParams()` to retrieve the `$_FILES` structure; *or*
- `getUploadFiles()` to retrieve the normalized `$_FILES` structure, and work with `UploadedFileInterface` instances; *or*
- the message body directly.

All of these are valid choices, and you'll choose based on the needs of your application.

If you choose one of the first two, the interface is suggesting that implementations are responsible for populating them. As a consumer, you can then rely on that data being present. In the case of the latter, the interface indicates that it, too, must be populated by the implementation, so again, the consumer can rely on that data being present.

weierophinney:
> I am not sure about getClientSize(), btw. Is it the size that browser really provides to the server, or is it calculated by PHP itself? Maybe this should be just getSize()?

That really is the size as provided by the browser/client; the value is the whatever is specified in the `Content-Length` header on that particular MIME part. This is why it is considered suspect, and all security guides indicate you should not rely on the value.

m00t:
@weierophinney as far as I understand, `getClientSize()` is not. It is actually calculated mime part size:
* https://github.com/php/php-src/blob/master/main/rfc1867.c#L1070
* https://github.com/php/php-src/blob/master/main/rfc1867.c#L1264

So it should be renamed to `getSize()` in my opinion.

weierophinney:
@m00t Thanks for the link to the source; I've updated to reflect this.

pmjones:
Again, this is why I think the best, least-contentious names are the original key names. We're now back to 3/5 of them being the original names; I continue to suggest we go the whole way and use `getName()` and `getType()`.

m00t:
@weierophinney Lest assume we have this client code:
```php
function update_user_action(ServerRequestInterface $request) {
    $avatar = $request->getFiles()['avatar'];
    move_uploaded_file($avatar->getTemporaryName(), AVATARS_DIR.'/'.$avatar->getClientName());

    // do other stuff

}
```
Is this code correct? This code is 100% compatible with current PSR-7, but it is incorrect in my opinion. Because there can be `ServerRequestImplementation` which do not use SAPI `$_FILES` functionality. For example, it can be some kind of `MockServerRequest`. 
Lets rewrite this code. I see several possible ways to make this client code correct:
```php
function update_user_action(ServerRequestInterface $request) {
    $avatar = $request->getFiles()['avatar'];
    if ($avatar instanceof SAPIUploadedFileInterface) {
        move_uploaded_file($avatar->getTemporaryName(), AVATARS_DIR.'/'.$avatar->getClientName());
    } else {
        rename($avatar->getTemporaryName(), AVATARS_DIR.'/'.$avatar->getClientName());
    }

    // do other stuff

}
```
```php
function update_user_action(ServerRequestInterface $request) {
    $avatar = $request->getFiles()['avatar'];
    if ($avatar->isSAPIUploadedFile()) {
        move_uploaded_file($avatar->getTemporaryName(), AVATARS_DIR.'/'.$avatar->getClientName());
    } else {
        rename($avatar->getTemporaryName(), AVATARS_DIR.'/'.$avatar->getClientName());
    }

    // do other stuff

}
```
```php
function update_user_action(ServerRequestInterface $request) {
    $avatar = $request->getFiles()['avatar'];
    rename($avatar->getTemporaryName(), AVATARS_DIR.'/'.$avatar->getClientName());

    // do other stuff

}
```
```php
function update_user_action(ServerRequestInterface $request) {
    $avatar = $request->getFiles()['avatar'];
    $avatar->move($avatar->getClientName());

    // do other stuff

}
```
All four possibilities change PSR-7 in some way and 100% correct in my opinion. Changes which should be done to PSR-7 to make them be PSR-7 compatible:

1. Introduce `SAPIUploadedFileInterface`.
2. Introduce `isSAPIUploadedFile()` method.
3. Forbid using `move_uploaded_file()` by clients by introducing `clients MUST NOT use move_uploaded_file()` statement in standard.
4. Remove `getTemporaryName()` and introduce `move()` method which will use `move_uploaded_file()` or `rename()` decided by implementation itself.

I personally **strongly** prefer variant 4 and think this is the best option. I do not agree with @webmozart  that `move()` method will violate immutability. It will not if we introduce it that way: `move()` method can be called once per file and should not update file path. By this way calling `move()` method will be 100% identical to calling `move_uploaded_file()` or `rename()` functions, but will be much more safe, because client will know less about `UploadedFileInterface` implementation and this will make client code and standard itself much more simple in my opinion.

m00t:
Another problem with files at their current state in my opinion: streams. I am not sure whether it should be separate issue or not.
Lets assume again that we need possibility to upload large file to S3. Client code will be something like that:
```php
function upload_large_file_action(ServerRequestInterface $request) {
   $stream = get_stream_for_file($request, 'large_file_key');
   pipe_stream_to_s3($stream);
}
```
How would we write `get_stream_for_file()` function? Now we have 1 possibility: 
```php
function get_stream_for_file(ServerRequestInterface $request, $fileKey) {
    $requestBodyStream = $request->getBody();
    return get_rfc1867_mime_parameter_stream_from_body($requestBodyStream(), $fileKey);
}
```
Where `get_rfc1867_mime_parameter_stream_from_body()` is some function which use something like `MultipartContentParser` from zf-content-negotiation. And every client should have its own `get_rfc1867_mime_parameter_stream_from_body()` function, may be use some library for that or what. I think this functionality should be incorporated to PSR-7 this way:
```php
function get_stream_for_file(ServerRequestInterface $request, $fileKey): StreamableInterface {
    return $request->getFiles()[$fileKey]->getBody();
}
```
So new method `getBody(): StreamableInterface` should be added to `UploadedFileInterface` in my opinion.

m00t:
I reread this thread again. Actually, I think I said in my comment about `move_uploaded_file()` with 4 variants https://github.com/php-fig/fig-standards/pull/498#issuecomment-91711058 absolutely the same @webmozart said in his comment about `move_uploaded_file()` https://github.com/php-fig/fig-standards/pull/498#issuecomment-91579507 except our disagreement about introducing the `move()` method. Right?

weierophinney:
@m00t — Option 1 is do-able right now without changes, but would require the user to depend on a concretion (i.e., the library implementing PSR-7 would need to define the interface and/or a concrete implementation to test against).

Based on the arguments you've outlined, I definitely see your point now, and think that likely the best path forward is adding the `move()` method; implementations can then decide if they are in an SAPI environment or not.

m00t:
@weierophinney 
> Option 1 is do-able right now without changes, but would require the user to depend on a concretion

Client SHOULD NOT (or MUST NOT?) depend on a concretion, right? Depending on concretion makes interface and this standard irrelevant.

weierophinney:
> How would we write get_stream_for_file() function? Now we have 1 possibility:

There's another possibility, assuming the implementation seeds the file metadata:

```php
function get_stream_for_file(ServerRequestInterface $request, $fileKey) {
    $tempFile = $request->getUploadFiles()[$fileKey]->getTemporaryFile();
    return fopen($tempFile);
}
```

Since uploaded files are already written to disk under a typical SAPI, and likely written to disk with utilities such as found in zf-content-negotiation when non-POST or not under an SAPI, the above should work fine.

The harder part would be if you want to skip the step of pushing the file to disk. In that particular case, I'd expect to write a stream filter implementation that iterates over the request body until it finds a MIME part that fulfills the specified key; once it does, it would start piping the data, ending when the MIME boundary is found. Any data after that would be ignored. This can be done using the native PHP stream functions, though, and it's relatively easy to wrap the message body in StreamWrapper for this purpose.

weierophinney:
> Client SHOULD NOT (or MUST NOT?) depend on a concretion, right? Depending on concretion makes interface and this standard irrelevant.

Right. Which is why I agreed that option 4 is the best approach. ;-)

weierophinney:
One point, @m00t : I want to keep `getTemporaryFile()`, but ALSO add `move()`. The reason is as outlined in my answer about streams above: there are times when you may not want to keep the file, but instead stream it elsewhere.

m00t:
@weierophinney 
> there are times when you may not want to keep the file, but instead stream it elsewhere

That is why I suggest introducing `UploadedFileInterface::getBody(): StramableInterface` method and remove `getTemporaryFile()`. `getBody()` will have some advantages over `getTemporaryFile()` in my opinion.

You can have non-SAPI implementations of PSR-7. For example, you can write php webserver in react-php or as php extension. Non-SAPI PSR-7 implementation could use algorithm you described here

> In that particular case, I'd expect to write a stream filter implementation that iterates over the request body until it finds a MIME part that fulfills the specified key; once it does, it would start piping the data, ending when the MIME boundary is found. Any data after that would be ignored.

For example, we stick with `getBody()`:
* Lets say we have a SAPI implementation. `$file->getBody()` will simply return `fopen($this->tmpName)`.
* Now lets say we have non-SAPI implementation. `$file->getBody()` will use this algorithm.

If we stick with `getTemporaryName()`:
* With SAPI implementation we call `fopen($file->getTemporaryName()` which is conceptually identical to calling `$file->getBody()` (despite that with `getBody()` we will have `StreamableInterface` instead of php's stream.
* But with non-SAPI implementation when we call `fopen($file->getTemporaryName()` something terrible happens: first, this implementation dumps all stream to file and returns its name. After that we create stream from this file.

We still can use `$request->getBody()`, but this stream will be constructed artificially by implementation when we use SAPI, because `php://input` stream is empty when SAPI receives `POST` request with `multipart/form-data` content-type. Therefore clients SHOULD NOT use `$request->getBody()` to receive raw content of `multipart/form-data` `POST` request.

m00t:
@weierophinney @webmozart I reread #492 also. And I think @weierophinney mised the point of @webmozart about interoperability. I think this is important.

>>> Having $_FILES diverge from that removes interoperability with existing libraries

>> How can "interoperability with existing libraries" be an argument in a PSR that has not even been published?

> Because existing libraries already target $_FILES and/or accept it as an argument. PSR-7 just becomes the source for that value for those libraries in the future

PSR-7 will not be compatible with existing libraries which use `move_uploaded_file()` anyway. Arguments for that are the same as here: https://github.com/php-fig/fig-standards/pull/498#issuecomment-91711058. It is impossible to make existing libraries PSR-7 compatible, so this goal should not be an argument for any design decision. Existing libraries just will be incompatible with PSR-7, and this is normal.

m00t:
I rewrote this comment https://github.com/php-fig/fig-standards/pull/498#issuecomment-91727135 to be more clear.

weierophinney:
@m00t —

> PSR-7 will not be compatible with existing libraries which use move_uploaded_file() anyway. Arguments for that are the same as here: #498 (comment). It is impossible to make existing libraries PSR-7 compatible, so this goal should not be an argument for any design decision. Existing libraries just will be incompatible with PSR-7, and this is normal.

You missed 2 things:

- The new `UploadedFileInterface` is only intended for use with `getUploadFiles()`. That particular method is the intended one when dealing with non-SAPI uploads, non-POST uploads, and/or when dealing with upload arrays. *This will be the recommended way to work with PSR-7*.
- `getFileParams()` is intended to contain the raw `$_FILES` structure. If that structure is unavailable, it MUST return an empty array. *This method exists only to allow usage with legacy code.*

While PSR-7 *is* a new specification, and, of course, existing code will need to be rewritten in order to use it, there are also many utility libraries that already exist that can be fed data from PSR-7 instances. I outline several such utilities that accept `$_FILES` and/or a compatible array in this patch, in the meta-document.. By including `getFileParams()`, developers can continue to use these libraries, and PSR-7 just becomes the source of the data.

My point is: adopting PSR-7 should not necessarily require a full rewrite, and should enable gradual refactoring of an application to make complete use of the spec.

weierophinney:
@m00t Per your suggestions, I'm working on changes now to implement `move()` and to change `getTemporaryFile()` to `getStream()` (which will now return either a `StreamableInterface` or a PHP stream; still trying to decide what the best approach may be).

weierophinney:
m00t: I've added `getStream()` and `move()` to the `UploadedFileInterface` at this time. `getStream()` returns a PHP stream; `move()` is written to indicate it should use the appropriate functionality for writing the stream to disk based on the environment (and specifically noted move_uploaded_file(), rename(), and stream operations as being viable).

I kept `getTemporaryName()` to allow use with existing code, but have indicated that consumers should use `getStream()` and/or `move()` for code portability.

Finally, I added more narrative to section 1.6 to discuss why those methods exist.

@webmozart  and @simensen — can you also review, please?

m00t:
@weierophinney Let's say client code use `getFileParams()` method. How `update_user_action()` should be written with this method?
```php
function update_user_action(ServerRequestInterface $request) {
    move_uploaded_file($request->getFileParams()['avatar']['tmp_name'], $request->getFileParams()['avatar']['name']);
    // update other info
}
```
This way? How we can write `MockServerRequest` to test this method?
Is it OK that client code may be not compatible with `MockServerRequest`?

m00t:
The main goal of proposed `move()` and `getBody()` (`getStream()`) methods is to replace `getFileParams()` and `getTemporaryName()`, so that **nobody never** will write `move_uploaded_file()` in client code. If nobody will write `move_uploaded_file()`, then **every client code will be guaranteed to work with any PSR-7 implementation**.

This is what I mean by interoperability and I think it is important: **every PSR-7 client MUST be compatible with every SPR-7 implementation**.

If some clients with legacy code will use `move_uploaded_file()`, then there is no sense in introducing the `move()` method.

And then we should admit this statement: PSR-7 does not guarantee your implementation will work with any client code. I am having a very hard time to admit it, because for me this should be the N1 standard goal.

m00t:
I would prefer to stick with `getBody(): StreamableInterface` because of these

1. We already have `MessageInterface::getBody(): StreamableInterface`, so having new `UploadedFileInterface::getBody(): StreamableInterface` seems a very consistent to me.
2. There are some advantages of `StreamableInterface` compared to native php streams: http://mtdowling.com/blog/2014/07/03/a-case-for-higher-level-php-streams/

What are arguments for using native streams instead of `StreamableInterface`?

m00t:
@weierophinney Finally. I've got your point. So you understand  that there may be PSR-7 clients which are incompatible with introduced `MockServerRequest`, but you think this is OK, because the use of `getFileParams()` is deprecated. Yes?

I think these legacy PSR-7 clients will never move to non-deprecated functionality, and we will have legacy code with brand new standard from its born. So we should not address them in favor of progress. But it is my subjective opinion, I may be wrong here.

Also, if we decide to address them, then I think `getTemporaryName()` should be removed. Because we already have `getFilesParams()` method to be used in legacy clients, so we will have a clear and understandable separation: `getUploadedFiles()` for cool brand new clients and `getFilesParams()` for boring old legacy clients.

kamazee:
OK, I'd like to make some induction and to ask you, gentlemen, to question it. The real issue here is not lack of some methods, but rather having some redundant ones. Namely, those that let client see some parts of `$_FILES`.

What is the reason behind exposing `$_FILES` (as a whole or just its items) *as is*? Meta states that it gives compatibility with some existing libraries that can work with what PHP spits out into `$_FILES`. While it looks like a gain at a glance -- is there a gain, really? These libraries rely on the fact that files are really uploaded -- so, they do need a `$_FILES` filled by PHP (not faked by a test) anyway (at least because [they](https://github.com/codeguy/Upload/blob/6dd8917c1549cfaea2dbfc612e676ea7b78adc83/src/Upload/Storage/FileSystem.php#L107) [use](https://github.com/siriusphp/upload/blob/1fe2cf7bb7e33d5311601158568493bd174bc2ed/src/Container/Local.php#L97) `move_uploaded_file`).
So, PSR-7 creates methods 'for legacy code' that merely pass through contents of `$_FILES` (which those libraries need to be correctly filled by PHP anyway, `move_uploaded_file` would fail otherwise). That said, it creates parts of interface that can be abused by folks who are used to `$_FILES` structure; and from the previous statements I'm inclined to conclude that it's being done basically for nothing. And very same parts of the interface kind of cancel the abstraction as shown by @m00t.

IMHO, PSR-7 is a great chance to get rid of `$_FILES`, `$_POST` et cetera in the field (to hide them as far as humanly possible). Let's use it, shall we?

webmozart:
I tried to digest the conversation and the changes above now. I agree with @m00t and @kamazee:

* I think `getFileParams()` should be removed. While I appreciate the intention behind it, I think that it is unnecessary in the end. Anything that can be done with `getFileParams()` can be done with `getUploadedFiles()`. Backwards compatibility should not be an issue for us at this point, I think that simplicity is more important.

* I think that `getTemporaryPath()` should be removed for the points outlined by @m00t. Let's keep the standard simple: Clients must use `move()`.

* To underline that `move()` is a once-only operation, and to underline that the file could only be partially uploaded at the point that `move()` is called (from what I understood), I would change its name to `save($path)`. The implementation decides how that operation is done.

* I agree with @m00t that `getStream()` should be called `getBody()` for consistency with `MessageInterface`.

* At last, I realized that `getUploadedFiles()` could be just `getFiles()`. `$request->getFiles()['logo']` is pretty clear code to me, and shorter than the current version.

m00t:
@weierophinney @webmozart I thought about `getStream()` vs `getBody()` naming again and found an argument for `getStream()`. "body" is a domain term when we speak about http messages, but there is no such term when we speak about uploaded files. So now I think `getBody()` for uploaded file should not be considered consistent with `getBody()` for http message. Maybe `getStream()` is actually a better name for this method.

However, my arguments about "`StreamableInterface` vs native streams" remain the same:
* consistence
* advantages of `StreamableInterface` compared to native php streams

weierophinney:
> What are arguments for using native streams instead of StreamableInterface?

To facilitate the most common operations, which would involve using PHP's native stream functions. If we use `StreamableInterface`, this then has to be passed to a `StreamWrapper` implementation first:

```php
$stream = new Psr7StreamableStreamWrapper($request->getFiles()['foo']->getStream());
```

I'm not terribly excited about this, as there is then an onus on the consumer to write such an implementation and/or import another library in order to support this functionality. That said, it *does* provide uniformity and consistency, and, as noted, has *some* advantages over native streams.

Preparing changes now.

weierophinney:
@webmozart :

> To underline that move() is a once-only operation, and to underline that the file could only be partially uploaded at the point that move() is called (from what I understood), I would change its name to write($path) (analogous to fwrite()). The implementation decides how that operation is done.

The problem I see with this is that as a consumer, I may not expect the original file to be removed when the naming is `write()`; I only expect that the path specified when invoking that method now exists. If I wanted to later access the stream and discovered I couldn't, this would be unexpected. Considering that we're recommending that `move_uploaded_file()` be used under SAPI environments, and that function is destructive, this scenario would be common.

`move()` more properly connotes that the original stream/file is no longer accessible after the operation, and, as such, I'll retain that verbiage.

weierophinney:
> At last, I realized that getUploadedFiles() could be just getFiles().

I have two arguments against that:

- `getFiles()` has a connotation that what is returned is analogous to `$_FILES`.
- `getUploadedFiles()` connotes that the return value contains instances of the `UploadedFileInterface`.

m00t:
@weierophinney Could you please rephrase your comment about streams? I am not sure I understand it correctly.

weierophinney:
@m00t Basically, I'm saying that it would be *easier* to have a PHP stream returned, as returning a `StreamableInterface` instance requires wrapping/decorating it in order for it to be used as a native PHP stream.

However, I've conceded the point, and have `getStream()` returning `StreamableInterface` as of the latest commit.

m00t:
@weierophinney thank you! I understood now.

weierophinney:
@webmozart and @m00t — I've incorporated feedback at this time. The changes made include:

- Removal of `getTemporaryFile()`.
- Updating `getStream()` to return a `StreamableInterface` instance.
- Moving the discussion of the `*_uploaded_file()` functions to the `move()` method docblock.
- Indication in `move()` and `getStream()` that calling either after `move()` has been called once will raise an exception.
- Removal of `getFileParams()` from `ServerRequestInterface`, and updating of verbiage around file uploads in that interface to remove references to `$_FILES`.
- Additions to section 1.6 to remove references to `getFileParams()` and add more examples around retrieving a stream and/or moving an upload.

I retained the verbiage `getUploadedFiles()` and `getStream() to have symmetry with the return values, as noted in previous comments.

Please review!

weierophinney:
One further note: I also retained the verbiage `move()` as indicated in a previous comment, as it connotes that the original stream is no longer available after the operation.

m00t:
@weierophinney I agree with you about `move()` method. This name seems perfect to me.

Reviewed interface changes: ready to merge IMO.

weierophinney:
I've updated this to reflect the changes from #490 (which renames `StreamableInterface` to `StreamInterface`.

Additionally, I've taken the liberty of adding @webmozart and @m00t as contributors to the spec at this time.

weierophinney:
@pmjones I've rebased this one now. :)

proposed/http-message-meta.md
```diff
@@ -594,6 +594,32 @@ specific, and the results of detection can be easily injected into objects that
 need it, and/or calculated as needed using utility functions and/or classes from
 the `RequestInterface` instance itself.
 
+### Why are both getFileParams() and getUploadedFiles() present?
+
```
webmozart: This paragraph seems to be superfluous now, or at least needs to be rephrased.

webmozart:
Thanks for all this @weierophinney! Great job :)

andersonamuller:
The term `MIME Type` is deprecated, now is used as `Media Type`, see: 

> The identifiers were originally defined in RFC 2046, and were called MIME types because they referred  to the non-ASCII parts of email messages that were composed using the MIME (Multipurpose Internet  Mail Extensions) specification. They are also sometimes referred to as Content-types.
http://en.wikipedia.org/wiki/Internet_media_type


> [RFC2046] specifies that Media Types (formerly known as MIME types) and Media Subtypes will be assigned and listed by the IANA.
https://www.iana.org/assignments/media-types/media-types.xhtml

proposed/http-message.md
```diff
@@ -1488,3 +1628,123 @@ interface UriInterface
...
+     * a malicious mime type with the intention to corrupt or hack your
+     * application.
+     *
+     * Implementations SHOULD return the value stored in the "type" key of
+     * the file in the $_FILES array.
+     *
+     * @return string|null The mime type sent by the client or null if none
+     *     was provided.
+     */
+    public function getClientMimeType();
```
andersonamuller: The term `MIME Type` is deprecated, now is used as `Media Type`, see: 

> The identifiers were originally defined in RFC 2046, and were called MIME types because they referred  to the non-ASCII parts of email messages that were composed using the MIME (Multipurpose Internet  Mail Extensions) specification. They are also sometimes referred to as Content-types.
http://en.wikipedia.org/wiki/Internet_media_type


> [RFC2046] specifies that Media Types (formerly known as MIME types) and Media Subtypes will be assigned and listed by the IANA.
https://www.iana.org/assignments/media-types/media-types.xhtml

PR #500, opened 2015-04-09
Making withUri() and Host header modifications more explicit

mtdowling:
This PR updates the `witUri` method of a request to be more explicit about its relationship to the Host header, and updates this method to perform the expected behavior of assuming the host component of the provided URI by default.

Associating a URI with a request is an abstraction on top of an HTTP message. This abstraction isn't technically required to represent an HTTP message, but it is provided in PSR-7 as a convenience for clients and servers to know more about the transport of the message (destination or source) and to more easily slice up things in the URI (path, query, etc).

One of the components of the URI is the host, and an HTTP request also has a Host header. When calling `withRequest()` with a URI that contains a host component the HTTP request should be updated to use the host component of the URI by default, however, there should be a way to opt out of this transformation if desired. Updating the Host header by default reflects the most common usage of an HTTP message: you typically want to send a request to the same destination as the URI you provided, and diverging from this path is atypical.

This change also simplifies the current behavior of only overriding the Host header in the request if a Host header was not explicitly set. This change removes that concept entirely. For example, all of the "softHost" stuff from Guzzle's PSR7 package would be removed and this method would be greatly simplified (and easier to use IMO): https://github.com/guzzle/psr7/blob/master/src/Request.php#L136

@weierophinney @evert

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -654,7 +654,17 @@ interface RequestInterface extends MessageInterface
     public function getUri();
 
     /**
-     * Return an instance with the provided URI.
+     * Returns an instance with the provided URI.
+     *
+     * This method will update the Host header of the returned request by
+     * default if the URI contains a host component. If the URI does not
+     * contain a host component, any pre-existing Host header will be carried
+     * over to the returned request.
+     *
+     * You can disable modifying the Host header of the request by setting
+     * `$overrideHost` to false. Note that when `$overrideHost` is set to
+     * false, the returned request will not update the Host header of the
+     * message-- even if the message contains no Host header.
      *
      * This method MUST be implemented in such a way as to retain the
      * immutability of the message, and MUST return an instance that has the
@@ -662,9 +672,10 @@ interface RequestInterface extends MessageInterface
      *
      * @link http://tools.ietf.org/html/rfc3986#section-4.3
      * @param UriInterface $uri New request URI to use.
+     * @param bool $overrideHost Set to false to not change the Host header.
      * @return self
      */
-    public function withUri(UriInterface $uri);
+    public function withUri(UriInterface $uri, $overrideHost = true);
 }
 ```
 
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -664,7 +664,7 @@ interface RequestInterface extends MessageInterface
      * You can opt-in to preserving the original state of the Host header by
      * setting `$preserveHost` to `true`. When `$preserveHost` is set to
      * `true`, the returned request will not update the Host header of the
-     * returned message-- even if the message contains no Host header. This
+     * returned message -- even if the message contains no Host header. This
      * means that a call to `getHeader('Host')` on the original request MUST
      * equal the return value of a call to `getHeader('Host')` on the returned
      * request.
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -661,10 +661,13 @@ interface RequestInterface extends MessageInterface
      * contain a host component, any pre-existing Host header will be carried
      * over to the returned request.
      *
-     * You can disable modifying the Host header of the request by setting
-     * `$overrideHost` to false. Note that when `$overrideHost` is set to
-     * false, the returned request will not update the Host header of the
-     * message-- even if the message contains no Host header.
+     * You can opt-in to preserving the original state of the Host header by
+     * setting `$preserveHost` to `true`. When `$preserveHost` is set to
+     * `true`, the returned request will not update the Host header of the
+     * returned message-- even if the message contains no Host header. This
+     * means that a call to `getHeader('Host')` on the original request MUST
+     * equal the return value of a call to `getHeader('Host')` on the returned
+     * request.
      *
      * This method MUST be implemented in such a way as to retain the
      * immutability of the message, and MUST return an instance that has the
@@ -672,10 +675,10 @@ interface RequestInterface extends MessageInterface
      *
      * @link http://tools.ietf.org/html/rfc3986#section-4.3
      * @param UriInterface $uri New request URI to use.
-     * @param bool $overrideHost Set to false to not change the Host header.
+     * @param bool $preserveHost Preserve the original state of the Host header.
      * @return self
      */
-    public function withUri(UriInterface $uri, $overrideHost = true);
+    public function withUri(UriInterface $uri, $preserveHost = false);
 }
 ```
 


weierophinney:
:+1: 

Unless @evert disagrees, ready to merge!

Thanks for the clarification, @mtdowling .

simensen:
@mtdowling @evert @weierophinney how would you feel about making this its own method vs adding an optional parameter with a default?

    withUri(UriInterface $uri)
    withUriThatOverridesHost(UriInterface $uri)

... or something along those lines?

 * withUriThatOverridesHost(UriInterface $uri)
 * withUriAndOverrideHost(UriInterface $uri)
 * ...?

evert:
Don't mind the change. No opinion on a separate method vs. an argument, but if it's in the original form I do feel I would negate the argument as such:

    public function withUri(UriInterface $uri, $preserveHost = false);

Not sure why, but that argument name makes it more clearer to me. But no biggie either way ;)

simensen:
Even though I try hard with this, I still have a hard time naming things. :) I think the optional boolean argument is confusing, though, as I find myself having to look them up ever time to know 1) what is the default if I don't pass anything, 2) decode from the argument whether i should pass true or false, 3) determine if the default value is what i was looking for to begin with or not, 4) 50% of the time i get it wrong and end up changing true => false or false => true after running the code and seeing that I had messed up one of the steps.

The alternative is potentially having an explosion in the number of methods with varying combinations of things. With this interface it seems like that possibility is limited.

weierophinney:
@simensen From an implementor's perspective, adding the flag is easiest. From a usability perspective, flags are often hard to remember, though IDEs will typically provide the information you need as you're coding an invocation. There are some ways to make it easier for a reader of your code to understand what you're doing:

```php
$message->withUri($uri, $overrideHost = false);
```

which makes it clear what your intent is.

I'd personally prefer not to add another method, as every method we add:

- becomes more to implement
- opens the door for more proposals that add methods ("but you just added one in this PR")

I can go with either the proposed flag or the alternate proposed by @evert ; @mtdowling — thoughts?

simensen:
I realize my examples before were opposite of the desired change that the *default* behavior is for `withUri` to update the host header.

    $message->withUri($uri) // updates host header
    $message->withUriButPreserveHostHeader($uri) // does not update host header

or we can do away with `withUri` and name them more explicitly so that it is always clear what is going on:

    $message->withUriAndOverrideHostHeader($uri) // updates host header
    $message->withUriButPreserveHostHeader($uri) // does not update host header

I'm able to go either way on this. If we go the optional argument route I have no opinions on which way to go for default to true/false. Either is fine to me except I'd defer to what is usually done in PHP if there is such a thing as a "usual" default boolean value in PHP.

mtdowling:
I updated the PR to use @evert's suggestion of `preserveHost` and added much clearer language to how it handles the host header.

evert:
:+1:

proposed/http-message.md
```diff
@@ -654,7 +654,20 @@ interface RequestInterface extends MessageInterface
...
+     *
+     * This method will update the Host header of the returned request by
+     * default if the URI contains a host component. If the URI does not
+     * contain a host component, any pre-existing Host header will be carried
+     * over to the returned request.
+     *
+     * You can opt-in to preserving the original state of the Host header by
+     * setting `$preserveHost` to `true`. When `$preserveHost` is set to
+     * `true`, the returned request will not update the Host header of the
+     * returned message-- even if the message contains no Host header. This
```
weierophinney: Add a space between "message" and the trailing "--".

weierophinney:
Other than the change noted, :+1: — ready to merge, @pmjones and/or @simensen !

simensen:
@weierophinney OK, we'll wait for @mtdowling to make the suggested change before merging.

mtdowling:
Done

weierophinney:
@simensen and/or @pmjones — :shipit: ! :)

Tobion:
I realyl don't see what this new parameter addresses that was not possible before. Pleaes give concrete example. From what I see, preserving the host was already possible by just passing a URI without a host.

weierophinney:
@Tobion — The use case is primarily in clients. In clients, the URI will be used to make the connection, and needs to have the host (and potentially port) in order to make the TCP connection. In some of these cases, especially when working with proxies, the `Host` header value may need to differ from the URI.

The idea behind the `$preserveHost` flag is because in most cases you will want the URI host synced with the `Host` header, and this should be automated. But in cases like the above, when you want to specify a URI with a different host, this flag enables that.

Tobion:
I see the use-case. But I asked why we need that parameter since it's already possible to do the same without paramter:

```
$uriWithoutHost = $uri->withHost('');
$request->withUri($uriWithoutHost);
```

So why add a parameter to an interface that is already possible with one line of code?

evert:
@Tobion withUri doesn't just update the values you are interested in updating, the URI object needs to be 100% valid and self-contained. An empty host is not valid for HTTP URIs.

pjdietz:
@Tobion The idea is that you may want the URI to have a host AND have a Host header that differs from that host.

Tobion:
@evert an URI object without a host is 100% valid. Look at `__toString`. It's exactly to express a URI relative to another host.

@pjdietz that is also possible:

```
$request->withUri($uri)->withHeader('Host', 'myhost.com:8800');
```

weierophinney:
@Tobion That last example is problematic when you've already set the `Host` header. It's still possible to update it without the flag:

```php
$request->withUri($uri)->withHeader('Host', $request->getHeaderLine('Host'));
```

but that's a lot to remember. The new flag makes the relationship explicit, and reduces the potential for errors by consumers.

Tobion:
Big :-1: as there is no need for it. `$request->withUri($uri)->withHeader('Host', $request->getHeaderLine('Host'));` is way more explicit and understandadble than `$request->withUri($uri, true)`

Tobion:
First rule of APIs. Don't add boolean values to change code flow when it's not even necessary.

weierophinney:
@Tobion — Please be cognizant that there *will* be areas of the API that you will not like. You don't have to agree with every aspect; the question is if you can live with what we have.

Several contributors have indicated a need for this functionality, and as these are people implementing the spec and/or creating code consuming the spec, we've added the flag. The `boolean` flag argument was brought up already *in this very comment thread*, and the various parties felt it was a necessary evil to address the use case.

You can completely ignore the flag and use the syntax outlined above; the API does not prevent that. However, having the flag addresses existing use cases.

I ask that you respect *all* contributors and not judge the spec on individual aspects such as this, but on its usefulness as a whole. There *will* be imperfections; there always will, as it's written by humans, and our understanding of the context and our programming practices evolve and change over time.

Tobion:
The discussion so far said, there was no way to opt-out of the host update. But as I proved, there is is way. Actually there two ways. So the assumption was wrong.
And now you added a third way to do the same thing without realizing it I guess. And so far you didn't show why we still need that parameter.
And repeating that PSR-7 will have flaws is kinda a bad execuse.

simensen:
I'd like to hear what @mtdowling and @evert have to say about this. Is this an acceptable replacement for the new flag added?

```php
<?php
$request->withUri($uri)->withHeader('Host', $request->getHeaderLine('Host'));
```

I was opposed to adding the boolean flag but it was what the rest of the group agreed to and wanted. If the above is not considered acceptable by the people who initially raised the concern (especially @mtdowling) then I think we should stick with the proposal as-is.

The justification I can see for keeping this a flag is that the `withUri` method is automatically does some magic behind the scenes to update the Host header. I can see the case being made that it makes more sense to be able to *disable* that behavior of this method rather than to require the end-user to have to do more work because of it.

mtdowling:
I honestly don't care as long as the Host header of the message is updated using the host component of the URI by default instead of trying to preserve an "explicitly provided Host header".

Tobion:
As I said, this is also possible which is even shorter.
```
<?php
$request->withUri($uri->withHost(''));
```

weierophinney:
@Tobion that example *does not* work, as you then no longer have a URI that contains all the details necessary to make a connection. The only one that will work is the original example I provided.

@mtdowling IIRC, we were auto-updating the `Host` header using the URI host component previously; this change that you provided was simply to allow _not_ doing that, correct? Should we remove it the flag, knowing there's an alternate mechanism?

Tobion:
@weierophinney sure it works for the use-case this ticket is about. What information is missing to send a request?

```
GET /path HTTP/1.1
Host: <the one that was set previously>
```

mtdowling:
The previous behavior was stateful and a bit awkward because is would not update the Host header of the message if you previously set a Host header on the message, but it would update the Host header if the message had no previously set Host header value. What this caused is that that you had to track whether or not a message *really* has a host header or if it is just inheriting one from a URI. If it's just inheriting one from a previous `withUri` call, then the next `withUri` call would overwrite the host header of the message. If you actually did previously set a Host header, then a call to `withUri` did not update the Host header.

The crux of this change was to propose that the most common case is not to preserve the previously set Host header (whether as a result of withUri or setting a host header explicitly), but rather to always use the host component of the URI (if set) as the new Host header. It is atypical to want a different behavior.

I think @Tobion's proposal would work, as long as the Host header is always updated using the host component of the URI if it is present.

pjdietz:
@Tobion, you may want the client to use `$request->getUri()` to determine where to make the connection, but send a `Host` header in the request that differs from `$request->getUri()->getHost()`.

The problem with using:

```php
$request->withUri($uri->withHost(''));
```

is that this updates the `Host` header AND replaces the URI instance with a new URI that does not include a `Host`. The previous URI is no longer available to use to determine where to make the connection.

Tobion:
If the URI doesn't have a host but there is a Host header, then I assume the client to use the host header to make the connection of course. The only case where host header != connection host is for namebased virtual hosts. And for this you can use option 2. I just said that `$request->withUri($uri->withHost(''));` is also valid depending on the use-case.

mtdowling:
True, but keep in mind that the URI returned by getUri should always be something a client could connect to. In order for that to work, you'd need to do something like this:

```php
public function withUri(UriInterface $uri)
{
    if ($uri === $this->uri) {
        return $this;
    }

    $new = clone $this;
    
    if ($host = $new->uri->getHost()) {
        // Update the host header of the request
        if ($port = $uri->getPort()) {
            $host .= ':' . $port;
        }
        $new->setHeader('Host', $host);
    } elseif ($host = $this->getHeaderLine('Host')) {
        // Update the host component of the URI
        $uri = $uri->withHost($host);
    }
    
    $new->uri = $uri;
    return $new;
}
```

simensen:
@mtdowling Since this was your original PR, I want to make sure I know where you stand on this. Your main concern was that ->withUri() should always update the Host header by default and you provided a way to bypass that behavior in case someone wanted to do so.

Are you OK if we remove the additional boolean flag and the verbiage around it and only discuss the fact that withUri MUST update the host header? I believe that in the meta document we could discuss alternatives for people who want to NOT update the host header (or rather, how they can workaround the fact that the Host header is always updated....)

Since @mtdowling proposed this addition and he has now said that it doesn't matter to him if it is there or not, I'm inclined to consider removing the boolean argument. However, I'd like to get one more round of feedback from @evert and @weierophinney to make sure we are all in agreement on this. I don't want to go back and forth on this issue much longer.

mtdowling:
As long as the behavior of withUri matches  what I coded up above, then I don't care if there is a boolean flag or not.

evert:
I would personally prefer the ability to opt-out from updating the host header, or not automatically updating the host header at all. So the suggestion to remove the flag and make that behavior the default would definitely be -1 for me.

evert:
So to be clear, I'm ok with the PR, and I'm ok with the original behavior, but I think the new suggestion is a little bit worse.

We also had an earlier idea to simply make getHeader(), getHeaderLines() and getHeaders() to *use* the value from getUri() to return the host if it was not explicitly overwritten using `withHeader()`.

I think this also solves the main use-case from @mtdowling, and this will give it more consistency, as this is exactly how `getRequestTarget()` behaves. I think it would be poor design to make such similar concepts have so different behaviors.

evert:
Another issue with the proposed behavior, is that the 'end result' of your request object is dependent on the order in which you set up the object. These two lines of code would result in different objects:

```php
$r->withUri(...)->withHeader('Host', '...');
$r->withHeader('Host','...')->withUri('...');
```

pjdietz:
> We also had an earlier idea to simply make getHeader(), getHeaderLines() and getHeaders() to use the value from getUri() to return the host if it was not explicitly overwritten using withHeader().

@evert that behavior is actually still there as well, and I think having BOTH is problematic. I opened #522 which removes the overrides.

mtdowling:
PJ: that could work, but you'd need to force implementations to use the host component of the Uri as the Host header when the message is constructed if the message has no host header. With the current header behavior, we have the same behavior but don't dictate how messages are created.

pjdietz:
@mtdowling that's a good point. It is still possible though to have a Request with no Host header even with both approaches in place since `UriInterface::getHost()` can return an empty string. To guarantee a host header, we'd still need to spell out how messages are created.

mtdowling:
That's true. We aren't trying to prevent a request from being able to have no host header. We are just trying to make sure it returns the most appropriate host header at any point in time (which as you pointed out could be an empty string).

pjdietz:
I think there's still a consistency problem with the line "This means that a call to getHeader('Host') on the original request MUST equal the return value of a call to getHeader('Host') on the returned request." 

It's a goofy case where this breaks, but still an inconsistency.

pjdietz:
Thoughts on rephrasing the `withUri()` description to something like "If a request has an explicitly set Host header, a call to getHeader('Host') on that request MUST equal the return value of a call to getHeader('Host') on the returned request." It's a mouthful, but that would solve the inconsistency and provide the most appropriate Host header at any point.

mtdowling:
That nullifies how it currently overrides the host header by default if the uri has a host component. I think that overriding the host header by default would almost always be the intention of the user.

pjdietz:
Sorry for the confusion. That part is in the section specifically about opting in preserving the host. The full paragraph would be something like:

> You can opt-in to preserving the original state of the Host header by
> setting $preserveHost to true. When $preserveHost is set to
> true, the returned request will not update the Host header of the
> returned message -- even if the message contains no Host header. If a request has an explicitly set Host header, a call to getHeader('Host') on that request MUST equal the return value of a call to getHeader('Host') on the returned request.

pjdietz:
Or, more explicitly:

> You can opt-in to preserving the original state of the Host header by setting $preserveHost to true. When $preserveHost is set to true, the returned request will not update the Host header of the returned message -- even if the message contains no Host header. If a request has an explicitly set Host header, a call to getHeader('Host') on that request MUST equal the return value of a call to getHeader('Host') on the request returned by withUri() with $preserveHost set to true.

Or, just drop the last sentence entirely.

pjdietz:
Opened #523 which removes the line from the description of `RequestInterface::withUri()` and leaves `MessageInterface` as is.

Tobion:
So you decided we need that $preserveHost parameter?

pjdietz:
@Tobion I don't have any objection to rolling back the "MUST"s to "will"s, etc., but I think that makes the expected behavior less clear.

The more important issue in #523 is to resolve the conflict where `withUri` requires behavior that is at odds with the behavior required by `Message::getHeader*`.

Tobion:
IMO there are two ways:
- remove all magic and it's the responsibility of the user to set the stuff he needs, like the Host header in most cases.
- leave the default host update in place because that is the most common case, remove the `preserveHost` parameter and also remove the `getHeader*` magic which doesn't make sense in this case as proposed in #522. if a user wants to use a different host then provided by the URI, he can just set it explicitly.

mtdowling:
Tobias: can you open a PR to make your suggestion more concrete? I kind of agree with you but I'd like to see the specific implications of these changes.

-Michael

PR #503, opened 2015-04-10
[PSR-7] UriInterface: rootless path, delimiters, consistency

Tobion:
- support rootless path as continuation of #487
  - the path can now also be `$request->withPath('rootless/foo')`
  - it does not complicate anything and REMOVES the optional magic that was present that the path MAY be prefixed by `/` automatically
  - is more in line with RFC 3986 that also supports rootless paths, e.g. `mailto:tobias@example.org`
  - we need it for resolving URIs relative to a base path (another URI) as explained by @mtdowling in the ticket
- clarify delimiters that replaces #480
  - that getQuery() for http://example.org/path?foo=bar returns `foo=bar` (without `?`) is clear and also documented. this is task of the code that splits a URI into its components, e.g. the constructor of Uri. there is no method defined for that in PSR-7
  - but addditionally, before there was magic defined that `$request->withQuery('?foo=bar')->getQuery()` returns `foo=bar` which makes no sense as I tried to explain in https://github.com/php-fig/fig-standards/pull/480#issuecomment-91286771
    - magic
    - error prone
    - slows down every operation unessarily
    - it must return `%3Ffoo=bar` instead
- consistent wording, e.g. sometimes there was an article sometimes not and it now also uses the same wording for the same concepts for each method
- scheme and host are case insensitive and normalized to lowercase which is important for usability allowing simple comparison, e.g. `$request->getScheme() === 'http'`

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -1154,10 +1154,13 @@ interface StreamableInterface
 namespace Psr\Http\Message;
 
 /**
- * Value object representing a URI for use in HTTP requests.
+ * Value object representing a URI.
  *
- * This interface is meant to represent only URIs for use with HTTP requests,
- * and is not intended as a general-purpose URI implementation.
+ * This interface is meant to represent URIs according to RFC 3986 and to
+ * provide methods for most common operations. Additional functionality for
+ * working with URIs can be provided on top of the interface or externally.
+ * Its primary use is for HTTP requests, but may also be used in other
+ * contexts.
  *
  * Instances of this interface are considered immutable; all methods that
  * might change state MUST be implemented such that they retain the internal
@@ -1177,9 +1180,13 @@ interface UriInterface
      *
      * If no scheme is present, this method MUST return an empty string.
      *
+     * The value returned MUST be normalized to lowercase, per RFC 3986
+     * Section 3.1.
+     *
      * The trailing ":" character is not part of the scheme and MUST NOT be
      * added.
      *
+     * @see https://tools.ietf.org/html/rfc3986#section-3.1
      * @return string The URI scheme.
      */
     public function getScheme();
@@ -1199,6 +1206,7 @@ interface UriInterface
      * If the port component is not set or is the standard port for the current
      * scheme, it SHOULD NOT be included.
      *
+     * @see https://tools.ietf.org/html/rfc3986#section-3.2
      * @return string The URI authority, in "[user-info@]host[:port]" format.
      */
     public function getAuthority();
@@ -1225,6 +1233,10 @@ interface UriInterface
      *
      * If no host is present, this method MUST return an empty string.
      *
+     * The value returned MUST be normalized to lowercase, per RFC 3986
+     * Section 3.2.2.
+     *
+     * @see http://tools.ietf.org/html/rfc3986#section-3.2.2
      * @return string The URI host.
      */
     public function getHost();
@@ -1435,10 +1447,12 @@ interface UriInterface
     public function withFragment($fragment);
 
     /**
-     * Return the string representation of the URI.
+     * Return the string representation as a URI reference.
      *
-     * Concatenates the various components of the URI, using the appropriate
-     * delimiters:
+     * Depending on which components of the URI are present, the resulting
+     * string is either a full URI or relative reference according to RFC 3985,
+     * Section 4.1. The method concatenates the various components of the URI,
+     * using the appropriate delimiters:
      *
      * - If a scheme is present, "://" MUST append the value.
      * - If the authority information is present, that value will be
@@ -1449,6 +1463,7 @@ interface UriInterface
      * - If a query string is present, it MUST be prefixed by a "?" character.
      * - If a URI fragment is present, it MUST be prefixed by a "#" character.
      *
+     * @see http://tools.ietf.org/html/rfc3986#section-4.1
      * @return string
      */
     public function __toString();
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -1173,23 +1173,24 @@ namespace Psr\Http\Message;
 interface UriInterface
 {
     /**
-     * Retrieve the URI scheme.
-     *
-     * Implementations SHOULD restrict values to "http", "https", or an empty
-     * string but MAY accommodate other schemes if required.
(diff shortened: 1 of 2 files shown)


weierophinney:
@Tobion — thanks!

My argument in #480 regarding the delimiters is this: the verbiage about delimiters exists to ensure that consumers know that they are receiving *only the data for the component*, without the delimiters. It was never about normalization; it was about ensuring consistency of return value, so that the consumer knows what they can do with it.

I don't want them to *assume* anything; I want them to *know* what the value is.

Other than those changes, I'm fine with this PR. But I think the delimiter specification is important for interface consumers.

@nyamsprod and @simensen — what are your thoughts?

proposed/http-message.md
```diff
@@ -1173,23 +1173,24 @@ namespace Psr\Http\Message;
...
 
     /**
-     * Retrieve the authority portion of the URI.
+     * Retrieve the authority component of the URI.
+     *
+     * If no authority information is present, this method MUST return an empty
+     * string.
      *
-     * The authority portion of the URI is:
+     * The authority syntax of the URI is:
```
mtdowling: Maybe this should instead link to the ABNF in RFC 3986? https://tools.ietf.org/html/rfc3986#appendix-A

Tobion:
@weierophinney yes that's what I figured. I didn't remove the verbiage about delimiters in the getters, see diff. So users don't have to assume things.
But what caused this delimiters discussion was the wording in the `with` methods, e.g. `If the query string is prefixed by "?", that character MUST be removed.` in `withQuery`. This means something completely else as I demonstrated above. I guess it just wasn't meant this way.

weierophinney:
@Tobion Okay, thanks for the clarification; missed that the verbiage was still in the getters.

Will review fully in a bit; we should be able to close #480 and merge this one later today. Thanks for making clear what your intentions were!

proposed/http-message.md
```diff
@@ -1402,9 +1404,11 @@ interface UriInterface
      * This method MUST retain the state of the current instance, and return
      * an instance that contains the specified query string.
      *
-     * If the query string is prefixed by "?", that character MUST be removed.
-     * Additionally, the query string SHOULD be parseable by parse_str() in
-     * order to be valid.
+     * The query string SHOULD be parseable by parse_str() in order to be
```
mtdowling: What's the purpose of this statement? Why must it be parseable by PHP's parse_str()? Shouldn't the only requirement be that it adheres to RFC 3986?

proposed/http-message.md
```diff
@@ -1385,10 +1385,12 @@ interface UriInterface
...
      * an instance that contains the specified path.
      *
-     * The path MUST be prefixed with "/"; if not, the implementation MAY
-     * provide the prefix itself.
+     * The path can either be empty or absolute (starting with a slash) or
+     * rootless (not starting with a slash). Implementations MUST support all
+     * three syntaxes.
      *
-     * An empty path value is equivalent to removing the path.
+     * Users can provide the path both in encoded as well as decoded form.
```
Tobion: is this correct English? "both ... as well" sounds strange, I think I have to change it to "both in encoded and in decoded form".

proposed/http-message.md
Tobion: I didn't add it and also wondered. It basically means that only a subset of query strings need to be supported. Because `parse_str('=bar')` doesn't work but would be a valid query string AFAIK.

proposed/http-message.md
Tobion: I can add `@see https://tools.ietf.org/html/rfc3986#section-3.2` as we do in the other methods. Is that enough?

Tobion: But I think having `[user-info@]host[:port]` in the text is important to understand it without following the link.

weierophinney: Agreed with @Tobion — it's nice to have it in the text so that the consumer does not need to look it up, but an `@see` annotation can be used to provide more information.

proposed/http-message.md
weierophinney: Correct English as-is. A potential clarification would be "Users can provide the path in either encoded or decoded form." (It's an either-or, not both, situation.)

proposed/http-message.md
weierophinney: This is why it's a SHOULD and not a MUST. Typically, for PHP, if it's not parseable by `parse_str()`, `$_GET`, and thus `getQueryParams()`, will be empty. However, since we are also targeting client-side usage, I think we can safely remove it.

proposed/http-message.md
weierophinney: If you make that change, I see the same verbiage in a few of the other methods, and the change should be mirrored in those.

weierophinney:
Once @Tobion has made the last few requested changes, this one is ready to merge!

mtdowling:
LGTM

proposed/http-message.md
Tobion: Actually it could also be a both situation when some chars are encoded and some aren't. So maybe keep my wording?

weierophinney: How about "Users can provide both encoded and decoded path characters."?

proposed/http-message.md
Tobion: removed the sentence

proposed/http-message.md
Tobion: used your suggestion

Tobion:
I added two commits that clarify some more things. Also added lowercase normalization of of scheme and host as we already agreed on in #480

proposed/http-message.md
```diff
@@ -1154,10 +1154,13 @@ interface StreamableInterface
 namespace Psr\Http\Message;
 
 /**
- * Value object representing a URI for use in HTTP requests.
+ * Value object representing a URI.
  *
- * This interface is meant to represent only URIs for use with HTTP requests,
```
Tobion: Now that we have rootless path support as well, there is nothing that hinders UriInterface to be used generally. It's just that it doesn't define all possible methods to work with URIs, e.g. URI comparision etc. I hope I clarified that while also giving arguments why it's in the Http/Message namespace.

weierophinney:
:+1:

@pmjones and/or @simensen — ready to merge!

PR #504, opened 2015-04-10
[PSR-7] Use exceptions for error handling in streams

weierophinney:
Per discussion:

- https://groups.google.com/d/msgid/php-fig/d5a9b044-1862-4333-bb67-645de997078c%40googlegroups.com

@ircmaxell indicates that raising exceptions is acceptable for StreamWrapper implementations, making it acceptable within this domain.

This patch modifies `StreamableInterface` methods that were following PHP stream error handling conventions (i.e., multiple return values, using int/bool returns to indicate errors) to instead raise `RuntimeException` instances.

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -1029,6 +1029,10 @@ interface StreamableInterface
      *
      * Warning: This could attempt to load a large amount of data into memory.
      *
+     * This method MUST NOT raise an exception in order to conform with PHP's 
+     * string casting operations.
+     *
+     * @see http://php.net/manual/en/language.oop5.magic.php#object.tostring
      * @return string
      */
     public function __toString();
@@ -1059,7 +1063,8 @@ interface StreamableInterface
     /**
      * Returns the current position of the file read/write pointer
      *
-     * @return int|bool Position of the file pointer or false on error.
+     * @return int Position of the file pointer
+     * @throws \RuntimeException on error.
      */
     public function tell();
 
@@ -1088,6 +1093,7 @@ interface StreamableInterface
      *     offset bytes SEEK_CUR: Set position to current location plus offset
      *     SEEK_END: Set position to end-of-stream plus offset.
      * @return bool Returns TRUE on success or FALSE on failure.
+     * @throws \RuntimeException on failure.
      */
     public function seek($offset, $whence = SEEK_SET);
 
@@ -1100,7 +1106,8 @@ interface StreamableInterface
      *
      * @see seek()
      * @link http://www.php.net/manual/en/function.fseek.php
-     * @return bool Returns TRUE on success or FALSE on failure.
+     * @return bool Returns TRUE on success.
+     * @throws \RuntimeException on failure.
      */
     public function rewind();
 
@@ -1116,7 +1123,8 @@ interface StreamableInterface
      *
      * @param string $string The string that is to be written.
      * @return int|bool Returns the number of bytes written to the stream on
-     *     success or FALSE on failure.
+     *     success.
+     * @throws \RuntimeException on failure.
      */
     public function write($string);
 
@@ -1133,8 +1141,8 @@ interface StreamableInterface
      * @param int $length Read up to $length bytes from the object and return
      *     them. Fewer than $length bytes may be returned if underlying stream
      *     call returns fewer bytes.
-     * @return string|false Returns the data read from the stream, false if
-     *     unable to read or if an error occurs.
+     * @return string Returns the data read from the stream
+     * @throws \RuntimeException if unable to read or if an error occurs.
      */
     public function read($length);
 
@@ -1142,6 +1150,8 @@ interface StreamableInterface
      * Returns the remaining contents in a string
      *
      * @return string
+     * @throws \RuntimeException if unable to read or an error occurs while 
+     *     reading.
      */
     public function getContents();
 
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -1140,8 +1140,9 @@ interface StreamableInterface
      * @param int $length Read up to $length bytes from the object and return
      *     them. Fewer than $length bytes may be returned if underlying stream
      *     call returns fewer bytes.
-     * @return string Returns the data read from the stream
-     * @throws \RuntimeException if unable to read or if an error occurs.
+     * @return string|null Returns the data read from the stream, or null if no bytes
+         are available and the stream is in non-blocking mode.
+     * @throws \RuntimeException if an error occurs.
      */
     public function read($length);
 
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -1138,7 +1138,7 @@ interface StreamableInterface
      *     them. Fewer than $length bytes may be returned if underlying stream
      *     call returns fewer bytes.
      * @return string|null Returns the data read from the stream, or null if no bytes
-         are available and the stream is in non-blocking mode.
+     *     are available and the stream is in non-blocking mode.
      * @throws \RuntimeException if an error occurs.
      */
     public function read($length);
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -1122,8 +1122,7 @@ interface StreamableInterface
      * Write data to the stream.
      *
      * @param string $string The string that is to be written.
-     * @return int|bool Returns the number of bytes written to the stream on
-     *     success.
+     * @return int Returns the number of bytes written to the stream.
      * @throws \RuntimeException on failure.
      */
     public function write($string);
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -1092,7 +1092,6 @@ interface StreamableInterface
      *     PHP $whence values for `fseek()`.  SEEK_SET: Set position equal to
      *     offset bytes SEEK_CUR: Set position to current location plus offset
      *     SEEK_END: Set position to end-of-stream plus offset.
-     * @return bool Returns TRUE on success or FALSE on failure.
      * @throws \RuntimeException on failure.
      */
     public function seek($offset, $whence = SEEK_SET);
@@ -1100,13 +1099,11 @@ interface StreamableInterface
     /**
      * Seek to the beginning of the stream.
      *
-     * If the stream is not seekable, this method will return FALSE, indicating
-     * failure; otherwise, it will perform a seek(0), and return the status of
-     * that operation.
+     * If the stream is not seekable, this method will raise an exception;
+     * otherwise, it will perform a seek(0).
      *
      * @see seek()
      * @link http://www.php.net/manual/en/function.fseek.php
-     * @return bool Returns TRUE on success.
      * @throws \RuntimeException on failure.
      */
     public function rewind();


proposed/http-message.md
```diff
@@ -1100,7 +1106,8 @@ interface StreamableInterface
      *
      * @see seek()
      * @link http://www.php.net/manual/en/function.fseek.php
-     * @return bool Returns TRUE on success or FALSE on failure.
+     * @return bool Returns TRUE on success.
```
mtdowling: Why return a value at all?

proposed/http-message.md
```diff
@@ -1116,7 +1123,8 @@ interface StreamableInterface
      *
      * @param string $string The string that is to be written.
      * @return int|bool Returns the number of bytes written to the stream on
```
mtdowling: should this just be `int` then?

simensen: :+1:

proposed/http-message.md
weierophinney: For compatibility with StreamWrapper::stream_seek and fseek; the boolean true is needed to work with those.

simensen: This is a good point.

proposed/http-message.md
weierophinney: Yep; I'll fix that now.

proposed/http-message.md
```diff
@@ -1133,8 +1141,8 @@ interface StreamableInterface
      * @param int $length Read up to $length bytes from the object and return
      *     them. Fewer than $length bytes may be returned if underlying stream
      *     call returns fewer bytes.
-     * @return string|false Returns the data read from the stream, false if
-     *     unable to read or if an error occurs.
+     * @return string Returns the data read from the stream
+     * @throws \RuntimeException if unable to read or if an error occurs.
```
mtdowling: What does "unable to read" mean? Throw an exception when a non blocking stream has no data? Also, what would `''` mean? Does it signify EOF if you requested more than 0 bytes?

Here's how Python describes their read() method: https://docs.python.org/3.1/library/io.html#io.RawIOBase.read. The takeaway would be:

> Fewer than n bytes may be returned if the operating system call returns fewer than n bytes.
> If 0 bytes are returned, and n was not 0, this indicates end of file. If the object is in non-blocking mode and no bytes > are available, None is returned.

Maybe we need something along those lines to be more explicit.

proposed/http-message.md
simensen: Should we then also allow returning FALSE and leave RuntimeException for weird exceptions or not throw at all for this method? Seems weird to me.

proposed/http-message.md
weierophinney: In the case of this:

>  If the object is in non-blocking mode and no bytes > are available, None is returned.

what would be the PHP equivalent of Python's `None`? `null`?

proposed/http-message.md
mtdowling: I don't understand why the boolean is needed if the success case is not throwing an exception.

proposed/http-message.md
mtdowling: Perhaps that part can be removed in favor of just returning 0 bytes.

weierophinney: I've taken a stab at this now; let me know if it works.

proposed/http-message.md
weierophinney: @mtdowling `bool` is specified because the `true` and `false` values are not valid as annotation types. So, `@return bool Returns TRUE` is how we signal `true` as a return value.

proposed/http-message.md
```diff
@@ -1088,6 +1093,7 @@ interface StreamableInterface
      *     offset bytes SEEK_CUR: Set position to current location plus offset
      *     SEEK_END: Set position to end-of-stream plus offset.
      * @return bool Returns TRUE on success or FALSE on failure.
```
mtdowling: I have the same question here as in rewind(): why return a boolean success flag if the failure case is throwing an exception? If we absolutely need a return value, perhaps we could return the new position in the string ala https://docs.python.org/3.1/library/io.html#io.IOBase.seek

proposed/http-message.md
weierophinney: We could remove the return value. Since you'd have to wrap it anyways in order to use it with StreamWrapper, it may not be necessary:

```php
public function stream_seek($offset, $whence = SEEK_SET)
{
    $this->seek($offset, $whence);
    return true;
}
```

I'm going to update these various cases.

mtdowling: I apologize, I don't think I was being clear. This method says that it will throw a RuntimeException on error and will return `true` on success, with false not allowed. My point is that returning any value at all is unnecessary because the failure case here is to throw an exception meaning that the return value is irrelevant. I also don't think we need to return the value returned by an underlying fseek() or rewind() call as those also have a true/false return value which we have abstracted into throwing an exception.

weierophinney:
@mtdowling I've incorporated your feedback now. :)

proposed/http-message.md
```diff
@@ -1133,8 +1137,9 @@ interface StreamableInterface
      * @param int $length Read up to $length bytes from the object and return
      *     them. Fewer than $length bytes may be returned if underlying stream
      *     call returns fewer bytes.
-     * @return string|false Returns the data read from the stream, false if
-     *     unable to read or if an error occurs.
+     * @return string|null Returns the data read from the stream, or null if no bytes
+         are available and the stream is in non-blocking mode.
```
mtdowling: Missing a leading `*`

weierophinney:
@simensen and/or @pmjones — ready to merge.

proposed/http-message.md
```diff
@@ -1133,8 +1137,9 @@ interface StreamableInterface
      * @param int $length Read up to $length bytes from the object and return
      *     them. Fewer than $length bytes may be returned if underlying stream
      *     call returns fewer bytes.
-     * @return string|false Returns the data read from the stream, false if
-     *     unable to read or if an error occurs.
+     * @return string|null Returns the data read from the stream, or null if no bytes
```
Tobion: is there a semantic difference between `''` and `null`? otherwise this seems strange to me.

weierophinney: Please see the discussion; this was suggested by @mtdowling based on similar verbiage in python specs.

Tobion: Then we should add that '' means OEF if $length > 0.

mtdowling:
I think we should remove the EOF claims as I think you need to call `read` and `eof` in order to truly know if a stream is EOF. If we required that the `StreamableInterface` tell callers of `read()` whether the stream is EOF, then it would require a call to `eof()` on each read (which would typically mean a call to PHP's `feof()`). That's too much overhead to be practical IMO.

I would really like to mimic whatever PHP does in these situations to be consistent with PHP... If, in fact, PHP only returns `false` when there's a parameter error, then we can skip that and throw an exception. If there are other cases for false, then we need to know what they are...

- `stream_get_contents` is a function used to read the remainder of a stream into memory. This seems like a good place to start because it needs to read from the stream until it encounters EOF, which means it will have to exercise reading from a string, detecting errors, and determining EOF: https://github.com/php/php-src/blob/dd16a7c82045383c8556a67c649a6b503099d040/ext/standard/streamsfuncs.c#L396

- This function uses `php_stream_copy_to_mem` to actually perform the read. It's actually a macro defined here: https://github.com/php/php-src/blob/4d9a1883aa764e502990488d2e8b9c978be6fbd2/main/php_streams.h#L459.

- This macro uses `_php_stream_copy_to_mem`: https://github.com/php/php-src/blob/4d9a1883aa764e502990488d2e8b9c978be6fbd2/main/streams/streams.c#L1406. The interesting part of `_php_stream_copy_to_mem` is the part that just reads until EOF:

```c
result = zend_string_alloc(max_len, persistent);
ptr = result->val;

while ((ret = php_stream_read(src, ptr, max_len - len)))    {
    len += ret;
    if (len + min_room >= max_len) {
        result = zend_string_extend(result, max_len + step, persistent);
        max_len += step;
        ptr = result->val + len;
    } else {
        ptr += ret;
    }
}
if (len) {
    result = zend_string_truncate(result, len, persistent);
    result->val[len] = '\0';
} else {
    zend_string_free(result);
    result = NULL;
}
```

This function uses `php_stream_read` to read from the stream, and it checks if the return value (the length of the data read) is > 0. Once a zero length read is encountered, it stops reading. If no data could be read before a zero length read is encountered, then `null` is returned. This null value would be returned to the stream_get_contents function, which converts it to an empty string return value.

Let's looks at `fread`'s implementation, including the previously utilized `php_stream_read`. `fread` is defined here: https://github.com/php/php-src/blob/4d9a1883aa764e502990488d2e8b9c978be6fbd2/ext/standard/file.c#L1766. This function validates the provided parameters and returns false if they are in error. There's the false return value.

It then uses `php_stream_read` to read data from the stream: https://github.com/php/php-src/blob/4d9a1883aa764e502990488d2e8b9c978be6fbd2/main/streams/streams.c#L683. This function reads from the stream until the amount of data requested has been retrieved and returns the number of read bytes.  In some cases, the amount data that is read from the stream can return `-1`, which causes the read loop here to terminate: https://github.com/php/php-src/blob/4d9a1883aa764e502990488d2e8b9c978be6fbd2/main/streams/streams.c#L713.

```c
if (!stream->readfilters.head && (stream->flags & PHP_STREAM_FLAG_NO_BUFFER || stream->chunk_size == 1)) {
	toread = stream->ops->read(stream, buf, size);
	if (toread == (size_t) -1) {
		/* e.g. underlying read(2) returned -1 */
		break;
	}
}
```

If no data is read from the stream or an underlying call returns `-1`, then `0` is returned.

Back to `fread`... it fills the buffer with the data read from the stream and then checks the amount of data returned: `Z_STRLEN_P(return_value) = php_stream_read(stream, Z_STRVAL_P(return_value), len);` (you can find the macros here: https://github.com/php/php-src/blob/dcaa79546b25723136e37589e56fc602a503db12/Zend/zend_types.h#L500).

Then it returns the string value:  `Z_STRVAL_P(return_value)[Z_STRLEN_P(return_value)] = 0;` I'm not all that familiar with how these macros work, so I'm not sure if this macro would always return a string or could possibly return `false`. I'm not sure what the `= 0` on the end means either...

It would be great if someone with more knowledge on how this all works could help us... Maybe @ircmaxell?

trowski:
`Z_STRVAL_P(return_value)[Z_STRLEN_P(return_value)] = 0;` is actually just putting a null-byte at the end of the read string because those read functions don't already do that. Since many PHP functions operate on the contents of a string as though it was a C-string, it is important that the string is always null-terminated. `return_value` will always be a string once the function reaches that point. It's allocated a few lines above in `ZVAL_NEW_STR(return_value, zend_string_alloc(len, 0));`.

`fread()` will return an empty string if the stream is at EOF, so the user is required to call `feof()` to see if the stream is at EOF. So if `StreamableInterface::read()` is suppose to return `false` or throw an exception if the stream is at EOF, then yes, `foef()` would have to be called within `read()`, adding overhead since the user will probably be calling `StreamableInterface::eof()` anyway.

It is probably better to mimic PHP's behavior and return an empty string from `read()` if the stream is at EOF.

PR #508, opened 2015-04-11
Use "status code" and "reason phrase" consistently.

pjdietz:
This PR uses "status code" and "reason phrase" consistently throughout the text and docblocks.

I can see a reason to use "Status-Code" and "Reason-Phrase" instead as the RFCs use this style (although only in text, not in the section headers). However, since we don't use "Header-Field", "Field-Name", and "Field-Value" when referring to headers, it seems to odd to use the hyphenated and capped style for "Status-Code" and "Reason-Phrase".

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -956,9 +956,9 @@ namespace Psr\Http\Message;
 interface ResponseInterface extends MessageInterface
 {
     /**
-     * Gets the response Status-Code.
+     * Gets the response status code.
      *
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
:+1:

weierophinney:
Honestly, we're going to likely keep finding little nits like this to pick forever. However, we're in draft mode for the short term, so :+1:.

@simensen and/or @pmjones : ready to merge.

PR #509, opened 2015-04-12
[PSR-7] Always return a string from read()

weierophinney:
Per notes on #504 post-merge, `read()` should return an empty string if no bytes
are available, and not null.


weierophinney:
@mtdowling — please review!

mtdowling:
LGTM

weierophinney:
@simensen or @pmjones - ready to merge!

PR #511, opened 2015-04-13
[PSR-7] Added exception interfaces.

trowski:
This pull request changes the types of exceptions thrown to package specific interfaces, similar to how PSR-6 has package specific exception interfaces. This makes it clearer that the exceptions thrown don't have to be, and really shouldn't be, direct instances of `\InvalidArgumentException` and `\RuntimeException`, but rather exceptions specific to the package. The exceptions implementing these interfaces SHOULD extend the built-in classes `\InvalidArgumentException` and `\RuntimeException`.

I left `Interface` off the names of these exceptions to be consistent with the current version of PSR-6.

Concrete implementations could also be provided for convince in the final distributed package.

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -1769,7 +1769,8 @@ namespace Psr\Http\Message;
  * Exception interface for invalid arguments.
  *
  * When an invalid argument is given to a method it must throw an exception
- * implementing this interface.
+ * implementing this interface. Exception classes implementing this interface
+ * SHOULD extend \InvalidArgumentException.
  */
 interface InvalidArgumentException {}
 ```
@@ -1784,7 +1785,8 @@ namespace Psr\Http\Message;
  * Exception interface for failed operations.
  *
  * When an operation fails the method must thrown an exception implementing
- * this interface.
+ * this interface. Exception classes implementing this interface SHOULD extend
+ * \RuntimeException.
  */
 interface RuntimeException {}
 ```
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -562,7 +562,8 @@ interface MessageInterface
      * @param string $name Case-insensitive header field name.
      * @param string|string[] $value Header value(s).
      * @return self
-     * @throws \InvalidArgumentException for invalid header names or values.
+     * @throws \Psr\Http\Message\InvalidArgumentException for invalid header
+     *     names or values.
      */
     public function withHeader($name, $value);
 
@@ -581,7 +582,8 @@ interface MessageInterface
      * @param string $name Case-insensitive header field name to add.
      * @param string|string[] $value Header value(s).
      * @return self
-     * @throws \InvalidArgumentException for invalid header names or values.
+     * @throws \Psr\Http\Message\InvalidArgumentException for invalid header
+     *     names or values.
      */
     public function withAddedHeader($name, $value);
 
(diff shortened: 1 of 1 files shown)


Tobion:
What is the value of this? For some symfony components we also did this. But usually they are more specific than a  general invalid argument exception.

weierophinney:
This was brought up previously (please look on the ML or in the issues in this repo), and the determination made was that the policy with FIG so far has been *not* to define proposal-specific exceptions.

trowski:
The value comes from being able to catch the exception from a specific action and handle it locally while letting exceptions from other operations bubble up. The example below is somewhat contrived, but helps illustrate my point.

```php
function send(\Psr\Http\Message\RequestInterface $request, \Awesome\Package\ConnectionInterface $connection)
{
    $stream = $request->getBody();
    $stream->seek(0); // seek() can throw a \RuntimeException
    $connection->write($stream->getContents()); // write() can throw a \RuntimeException.
}
```
Code using the above function can catch `\RuntimeException`, but it has no idea where it came from. If seeking failed, maybe a response with status code 500 should be sent. If sending on the connection failed, maybe the exception should bubble up and close the connection.

Specific exceptions help in situations like these. Symfony components are great examples because of the specific exceptions they throw, it's easy to target specific operations.

When the issue was discussed on the mailing list, only `\InvalidArgumentException` was being thrown from any method. Invalid argument exceptions really shouldn't happen at runtime, as it's really more of a coding error and should be fixed, but runtime exceptions are of course more unpredictable.

The other reason I brought it up is for consistency. If PSR-6 has package-specific exception interfaces (including one for invalid argument exceptions), it would seem odd for PSR-7 to not. Since either PSR can be changed right now, I think it would be wise to choose one path or another.

weierophinney:
Interestingly, PSR-3 *does not* define any exceptions, but is ambiguous on that matter as [Section 1.1](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-3-logger-interface.md#11-basics) indicates that the `log()` method "MUST throw a Psr\Log\InvalidArgumentException" if called with an invalid log level constant.

I've been a proponent of exception segregation (the rules we adopted for ZF2 use this heavily, and introduced a pattern used in many libraries). However, when the issue was raised previously, regardless of the type of exception, the opinions were strongly against adding proposal-specific exceptions to the specification.

mtdowling:
:-1: to duplicating SPL interfaces per/PSR. This is a super weird pattern for unchecked exceptions (i.e., `InvalidArgumentException`) that cannot and should not be recovered from: See https://groups.google.com/d/msg/php-fig/XjPytz5d9go/I2a_zED7xCMJ

As for RuntimeExceptions, if you need to wrap these in application specific exceptions, then you can more granularly try/catch calls that have the potential to throw RuntimeExceptions. Implementations can also choose to use the exception interface marker pattern in their frameworks if desired and still adhere to throwing a `\RuntimeException`.

trowski:
Packages implementing PSR-7 can use the exception interface pattern like Symfony and ZF2, but then code catching those exceptions is coupled to that package. I thought if the exception interface was part of the specification you could catch specific exceptions without being coupled to a particular implementation.

I actually don't care for the pattern implemented in the PR, I was just following the style used by PSR-6. I prefer the pattern used by Symfony and ZF2 and use it in my own libraries.

If we want to keep the specification simpler and only require the basic SPL exceptions that sounds reasonable, but then I think it's best to use that philosophy for PSR-6 and future PSRs.

GrahamCampbell:
I agree with @mtdowling.

PR #512, opened 2015-04-13
Renamed UploadedFileInterface to RequestFileInterface

webmozart:
I reread PSR-7 after the latest changes and found that I made a mistake when creating the file upload PR. I'm coming from Symfony, where requests are solely regarded at from the receiving end (the server side). This PSR, however, takes a neutral position: You can write client code just as well as server code.

From the client perspective, the term "uploaded file" does not make sense. You want to send a file with the request, but the file has not been uploaded yet. Therefore I propose to call these files simply "files sent with the request" or "request files":

```php
// does not make sense
$request = Request::create('http://example.com')
    ->withUploadedFiles(array('logo' => new File('/path/to/logo.png')));

$client->send($request);

// better
$request = Request::create('http://example.com')
    ->withFiles(array('logo' => new File('/path/to/file')));

$client->send($request);
```

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -296,11 +296,11 @@ application-specific rules (such as path matching, scheme matching, host
 matching, etc.). As such, the server request can also provide messaging between
 multiple request consumers.
 
-### 1.6 Uploaded files
+### 1.6 Files
 
-`ServerRequestInterface` specifies a method for retrieving a tree of upload
-files in a normalized structure, with each leaf an instance of
-`UploadedFileInterface`.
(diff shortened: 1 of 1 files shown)


weierophinney:
@webmozart This interface is used only with the `ServerRequestInterface`, so the server-side language of "uploaded file" is still relevant.

For client-side applications, you would be manipulating the message body directly, likely with a custom `StreamInterface` implementation (such as the approaches used in Guzzle already).

mtdowling:
Yeah, :-1:. As a client library author, I have no idea how I'd use this with a client. We already have the body of a request that functions as a stream of data.

PR #513, opened 2015-04-13
Renamed UploadedFileInterface::getStream() to getBody()

webmozart:
When #498 was merged, one issue was left open: Originally, @m00t proposed to rename `UploadedFileInterface::getStream()` to `getBody()` to be consistent with `MessageInterface::getBody()`.

He also found a counter argument:

> I thought about getStream() vs getBody() naming again and found an argument for getStream(). "body" is a domain term when we speak about http messages, but there is no such term when we speak about uploaded files.

However:

* The "body of a file" or "file body" is a proper expression.
* The PSR becomes more intuitive to use if methods with the same intent have the same name.
* Streams can be cast to string, so `echo $file->getBody()` works as expected.

Therefore I propose to rename `getStream()` to `getBody()`.

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -396,10 +396,10 @@ operations will work regardless of environment. In particular:
 - `move($path)` is provided as a safe and recommended alternative to calling
   `move_uploaded_file()` directly on the temporary upload file. Implementations
   will detect the correct operation to use based on environment.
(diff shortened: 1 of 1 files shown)


weierophinney:
Since the method returns a `StreamInterface` instance, we can also argue that it's named after what it's returning. I'd like to keep it as-is.

PR #518, opened 2015-04-13
[PSR-7] Removed verbiage "if any" from `getUri()`

weierophinney:
Per a point made [on the mailing list](https://groups.google.com/d/msgid/php-fig/8f3c71aa-bbfb-4178-9806-8d740425ef0e%40googlegroups.com), this patch removes the phrase "if any" from `RequestInterface::getUri()` to remove any confusion about whether or not it should return a `UriInterface` instance.


weierophinney:
@simensen and/or @pmjones — please merge (per feedback on the ML).

PR #519, opened 2015-04-13
[PSR-7] Adding cookies MUST NOT update headers or server params

weierophinney:
Per discussion with @simensen, added verbiage to indicate that setting cookie parameters MUST NOT update headers or server params. This mirrors language that already exists in `withQueryParams()` (which specifies implementations MUST NOT update the URI instance based on the changes).


weierophinney:
Ping @simensen for review.

simensen:
:+1:

weierophinney:
@pmjones or @simensen — let's merge, then. :)

PR #520, opened 2015-04-13
[PSR-6] Documentation bug fixes

AndrewCarterUK:
There were multiple references to previous names of the interfaces in the documentation which have been fixed.


Crell:
Unfortunately this conflicted with the other PR, which I saw/merged first. :frowning: I went ahead and made the changes to the meta document myself.  We should be good now.  If you spot other inconsistencies where the wrong name is used, please file a new PR.  Thanks!

PR #522, opened 2015-04-16
Remove RequestInterface::getHeaders, getHeader, and getHeaderLine

pjdietz:
The new behavior in `RequestInterface::withUri()` from PR #500 makes the overrides
in `Request::getHeaders()`, `getHeader()`, `getHeaderLine()` unecessary and incompatible with the description for `withUri()`.

Specifically, the problematic section is:

> You can opt-in to preserving the original state of the Host header by
> setting `$preserveHost` to `true`. When `$preserveHost` is set to
> `true`, the returned request will not update the Host header of the
> returned message -- even if the message contains no Host header. This
> means that a call to `getHeader('Host')` on the original request MUST
> equal the return value of a call to `getHeader('Host')` on the returned
> request.

Here's an example:

```php
// $request1 has no explictily set Host header.
$request1->getHeader('Host'); // [] Not set, so empty array

$request2 = $request1->withUri(new Uri('http://localhost/'), true);
$request2->getHeader('Host'); // ['localhost'] Read from the URI

$request3 = $request2->withUri(new Uri('http://someotherhost/'), true);
$request3->getHeader('Host'); // ['someotherhost'] Read from the URI
```

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -649,64 +649,6 @@ namespace Psr\Http\Message;
 interface RequestInterface extends MessageInterface
 {
     /**
-     * Extends MessageInterface::getHeaders() to provide request-specific
-     * behavior.
-     *
-     * Retrieves all message headers.
-     *
-     * This method acts exactly like MessageInterface::getHeaders(), with one
-     * behavioral change: if the Host header has not been previously set, the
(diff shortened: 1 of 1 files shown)


pjdietz:
Closing in favor of #523

PR #523, opened 2015-04-17
Update description for RequestInterface::withUri().

pjdietz:
This PR updates the description for `RequestInterface::withUri()`. The new description uses "MUST" in place of "will" and removes a line from the description that was inconsistent with `Message::getHeaders()`, `getHeader()`, and `getHeaderLines()` when the original request did not include an explicitly set `Host` header.

Specifically, the problematic section is:

> You can opt-in to preserving the original state of the Host header by setting $preserveHost to true. When $preserveHost is set to true, the returned request will not update the Host header of the returned message -- even if the message contains no Host header. **This means that a call to getHeader('Host') on the original request MUST equal the return value of a call to getHeader('Host') on the returned request.**

Here's an example:

```php
// $request1 has no explictily set Host header.
$request1->getHeader('Host'); // [] Not set, so empty array

$request2 = $request1->withUri(new Uri('http://localhost/'), true);
$request2->getHeader('Host'); // ['localhost'] Read from the URI

$request3 = $request2->withUri(new Uri('http://someotherhost/'), true);
$request3->getHeader('Host'); // ['someotherhost'] Read from the URI
```

I think we can drop this last sentence and avoid the inconsistency. With the stronger language ("MUST" in place of "will", etc.), the description should be clear enough.

Alternatively, we could replace the last sentence with a  a more explicit explanation, but they end up being really wordy and convoluted. Something like: 

> You can opt-in to preserving the original state of the Host header by setting $preserveHost to true. When $preserveHost is set to true, the returned request will not update the Host header of the returned message -- even if the message contains no Host header. **If a request has an explicitly set Host header, a call to getHeader('Host') on that request MUST equal the return value of a call to getHeader('Host') on the request returned by withUri() with $preserveHost set to true.**

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -676,64 +676,6 @@ namespace Psr\Http\Message;
 interface RequestInterface extends MessageInterface
 {
     /**
-     * Extends MessageInterface::getHeaders() to provide request-specific
-     * behavior.
-     *
-     * Retrieves all message headers.
-     *
-     * This method acts exactly like MessageInterface::getHeaders(), with one
-     * behavioral change: if the Host header has not been previously set, the
-     * method MUST attempt to pull the host component of the composed URI, if
-     * present.
-     *
-     * @see MessageInterface::getHeaders()
-     * @see UriInterface::getHost()
-     * @return array Returns an associative array of the message's headers. Each
-     *     key MUST be a header name, and each value MUST be an array of strings.
-     */
-    public function getHeaders();
-
-    /**
-     * Extends MessageInterface::getHeader() to provide request-specific
-     * behavior.
-     *
-     * This method acts exactly like MessageInterface::getHeader(), with
-     * one behavioral change: if the Host header is requested, but has
-     * not been previously set, the method MUST attempt to pull the host
-     * component of the composed URI, if present.
-     *
-     * @see MessageInterface::getHeader()
-     * @see UriInterface::getHost()
-     * @param string $name Case-insensitive header field name.
-     * @return string[] An array of string values as provided for the given
-     *    header. If the header does not appear in the message, this method MUST
-     *    return an empty array.
-     */
-    public function getHeader($name);
-
-    /**
-     * Extends MessageInterface::getHeaderLine() to provide request-specific
-     * behavior.
-     *
-     * This method returns all of the header values of the given
-     * case-insensitive header name as a string concatenated together using
-     * a comma.
-     *
-     * This method acts exactly like MessageInterface::getHeaderLine(), with
-     * one behavioral change: if the Host header is requested, but has
-     * not been previously set, the method MUST attempt to pull the host
-     * component of the composed URI, if present.
-     *
-     * @see MessageInterface::getHeaderLine()
-     * @see UriInterface::getHost()
-     * @param string $name Case-insensitive header field name.
-     * @return string|null A string of values as provided for the given header
-     *    concatenated together using a comma. If the header does not appear in
-     *    the message, this method MUST return a null value.
-     */
-    public function getHeaderLine($name);
-
-    /**
      * Retrieves the message's request target.
      *
      * Retrieves the message's request-target either as it will appear (for
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -781,18 +781,15 @@ interface RequestInterface extends MessageInterface
     /**
      * Returns an instance with the provided URI.
      *
-     * This method will update the Host header of the returned request by
+     * This method MUST update the Host header of the returned request by
      * default if the URI contains a host component. If the URI does not
-     * contain a host component, any pre-existing Host header will be carried
+     * contain a host component, any pre-existing Host header MUST be carried
      * over to the returned request.
      *
      * You can opt-in to preserving the original state of the Host header by
      * setting `$preserveHost` to `true`. When `$preserveHost` is set to
-     * `true`, the returned request will not update the Host header of the
-     * returned message -- even if the message contains no Host header. This
-     * means that a call to `getHeader('Host')` on the original request MUST
-     * equal the return value of a call to `getHeader('Host')` on the returned
-     * request.
+     * `true`, the returned request MUST NOT update the Host header of the
+     * returned message -- even if the message contains no Host header.
      *
      * This method MUST be implemented in such a way as to retain the
      * immutability of the message, and MUST return an instance that has the
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -669,6 +669,9 @@ namespace Psr\Http\Message;
  * - Headers
  * - Message body
  *
+ * During construction, implementations MUST attempt to set the Host header from
+ * a provided URI if no Host header is provided as well.
+ * 
  * Requests are considered immutable; all methods that might change state MUST
  * be implemented such that they retain the internal state of the current
  * message and return an instance that contains the changed state.
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -815,11 +815,16 @@ interface RequestInterface extends MessageInterface
      *
      * You can opt-in to preserving the original state of the Host header by
      * setting `$preserveHost` to `true`. When `$preserveHost` is set to
-     * `true`, the returned request MUST NOT update the Host header of the
-     * returned message -- unless the original messages contains no Host header.
-     * This means that a call to `getHeader('Host')` on the original request
-     * MUST equal the return value of a call to `getHeader('Host')` on the
-     * returned request.
+     * `true`, this method interacts with the Host header in the following ways:
+     *
+     * - If the the Host header is missing or empty, and the new URI contains 
+     *   a host component, this method MUST update the Host header in the returned 
+     *   request.
+     * - If the Host header is missing or empty, and the new URI does not contain a
+     *   host component, this method MUST NOT update the Host header in the returned
+     *   request.
+     * - If a Host header is present and non-empty, this method MUST NOT update
+     *   the Host header in the returned request.
      *
      * This method MUST be implemented in such a way as to retain the
      * immutability of the message, and MUST return an instance that has the
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -146,6 +146,27 @@ present in the composed `UriInterface`, the value from the URI should be used.
 If a Host header is explicitly provided to the request instance, that value will
 be preferred.
 
+`RequestInterface::withUri()` also interacts with the Host header. By default,
+this method replaces the returned request's Host header with a Host header
+matching the host component of the passed `UriInterface`.
+
+You can opt-in to preserving the original state of the Host header by passing
+`true` for the second (`$preserveHost`) argument. When this argument is set to
+`true`, the returned request will not update the Host header of the returned
+message -- unless the message contains no Host header.
+
+This table illustrates what `getHeaderLine('Host')` will return for a request
+returned by `withUri()` with the `$preserveHost` argument set to `true` for
+various initial requests and URIs.
+
+Request Host header | Request host component | URI host component | Result
+--------------------|------------------------|--------------------|--------
+''                  | ''                     | ''                 | ''
+''                  | foo.com                | ''                 | foo.com
+''                  | foo.com                | bar.com            | foo.com
+foo.com             | ''                     | bar.com            | foo.com
+foo.com             | bar.com                | baz.com            | foo.com
+
 ### 1.3 Streams
 
 HTTP messages consist of a start-line, headers, and a body. The body of an HTTP
@@ -789,7 +810,10 @@ interface RequestInterface extends MessageInterface
      * You can opt-in to preserving the original state of the Host header by
      * setting `$preserveHost` to `true`. When `$preserveHost` is set to
      * `true`, the returned request MUST NOT update the Host header of the
-     * returned message -- even if the message contains no Host header.
+     * returned message -- unless the original messages contains no Host header.
+     * This means that a call to `getHeader('Host')` on the original request
+     * MUST equal the return value of a call to `getHeader('Host')` on the
+     * returned request.
      *
      * This method MUST be implemented in such a way as to retain the
      * immutability of the message, and MUST return an instance that has the
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -135,25 +135,21 @@ for retrieving such multi-valued headers.
 
 ##### Host header
 
-In requests, the Host header typically mirrors the host segment of the URI, as
+In requests, the `Host` header typically mirrors the host segment of the URI, as
 well as the host used when establishing the TCP connection. However, the HTTP
-specification allows the Host header to differ from each of the two.
+specification allows the `Host` header to differ from each of the two.
 
-The `RequestInterface` overrides the `MessageInterface::getHeader()`,
-`MessageInterface::getHeaders()`, and `MessageInterface::getHeaderLine()`
-methods to indicate that if no Host header is present, but a host segment is
-present in the composed `UriInterface`, the value from the URI should be used.
-If a Host header is explicitly provided to the request instance, that value will
-be preferred.
+During construction, implementations MUST attempt to set the `Host` header from
+a provided URI if no `Host` header is provided as well.
 
-`RequestInterface::withUri()` also interacts with the Host header. By default,
-this method replaces the returned request's Host header with a Host header
-matching the host component of the passed `UriInterface`.
+`RequestInterface::withUri()` will, by default, replace the returned request's
+`Host` header with a `Host` header matching the host component of the passed
+`UriInterface`.
 
-You can opt-in to preserving the original state of the Host header by passing
+You can opt-in to preserving the original state of the `Host` header by passing
 `true` for the second (`$preserveHost`) argument. When this argument is set to
-`true`, the returned request will not update the Host header of the returned
-message -- unless the message contains no Host header.
+`true`, the returned request will not update the `Host` header of the returned
+message -- unless the message contains no `Host` header.
 
 This table illustrates what `getHeaderLine('Host')` will return for a request
 returned by `withUri()` with the `$preserveHost` argument set to `true` for
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -159,13 +159,19 @@ This table illustrates what `getHeaderLine('Host')` will return for a request
 returned by `withUri()` with the `$preserveHost` argument set to `true` for
 various initial requests and URIs.
 
-Request Host header | Request host component | URI host component | Result
---------------------|------------------------|--------------------|--------
-''                  | ''                     | ''                 | ''
-''                  | foo.com                | ''                 | foo.com
-''                  | foo.com                | bar.com            | foo.com
-foo.com             | ''                     | bar.com            | foo.com
-foo.com             | bar.com                | baz.com            | foo.com
+Request Host header<sup>[1](#rhh)</sup> | Request host component<sup>[2](#rhc)</sup> | URI host component<sup>[3](#uhc)</sup> | Result
+----------------------------------------|--------------------------------------------|----------------------------------------|--------
+''                                      | ''                                         | ''                                     | ''
+''                                      | foo.com                                    | ''                                     | foo.com
+''                                      | foo.com                                    | bar.com                                | foo.com
+foo.com                                 | ''                                         | bar.com                                | foo.com
+foo.com                                 | bar.com                                    | baz.com                                | foo.com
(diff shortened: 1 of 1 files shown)


mtdowling:
I don't think there's an inconsistency in the existing language. Removing this line from the description of the method can actually has unintended side-effects on the Host header:

```
When `$preserveHost` is set to
`true`, the returned request will not update the Host header of the
returned message -- even if the message contains no Host header. This
means that a call to `getHeader('Host')` on the original request MUST
equal the return value of a call to `getHeader('Host')` on the returned
request.
```

I think that needs to stay in the spec. If a user provides this flag as `true` and you detect that the Host header is not set on the current message but the host component is set on the current URI, then the created message will use the host component of the current URI (not the newly utilized URI) as the Host header. This ensures that the Host header is the same for both message which is what a user asked for when the `$preserveHost` flag is set.

Here's how it should work IMO:

```php
// $request1 has no explictily set Host header.
$request1->getHeader('Host'); // [] Not set, so empty array

$request2 = $request1->withUri(new Uri('http://localhost/'), true);
$request2->getHeader('Host'); // ['localhost'] Read from the URI (but this is irrelevant)

$request3 = $request2->withUri(new Uri('http://someotherhost/'), true);
$request3->getHeader('Host'); // ['localhost'] Read from the Host header that was added
```

pjdietz:
Ah, ok, thanks! I think I understand now. I'm reading "update the Host header" to mean something like "change the explicitly set header" rather than "change the return value of `getHeader('Host')`".

I'm also reading "contains no Host header" to mean that a call to `hasHeader('Host')` will return false, not that a call to `getHeader('Host')` will return a non-empty array.

The phrase "even if the message contains no Host header" still seems problematic though since these two `getHeader('Host')` will still be different:

```php
// $request1 has no explicitly set Host header.
$request1->getHeader('Host'); // [] Not set, so empty array

$request2 = $request1->withUri(new Uri('http://localhost/'), true);
$request2->getHeader('Host'); // ['localhost'] Read from the URI (but this is irrelevant)
```

Should this maybe be "unless the original message contains no Host header"? I don't think we want to see this, which is what it currently describes:

```php
// $request1 has no explicitly set Host header.
$request1->getHeader('Host'); // [] Not set, so empty array

$request2 = $request1->withUri(new Uri('http://localhost/'), true);
$request2->getHeader('Host'); // [] Still not set
```

mtdowling:
> Should this maybe be "unless the original message contains no Host header"? I don't think we want to see this, which is what it currently describes:

Yeah, I think you're right about that one. Maybe the permutations should be spelled out in the docblock?

Does this look right:

Calling withUri() with preserveHost:

Request Host header | Request host component | URI host component | Result
---------------------------   | --------------------------------- | ----------------------------| ----------
'' | '' | '' | ''
'' | foo.com | '' | foo.com
'' | foo.com | bar.com | foo.com
foo.com | '' | bar.com | foo.com
foo.com | bar.com | baz.com | foo.com

pjdietz:
Yup! That table looks right to me. If it looks like too much to fit in the docblock, maybe include it in the "host header" subsection of 1.2?

pjdietz:
I've added a new commit where I added this table to section 1.2 along with some other language describing `withUri`'s interaction with the Host header.

The commit also restores that last line "This means that..." and changes the part about no host header from "even if" to "unless". The whole paragraph reads:

> You can opt-in to preserving the original state of the Host header by setting `$preserveHost` to `true`. When `$preserveHost` is set to `true`, the returned request MUST NOT update the Host header of the returned message -- unless the original messages contains no Host header. This means that a call to `getHeader('Host')` on the original request MUST equal the return value of a call to `getHeader('Host')` on the returned request.

proposed/http-message.md
```diff
@@ -146,6 +146,27 @@ present in the composed `UriInterface`, the value from the URI should be used.
...
+message -- unless the message contains no Host header.
+
+This table illustrates what `getHeaderLine('Host')` will return for a request
+returned by `withUri()` with the `$preserveHost` argument set to `true` for
+various initial requests and URIs.
+
+Request Host header | Request host component | URI host component | Result
+--------------------|------------------------|--------------------|--------
+''                  | ''                     | ''                 | ''
+''                  | foo.com                | ''                 | foo.com
```
Tobion: Do I see that correctly, that `$preserveHost` does actually NOT preserve the host? It changes the host header anyway?

Tobion:
Btw, the given table does not show all permutations.

weierophinney:
> Btw, the given table does not show all permutations.

@Tobion — comments like that are really, really unhelpful. A better phrasing would be, "The given table does not show all permutations; it needs to add …", followed by the permutations you think are missing. When you omit that information, you're making the rest of us *guess* what your intentions are, which, as I've discovered, can often lead down unintended paths.

So, what permutations do you feel are missing, @Tobion ?

proposed/http-message.md
weierophinney: > Do I see that correctly, that $preserveHost does actually NOT preserve the host? It changes the host header anyway?

Not as currently specified: "When `$preserveHost` is set to `true`, the returned request will not update the Host header of the returned message -- even if the message contains no Host header."

proposed/http-message.md
```diff
@@ -146,6 +146,27 @@ present in the composed `UriInterface`, the value from the URI should be used.
...
+You can opt-in to preserving the original state of the Host header by passing
+`true` for the second (`$preserveHost`) argument. When this argument is set to
+`true`, the returned request will not update the Host header of the returned
+message -- unless the message contains no Host header.
+
+This table illustrates what `getHeaderLine('Host')` will return for a request
+returned by `withUri()` with the `$preserveHost` argument set to `true` for
+various initial requests and URIs.
+
+Request Host header | Request host component | URI host component | Result
```
weierophinney: What is meant by "Request host component"? Is this the host component of the currently compose URI instance? If so, this needs to be made more clear.

proposed/http-message.md
weierophinney: @Tobion — a follow-up: `RequestInterface` notes in its `getHeader*()` methods:

> if the Host header is requested, but has not been previously set, the method MUST attempt to pull the host component of the composed URI, if present.

This is why you're seeing the result of `foo.com`: the composed URI has a host component.

That said: @pjdietz, this example is wrong. If you set a new URI without a host component, and do not have a `Host` header currently, the behavior should be to have an empty `Host` header.

pjdietz:
> That said: @pjdietz, this example is wrong. If you set a new URI without a host component, and do not have a Host header currently, the behavior should be to have an empty Host header.

@weierophinney That's consistent with my original reading (see the first comment on this PR), but @mtdowling clarified, and I've updated to go by his interpretation since this builds off of the functionality added in his PR.

Part of the confusion here is that it's ambiguous what it means to **have** a host header. I originally took this to mean something like "has an explicitly set Host header in the header bag." (or, `hasHeader('Host') === true`). @mtdowling takes this to mean something more like "has a way to return a Host header from a call to `getHeader('Host')` (either from the header bag or derived from the Uri).

I think the intention in PR #500 also was that "preserving the host header" doesn't mean preserving the header in the header bag so much as it means preserving the return value of `getHeader('Host')`.

(Michael, please let me know if I'm misinterpreting or misstating anything here.)

Under this second reading, you would get this functionality:

```php
// $request1 has no explicitly set Host header.
$request1->getHeader('Host'); // [] Not set, so empty array

$request2 = $request1->withUri(new Uri('http://localhost/'), true);
$request2->getHeader('Host'); // ['localhost'] Read from the URI

$request3 = $request2->withUri(new Uri('/'), true);
$request3->getHeader('Host'); // ['localhost'] Read from the previously set URI.
```

If `$request2->getHeader('Host')` returns `['localhost']` and `$request3->getHeader('Host')` returns `[]`, then this line is not satisfied:

> This means that a call to getHeader('Host') on the original request MUST equal the return value of a call to getHeader('Host') on the returned request.

If we want $request3 to have a different return value for `getHeader('Host')`, we need to lose that line. But that changes the meaning of how we're "preserving" the host.

proposed/http-message.md
pjdietz: Yes. "Request host component" is the return value for `$request->getUri()->getHost()`. Sorry trying to cram a lot into a tiny table.

weierophinney:
@pjdietz, @mtdowling, and @evert : I think there may be a simpler way to clarify this:

- Modify `RequestInterface::withUri()` to state that if a host component is present in the provided URI, it MUST **set** the `Host` header UNLESS the `$preserveHost` flag is set to `true`.

Note the emphasis: the `withUri()` method would be responsible for **setting** the `Host` header in default use cases. In most cases, then, the `getHeader*()` method overrides in `RequestInterface` are unnecessary, but by keeping them present, we preserve the use case that @evert has noted previously (of being opportunistic in providing a `Host` header from the URI if none is set currently).

We can still provide the table, if necessary, but I think this would simplify the verbiage and make the intention blatantly clear. Thoughts?

weierophinney:
Ha! @pjdietz already has this covered:

> This method MUST update the Host header of the returned request

With that in place, the only bit I'm not satisfied with is the table headers. I think in that case, footnotes might make sense. I'll see if I can figure something like that out and provide a PR to @pjdietz .

pjdietz:
I think we may still need more clarity on what `$preserveHost` is actually preserving. Is it clear enough as is that it's the returned value of `getHeader('Host')` we want to preserve, and not the header in the header bag (if there is one)?

> You can opt-in to preserving the original state of the Host header by setting $preserveHost to true. When $preserveHost is set to true, the returned request MUST NOT update the Host header of the returned message -- unless the original messages contains no Host header. This means that a call to getHeader('Host') on the original request MUST equal the return value of a call to getHeader('Host') on the returned request.

weierophinney:
@pjdietz — here's my take on the table; all I changed was adding footnotes for each column head (except the last): https://github.com/weierophinney/fig-standards/blob/hotfix/host-uri-connection/proposed/http-message.md#host-header

pjdietz:
Thank you! I pulled your footnotes in, merged, and and pushed it up.

weierophinney:
> I think we may still need more clarity on what $preserveHost is actually preserving.

Personally, I think the current verbiage in this PR is acceptable. The only way I can see to make it more clear is to delineate each of the paths precisely; here's a possible take on that:

```php
    /**
     * Returns an instance with the provided URI.
     * 
     * This method interacts with the Host header in the following ways:
     * 
     * - If the `$preserveHost` flag is `false` and the URI contains a host
     *   component, this method MUST update the Host header in the returned
     *   request.
     * - If the `$preserveHost` flag is `false` and the URI does not contain a
     *   host component, this method MUST NOT update the Host header in the
     *   returned request.
     * - If the `$preserveHost` flag is `true`, the Host header is missing or
     *   empty, and the new URI contains a host component, this method MUST
     *   update the Host header in the returned request.
     * - If the `$preserveHost` flag is `true`, the Host header is missing or
     *   empty, and the new URI does not contain a host component, this method
     *   MUST NOT update the Host header in the returned request.
     * - If the `$preserveHost` flag is `true`, and a Host header is present and
     *   non-empty, this method MUST NOT update the Host header in the returned
     *   request.
     * 
     * This method MUST be implemented in such a way as to retain the
     * immutability of the message, and MUST return an instance that has the new
     * UriInterface instance.
     *
     * @link http://tools.ietf.org/html/rfc3986#section-4.3
     * @param UriInterface $uri New request URI to use.
     * @param bool $preserveHost Preserve the original state of the Host header;
     *     defaults to false.
     * @return self
     */
```

Thoughts? @pjdietz , if you think this makes sense, feel free to copy/paste into the PR.

weierophinney:
Pinging @mtdowling , @simensen , @evert, and @Tobion — what are your thoughts on the proposed change to the `withUri()` docblock per my previous comment? Would that make the interactions more clear/less ambiguous?

Next question: with the changes proposed here by @pjdietz , the verbiage is that the method MUST update the `Host` header now. With that in mind, do we *need* the `getHeader*()` override methods in `RequestInterface` still? The only situation where those might come into play are when the URI in place is the one injected at instantiation; otherwise, if `withUri()` is used, the `Host` header will pretty much be guaranteed to be present, unless you specifically opt-out via the `$preserveHost` flag.

Thoughts?

pjdietz:
Let's see what others have to say. This makes it clearer, but also makes it look a little scary. Another thought it so split the difference, since the default behavior is pretty straight forward.

``` 
/**
 * Returns an instance with the provided URI.
 *
 * This method MUST update the Host header of the returned request by
 * default if the URI contains a host component. If the URI does not
 * contain a host component, any pre-existing Host header MUST be carried
 * over to the returned request.
 *
 * You can opt-in to preserving the original state of the Host header by
 * setting `$preserveHost` to `true`. When `$preserveHost` is set to
 * `true`, this method interacts with the Host header in the following ways:
 *
 * - If the the Host header is missing or empty, and the new URI contains 
 *   a host component, this method MUST update the Host header in the returned 
 *   request.
 * - If the Host header is missing or empty, and the new URI does not contain a
 *   host component, this method MUST NOT update the Host header in the returned
 *   request.
 * - If and a Host header is present and non-empty, this method MUST NOT update
 *   the Host header in the returned request.
 *
 * This method MUST be implemented in such a way as to retain the
 * immutability of the message, and MUST return an instance that has the
 * new UriInterface instance.
 *
 * @link http://tools.ietf.org/html/rfc3986#section-4.3
 * @param UriInterface $uri New request URI to use.
 * @param bool $preserveHost Preserve the original state of the Host header.
 * @return self
 */
```

Thoughts?

weierophinney:
@pjdietz That looks much cleaner, and consolidates the cases nicely.

One point: last bullet point should read:

> If a Host header is present and non-empty…

(Omit the "and", essentially.)

pjdietz:
Great! I updated the PR with the new docblock (minus the extra "and").

mtdowling:
:+1:  I like this update.

> With that in mind, do we need the getHeader*() override methods in RequestInterface still? The only situation where those might come into play are when the URI in place is the one injected at instantiation; otherwise, if withUri() is used, the Host header will pretty much be guaranteed to be present, unless you specifically opt-out via the $preserveHost flag.

:+1: I think it would be safe to remove those overrides. While we could suggest that implementations set a Host header when instantiating a request if possible, I would still need to compute the host header rather than expect the Host header to always be present. I think that's reasonable.

evert:
While I think it's a decent update. I guess I'm still a little confused about the need for this.

In the original approach, we did things in a way that the `getHeader()` functions would return the Host header (even if it was not explicitly set).

By just having that behavior, the implication is that calling `withUri()` _already_ has the desired effect you are looking for: it influences the value of the `Host` header. 

Only when the user explicitly chooses to override this header, because the user wants this header to actually be different from the Host in the Uri object, it changes.

The drawback of the proposed approach is indeed that there's a new flag, and it introduces an inconsistency with how `getRequestTarget()` behaves (which does behave similar to the old Host behavior).

So in some ways I feel this is a step backward, and the benefit of this is not entirely clear to me. Is it supposed to be a benefit to the user, or is it just to make it a bit easier for the implementer? The latter is not that important in case I think. API is more important than implementation.

evert:
I also think that this is wrong:

> If the `$preserveHost` flag is `true`, the Host header is missing or
> empty, and the new URI does not contain a host component, this method
> MUST NOT update the Host header in the returned request.

Is there ever a case where you are making a rqeuest object, and there's no target tcp host? I think the `host` in the URI object should never be empty?

weierophinney:
@evert —

> I think the host in the URI object should never be empty?

We currently allow a URI instance to have an empty host. One rationale is to allow iterative creation:

```php
$uri = (new Uri())
    ->withPath('/foo')
    ->withScheme('https')
    ->withQuery('bar=baz')
    ->withHost('example.com');
```

The other rationale is to allow resolving a relative URI (path + query + fragment only).

> Is it supposed to be a benefit to the user, or is it just to make it a bit easier for the implementer?

Both. For the user, there are fewer permutations to track, making understanding the behavior easier. For the implementor, there's no need to test if the `Host` header is present in the various `getHeader*()` methods; if `withUri()` has been called, and the URI had a host component, it's set. This reduces execution paths in those methods, making them less prone to error.

The only situation that can now lead to a potential unexpected result is in *concrete* implementations, specifically in *how the instance is created*. If the implementation's constructor or factory does not set the Host header and/or does not update it from the supplied URI, then there's the possibility of an empty Host header. This leads to two questions on my part:

- Does the specification need to worry about that? One could argue that this is accounted for in the spec, if we allow an empty Host header (which we *do*, though it's quite difficult to accomplish currently if the URI has a host component).
- If it does matter, could we do specify behavior in the interface class-level docblock? Something along the lines of, "During construction, implementations MUST attempt to set the Host header from a provided URI if no Host header is provided as well." This could address it succinctly.

Most of the confusion I'm seeing is due to people trying to understand the relation between `withUri()` and the various `getHeader*()` methods; I think the method overrides are complicating the message, which is why I suggested we could remove them.

pjdietz:
I've added new commits that remove the `getHeader*` overrides and describe the new behavior in section 1.2. I also added "During construction, implementations MUST attempt to set the Host header from a provided URI if no Host header is provided as well." to section 1.2 and to the docblock for `RequestInterface`. (Do we need this in ServerRequestInterface as well? There's a lot there already, so I left it out.)

I think these changes do a nice job of simplifying the interfaces. As @mtdowling and @weierophinney  pointed out, the changes put the burden of setting the initial Host header on the implementation, but this seems pretty reasonable, and is easy enough to indicate clearly in the spec.

We also no longer have the somewhat weird behavior that `$request->hasHeader('Host')` could return `false` while `$request->getHeader('Host')` returns `['localhost']`.

weierophinney:
@pjdietz Can you rebase from current master, please? I see a lot of "changes" that are items that are actually in the current master, but were likely not in the commit from which you branched, which is making finding the relevant pieces a bit harder.

weierophinney:
> Do we need this in ServerRequestInterface as well?

No. We may want to add an `{@inheritdoc}` to that interface's class-level docblock, though, to be certain.

One note on that: when you marshal a `ServerRequestInterface` instance, you'll typically be aggregating incoming headers, which will include the `Host` header, so it's relatively unlikely that the issue will even come up in such a situation.

weierophinney:
On reviewing of the relevant changes, I'm :+1:, as I think they simplify both understanding expected behavior for consumers, and simplify implementation.

evert:
I'm going to put the issue about the flag vs. overriding the header methods aside, because while I feel that the new solution is a actually a bit worse, it's not worth having a long argument about it.

However, I do want to comment on this:

> We currently allow a URI instance to have an empty host. One rationale is to allow iterative creation:

I think that's wrong. I think the uri should be 'fully materialized' after construction and not ever be allowed to be in an invalid state. A uri without a host is an invalid URI, and the concept of a relative URI does not really apply to the URI object.

We can't really make HTTP requests if we don't know which host we're sending it to. The host in the URI object is canonical, the Host header is simply the value that we send in the body, but it's unrelated to the host being used to create the tcp request.

On the server case, I think the URI object should be filled in with 'reasonable defaults'.

If the URI object is a true value object, I don't think it should ever have a state where it's not valid. Building the object iteratively should be the responsibility of an external builder or factory.

pjdietz:
Ok, rebasing done. Changes and commits look good now.

weierophinney:
@mtdowling Can you have another look, please?

proposed/http-message.md
```diff
@@ -135,16 +135,39 @@ for retrieving such multi-valued headers.
...
-The `RequestInterface` overrides the `MessageInterface::getHeader()`,
-`MessageInterface::getHeaders()`, and `MessageInterface::getHeaderLine()`
-methods to indicate that if no Host header is present, but a host segment is
-present in the composed `UriInterface`, the value from the URI should be used.
-If a Host header is explicitly provided to the request instance, that value will
-be preferred.
+specification allows the `Host` header to differ from each of the two.
+
+During construction, implementations MUST attempt to set the `Host` header from
+a provided URI if no `Host` header is provided as well.
```
mtdowling: I suggest you remove "as well"

proposed/http-message.md
```diff
@@ -642,6 +665,9 @@ namespace Psr\Http\Message;
  * - Headers
  * - Message body
  *
+ * During construction, implementations MUST attempt to set the Host header from
+ * a provided URI if no Host header is provided as well.
```
mtdowling: Same thing about "as well"

mtdowling:
Looks good.

pjdietz:
> I suggest you remove "as well"

All set! (removed both occurrences.)

weierophinney:
@Tobion — can you review, please?

weierophinney:
@simensen and/or @pmjones — Based on the feedback received so far, I think this is ready for merge.

PR #524, opened 2015-04-17
[PSR-7] Update comment references to getHeaderLines() method

ptlis:
Should now be getHeaderLine()


weierophinney:
@simensen and/or @pmjones — :+1: (this is a clarification)

PR #526, opened 2015-04-20
Remove the reason phrase

fabpot:
The reason phrase is totally optional in HTTP 1.1 and should not be relied on. Moreover, it was removed from HTTP 2 altogether. So, I think it makes sense to not add it to PSR-7.

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -1088,7 +1088,7 @@ namespace Psr\Http\Message;
  * each of the following:
  *
  * - Protocol version
- * - Status code and reason phrase
+ * - Status code
  * - Headers
  * - Message body
  *
@@ -1109,43 +1109,19 @@ interface ResponseInterface extends MessageInterface
     public function getStatusCode();
 
     /**
-     * Return an instance with the specified status code, and optionally
-     * reason phrase, for the response.
-     *
-     * If no reason phrase is specified, implementations MAY choose to default
-     * to the RFC 7231 or IANA recommended reason phrase for the response's
-     * status code.
+     * Return an instance with the specified status code for the response.
      *
      * This method MUST be implemented in such a way as to retain the
      * immutability of the message, and MUST return an instance that has the
-     * updated status and reason phrase.
+     * updated status.
      *
      * @link http://tools.ietf.org/html/rfc7231#section-6
      * @link http://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml
      * @param int $code The 3-digit integer result code to set.
-     * @param null|string $reasonPhrase The reason phrase to use with the
-     *     provided status code; if none is provided, implementations MAY
-     *     use the defaults as suggested in the HTTP specification.
      * @return self
      * @throws \InvalidArgumentException For invalid status code arguments.
      */
-    public function withStatus($code, $reasonPhrase = null);
-
-    /**
-     * Gets the response reason phrase, a short textual description of the
-     * status code.
-     *
-     * Because a reason phrase is not a required element in a response
-     * status line, the reason phrase value MAY be null. Implementations MAY
-     * choose to return the default RFC 7231 recommended reason phrase (or those
-     * listed in the IANA HTTP Status Code Registry) for the response's
-     * status code.
-     *
-     * @link http://tools.ietf.org/html/rfc7231#section-6
-     * @link http://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml
-     * @return string|null Reason phrase, or null if unknown.
-     */
-    public function getReasonPhrase();
+    public function withStatus($code);
 }
 ```
 


fabpot:
Sorry for this late proposal but I totally missed it previously.

samdark:
HTTP 1.1 is still there + it's optional in the method signature. I don't see any problem with it.

fabpot:
`getReasonPhrase` is not optional, so everyone must implement it. And of course, if you implement a getter, you also need the setter. But for interoperability, and because there is no added value in this reason phrase, it clutters the interface for no real benefits.

Just my 2c.

samdark:
OK. Makes sense.

Tobion:
@fabpot we need the getter. otherwise it hurts interoperability. it would not be possible to write a client based on PSR-7 when you actually want to send a custom reason phrase when there would not be a getter. just because there is an optional setter and a getter, it doesn't mean a user has to use it.

Tobion:
PSR-7 has to cover the minimum set to represent http messages. and the reason phrase belongs to that in http 1.1. otherwise people rely on specific implementations that provide the reason phrase again which would contradicts the whole point of PSR-7 .

fabpot:
My reasoning is a bit different: the reason phrase is never useful and should not be used (even in HTTP 1.1). Why do we need it in PSR-7 where we are trying to find a common denominator for interoperability. I don't see how the reason phrase helps here. Why clutter the interface here with something not useful and totally optional?

Tobion:
Because it's part of the HTTP standard. Otherwise we will never find a "common denominator". You could also say `getRequestTarget` is not used by most people.

fabpot:
That's totally different, getRequestTarget might be useful, the reason phrase is never useful. Anyway, I won't argue forever.

Tobion:
Let's see what the other say

mnpenner:
@fabpot That's a pretty bold statement. You can't tell me no one ever uses the reason phrase. I think it's useful for putting a short error message in there when something goes wrong. I'm with @Tobion on this. Someone even tried to localize it](http://stackoverflow.com/q/5217932/65387).

fabpot:
@mnpenner It's not about me, it's just about the HTTP RFCs. Anyway, feel free to close it.

mnpenner:
@fabpot PSR-7 specifically targets HTTP 1.0 and 1.1 though; unless there are plans to support 2.0, I don't think it's relevant that it's removed in 2.0. What do you mean by "optional" in 1.1? Are there any browsers that don't implement it/don't make the reason phrase accessible in some way? I agree that you shouldn't rely on it for anything important, but it can still be useful.

fabpot:
Changing the default value of the reason phrases is optional and doing so is useless.

aequasi:
@mnpenner why would a PSR ignore future releases of the HTTP spec? that seems kinda silly. Gotta agree with @fabpot here.... If your project requires the reason phrases, add it in your own. No point in solving a problem for the 10%.

mnpenner:
>  why would a PSR ignore future releases of the HTTP spec? that seems kinda silly.

Indeed. That's why I thought it was strange that MessageInterface said the protocol version must be "1.0" or "1.1" only. But reading it again, it seems I may have misread. Those were only examples.

The problem with "adding it on your own" is that 3rd party libraries *can't* add it, because it's not in the interface. For example, a CSRF middleware might want to set a special error message when there's a token mismatch, or an auth library might want to give a specific reason you're not authorized, etc. But I digress. It really isn't that big of a deal; there are better places to put such messages. Remove it if you wish.

fabpot:
Using the reason phrase for messages is very wrong, and that's also why I find it a very bad idea to let developers customize it. We're trying to find the common denominator and at the same time we clutter the API with things that in no way should matter. Too bad.

But @weierophinney disagrees.

Closing this now.

weierophinney:
@fabpot The point is not whether or not it's a bad idea; it's whether or not it could cause existing client or server use cases to fail. In particular, some client use cases are reliant on the reason phrase, as are some consumers of server-side-driven code. Omitting the reason phrase in those situations means that users interacting with such systems would not be able to use PSR-7. If the point of FIG is to promote interoperability, and to (largely) codify what member projects are doing, why would we omit this?

We've already made the reason phrase optional, and tied it to the status code. When emitting the response from a server-side application, you can completely ignore the reason phrase if desired; the same is true for those consuming a response returned by a client. However, if you have a need for the reason phrase, it's available as part of the specification.

pmjones:
:+1: @weierophinney

philsturgeon:
It's tough to balance current user cases whilst future proofing, but PSR-7 could easily be replaced with a new PSR later on like PSR-4 did with PSR-0. 

I'd rather we don't try to make these last 10 years, but make them work for 3 or 4 and replace them with a better one. For example, PSR-2 is already showing its age and definitely needs a replacement. 

Changes to HTTP specifications and changes to the PHP-land will likely affect PSR-7 enough to warrant replacement someday, even if it is mostly pretty similar.

PR #530, opened 2015-04-24
allow multiple attributes assignment

pine3ree:
In the case of route parameters assignment to the request attributes array, I believe it would be preferable not to need calling withAttribute() multiple times i.e. cloning the request for every single parameter.

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -1039,6 +1039,23 @@ interface ServerRequestInterface extends RequestInterface
     public function getAttribute($name, $default = null);
 
     /**
+     * Return an instance with the specified derived request attributes.
+     *
+     * This method allows setting multiple derived request attributes at once as
+     * described in getAttributes().
+     *
+     * This method MUST be implemented in such a way as to retain the
+     * immutability of the message, and MUST return an instance that has the
+     * updated attributes.
+     *
+     * @see getAttributes()
+     * @param array $attributes Array of key/value pairs representing the new or 
+     *      updated $attributes.
+     * @return self
+     */
+    public function withAttributes(array $attributes);
+
+    /**
      * Return an instance with the specified derived request attribute.
      *
      * This method allows setting a single derived request attribute as


weierophinney:
@pine3ree — we specifically chose *not* to do this. The reason is that the expected behavior varies from one developer to another:

- Should doing so *replace all* attributes? (i.e., calling it would delete existing attributes and replace them with the new set provided)
- Should doing so *merge* with existing attributes? (i.e, the final set would contain all previous attributes merged with the new ones)
  - If so, what should happen when a particular attribute already exists? should it overwrite, or create an array with all values?

As you can see, it gets quite complex, and because expectations differ, the end result is either disagreement over the proposed path, or proposals for *more* methods, so that *all* behaviors can be met.

Our decision, then, was to only allow manipulating one at a time, with `withAttribute()` *always* overwriting if the attribute exists. This makes the behavior predictable and much simpler to specify.

@simensen and/or @pmjones — please close.

pine3ree:
Hello Matthwe,

all those cases came to my mind too (and not just here...). But the similar questions can be made about a single assignment:

ex, using withAttribute()

-  Should it overwrite an existing attribute, 
or 
- Should it create an array with all values?

my thought is that when you call a with method you want to SET something for the new instance. So i believe a withAttributes method should do either one of these:

1. replace the entire attributes array (same as withQueryParameters, etc, etc)
2. replace existing and add new.

i believe that attributes should add up (2).

As of now calling many times withAttribute() does what point 2 does.
so why not just standardize mass attribute assignment?

kindly

weierophinney:
@pine3ree The specification is very clear at this point: `withAttribute()` **overwrites**. This is a single behavior; if developers want to merge values, they do so manually before calling the method.

Mass assignment is quite different, and the moment we add it, we re-open all the same debates we've had before: "I think it should overwrite!" "**I** think it should merge!" "Let's add a flag to specify the behavior!" "No, boolean flags are evil! Add separate methods!"

As such, my recommendation to close this issue unmerged stands.

pine3ree:
@weierophinney 
Thanks for the clarification about the decision.

PR #533, opened 2015-04-26
Add PSR-10, add sponsors to all entries

Crell:
Add the newly-entranced PSR-10.

Add the Sponsor to all entries, which was previously missing.

### DIFF
--- a/index.md
+++ b/index.md
@@ -8,43 +8,46 @@ Unless a PSR is marked as "Accepted" it is subject to change. Draft can change d
 
 ### Draft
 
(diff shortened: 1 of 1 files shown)


simensen:
Why N/A for missing editor and missing coordinator but empty cell for sponsor? I can only assume this was deliberate so curious what the reasons were behind it.

image

Crell:
The N/As for Coordinator were already there, and I didn't think to add them for the Sponsors. :smile: Added now.

PR #534, opened 2015-04-26
Add Beau Simensen as the new PSR-6 sponsor.

Crell:
Since John Mertic is no longer a voting representative.


simensen:
Will hold off on merging this until we hear back from @padraic.

PR #537, opened 2015-04-27
getHeaderLine() MUST return a string

weierophinney:
Slight change: if the header does not exist, instead of returning null, return an empty string. This is more consistent with `getHeader()` (which returns an empty array in the same situation), and with the `UriInterface` (where absence of a value returns an empty string).

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -562,12 +562,12 @@ interface MessageInterface
      * and supply your own delimiter when concatenating.
      *
      * If the header does not appear in the message, this method MUST return
-     * a null value.
+     * an empty string.
      *
      * @param string $name Case-insensitive header field name.
-     * @return string|null A string of values as provided for the given header
+     * @return string A string of values as provided for the given header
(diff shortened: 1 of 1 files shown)


weierophinney:
@Crell — please review!

Crell:
+1!

weierophinney:
@simensen and/or @pmjones — ready to merge!

Tobion:
Are you sure there is no semantical difference between an absense of a header and the presence of a header with an empty value? I can image there are, e.g. a custom `Debug` header that is only checked if it is present like a boolean header similar to html attributes. Now there would be no way anymore to distinguish that.

Tobion:
The comparison with UriInterface is not valid because in URI terms there is no difference between `null` and `''`, both mean it's not there. But for headers it's different.

Tobion:
I figured there is also a `hasHeader` method which can be used for such "boolean" headers. So this change here is ok and does not prevent this use-case.

PR #538, opened 2015-04-27
One-line short descriptions in method docblocks

weierophinney:
Per @Crell in the ML, this patch modifies the short-description lines of a number of methods that were wrapping to a second line, shortening them to fit on a single line.

Additionally, a small number of signature changes were made for consistency:

- `withStatus()` sets the default value of `$reasonPhrase` to a string.
- `getReasonPhrase()` now only returns a string; if no value is present, an empty string is returned.

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -550,8 +550,7 @@ interface MessageInterface
     public function getHeader($name);
 
     /**
-     * Retrieves the line for a single header, with the header values as a
-     * comma-separated string.
+     * Retrieves a comma-concatenated string of the values for a single header.
      *
      * This method returns all of the header values of the given
      * case-insensitive header name as a string concatenated together using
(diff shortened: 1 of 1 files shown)


weierophinney:
Note: the changes to the reason phrase do not pose a change in behavior; existing code should continue to work. Essentially, they force a consistent return value, which is in line with other changes we've made recently.

weierophinney:
@Crell — please review!

proposed/http-message.md
```diff
@@ -550,8 +550,7 @@ interface MessageInterface
     public function getHeader($name);
 
     /**
-     * Retrieves the line for a single header, with the header values as a
-     * comma-separated string.
+     * Retrieves a comma-concatenated string of the values for a single header.
```
Crell: I've never heard the phrase "comma-concatenated" before. I know what you mean, but there may be a slightly clearer way of saying it?

Crell:
Looks great aside from that nit.

proposed/http-message.md
weierophinney: I'll revert to "comma-separated".

weierophinney:
@Crell — updated

Crell:
+1!

weierophinney:
@simensen and/or @pmjones — Ready to merge!

PR #539, opened 2015-04-27
s/move/moveTo/

weierophinney:
Per @Crell, this change clarifies the purpose of the method by renaming it to `moveTo()`. Additionally, the method argument name was changed to `$targetPath` to make the change completely clear.

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -416,7 +416,7 @@ This proposal also recognizes that implementations may operate in non-SAPI
 environments. As such, `UploadedFileInterface` provides methods for ensuring
 operations will work regardless of environment. In particular:
 
-- `move($path)` is provided as a safe and recommended alternative to calling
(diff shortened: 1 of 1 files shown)


Crell:
$targetPath is even better, yes!

One other question: What is that path relative to?  I would implicitly assume it's script-root-relative, as is PHP's default, but that's not clear.  (And script-root-relative is always messy anyway.)

weierophinney:
@Crell —

> What is that path relative to?

That would be dependent on the `$targetPath` value and the current working directory. If the `$targetPath` is absolute, you write to that; otherwise, you write relative to the current working directory.

Crell:
That should be made explicit, I think. Even if it's explicit to "the same logic as what PHP's move() function does" (which I think is the point).

Crell:
+1!

weierophinney:
@simensen and/or @pmjones — Ready to merge! I'll create updates later on http-message once all issues are addressed.

PR #540, opened 2015-04-27
More detail on normalization of the uploaded files

weierophinney:
Per @Crell, taking a stab at clarifying what the "tree of files" looks like, and
what normalization needs to be done.

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -394,9 +394,61 @@ Implementations are expected to:
   the tree.
 
 The tree structure referenced should mimic the naming structure in which files
-were submitted. As an example, if a file form element was provided using the name
-`my-form[details][avatar]`, the corresponding tree returned by
-`getUploadedFiles()` should be:
+were submitted. 
+
+In the simplest example, this might be a single named form element submitted as:
+
+```html
+<input type="file" name="avatar" />
+```
+
+In this case, the structure in `$_FILES` would look like:
+
+```php
+array(
+    'avatar' => array(
+        'tmp_name' => 'phpUxcOty',
+        'name' => 'my-avatar.png',
+        'size' => 90996,
+        'type' => 'image/png',
+        'error' => 0,
+    ),
+)
+```
+
+The normalized form returned by `getUploadedFiles()` would be:
+
+```php
+array(
+    'avatar' => /* UploadedFileInterface instance */
+)
+```
+
+In the case of an input using array notation for the name:
+
+```html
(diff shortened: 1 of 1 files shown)


proposed/http-message.md
```diff
@@ -393,11 +393,89 @@ Implementations are expected to:
...
+                    1 => '...',
+                    2 => '...',
+                ),
+            ),
+        ),
+    ),
+)
+```
+
+The implementation would convert the above to:
```
Crell: "The implementation" is rather vague.  You mean a PSR-7 implementation, correct?

weierophinney: Updated.

Crell:
What's missing from the description here is the common simple case. :smile: That is, if all I have is a single file field named "resume", it should then look like...?

weierophinney:
@Crell — added two more use cases:

- `<input type="file" name="avatar" />` — simplest use case.
- `<input type="file" name="my-form[details][avatar]" />` — existing use case already present, but now more explicit.

I've also clarified the final example to be consistent with the above.

proposed/http-message.md
```diff
@@ -393,11 +393,150 @@ Implementations are expected to:
...
+                    1 => '...',
+                    2 => '...',
+                ),
+            ),
+        ),
+    ),
+)
+```
+
+An implementation of this specification would convert the above to:
```
simensen: Should this say "could" instead of "would?" It makes it seem like implementations MUST do it this way. While that is sensible, perhaps, it might tend to tip too far in the direction of dictating implementation details.

weierophinney: They MUST do it that way; that's what we're specifying. The whole idea behind the `getUploadedFiles()` implementation is that any given leaf in the tree is an `UploadedFileInterface` instance, and that leaves can be retrieved using the same structure in which they were defined when uploaded.

Crell: Perhaps "the above $_FILES array would correspond to the following structure as returned by getUploadedFiles()"?  How it gets there is irrelevant, but there is only One True Way(tm) to represent that data (by design).

Crell:
Other than the thread from @simensen above, this looks good.  Huge improvement.  Thanks, @weierophinney!

weierophinney:
@Crell Updated with your verbiage, which makes it more consistent with the other examples anyways!

Thanks!

@simensen I think we're ready to merge!

weierophinney:
Pinging @simensen and/or @pmjones ; I have sign-off from @Crell . :shipit:

PR #541, opened 2015-04-29
fix php code syntax

iannsp:
sorry for move a ";" and propose a file change just for it, but I'm reading the document and found it.


weierophinney:
@simensen / @pmjones  - good to merge. :+1:

PR #543, opened 2015-04-29
Clarification of withPath

weierophinney:
Clarify domain-qualified vs relative paths; verbiage as suggested by @Crell.

### DIFF
--- a/proposed/http-message.md
+++ b/proposed/http-message.md
@@ -1533,6 +1533,11 @@ interface UriInterface
      * rootless (not starting with a slash). Implementations MUST support all
      * three syntaxes.
      *
+     * If the path is intended to be domain-relative rather than path relative then
+     * it must begin with a slash ("/"). Paths not starting with a slash ("/")
+     * are assumed to be relative to some base path known to the application or
+     * consumer.
+     *
      * Users can provide both encoded and decoded path characters.
      * Implementations ensure the correct encoding as outlined in getPath().
      *


weierophinney:
@simensen and/or @pmjones — let's go ahead and merge. The verbiage is essentially verbatim to the suggestion @Crell made on-list, and clarifies the intent (i.e., to allow empty paths, those with just a slash character, absolute paths, and relative paths).

proposed/http-message.md
```diff
@@ -1533,6 +1533,11 @@ interface UriInterface
      * rootless (not starting with a slash). Implementations MUST support all
      * three syntaxes.
      *
+     * If the path is intended to be domain-relative rather than path relative then
```
Tobion: I wouldn't change the terminology here again. We used "host" instead of domain everywhere else. Also the domain doesn't include the port.

proposed/http-message.md
```diff
@@ -1533,6 +1533,11 @@ interface UriInterface
...
+     * it must begin with a slash ("/"). Paths not starting with a slash ("/")
+     * are assumed to be relative to some base path known to the application or
```
Tobion: or the URI path is really not starting with a slash which is also a valid URI.

proposed/http-message.md
Tobion: and why is it "domain-relative" but "path relative" without dash?

Tobion:
Fixed remarks in #555

PR #544, opened 2015-04-29
Changed "case-insensitive" to "case-sensitive" in RequestInterface::with...

pjdietz:
Looks a like a typo since the longer part of the description makes a point to describe methods as case-sensitive.


weierophinney:
Good catch, @pjdietz 

@simensen and/or @pmjones — :+1:; please merge.

PR #547, opened 2015-05-04
[PSR7] Replace array with more narrow phpdoc types

Maks3w:


proposed/http-message.md
```diff
@@ -1038,7 +1038,7 @@ interface ServerRequestInterface extends RequestInterface
      * immutability of the message, and MUST return an instance that has the
      * updated query string arguments.
      *
-     * @param array $query Array of query string arguments, typically from
+     * @param scalar[] $query Array of query string arguments, typically from
```
weierophinney: This isn't necessarily true; `$_GET` allows array values for arguments.

proposed/http-message.md
```diff
@@ -1001,7 +1001,7 @@ interface ServerRequestInterface extends RequestInterface
      * immutability of the message, and MUST return an instance that has the
      * updated cookie values.
      *
-     * @param array $cookies Array of key/value pairs representing cookies.
+     * @param scalar[] $cookies Array of key/value pairs representing cookies.
```
weierophinney: Based on the behavior of `http_parse_cookie()`, it's possible for individual cookie values to potentially be an array, IIRC. This should likely be `mixed[]`.

proposed/http-message.md
```diff
@@ -1053,8 +1053,8 @@ interface ServerRequestInterface extends RequestInterface
      * These values MAY be prepared from $_FILES or the message body during
      * instantiation, or MAY be injected via withUploadedFiles().
      *
-     * @return array An array tree of UploadedFileInterface instances; an empty
-     *     array MUST be returned if no data is present.
+     * @return UploadedFileInterface[] An array tree of UploadedFileInterface
```
weierophinney: I actually specifically did *not* use this notation previously, as the structure is potentially a tree. While every leaf is an `UploadedFileInterface`, not every _value_ of the top-level array will necessarily be one.

proposed/http-message.md
```diff
@@ -1065,7 +1065,8 @@ interface ServerRequestInterface extends RequestInterface
      * immutability of the message, and MUST return an instance that has the
      * updated body parameters.
      *
-     * @param array An array tree of UploadedFileInterface instances.
+     * @param UploadedFileInterface[] An array tree of UploadedFileInterface
```
weierophinney: See previous comment.

weierophinney:
@pmjones and/or @simensen — :+1:, ready to merge.

PR #554, opened 2015-05-13
[PSR-7] fix leftover null reason phrase

Tobion:
Follow up to 5506a9c2a453969aaabb4e359c41d2bef938bc0b

ping @weierophinney

PR #556, opened 2015-05-13
[PSR-7] Replace null with empty string

JoeBengalen:
The return value MUST be an empty string, but the description said it could be null.
Also 'must' wasn't uppercase.

I know psr-7 is in Voting stage, but I guess this is just a forgotten textual change.


Maks3w:
Duplicate #554

JoeBengalen:
Oh well, kinda missed that I guess.

PR #557, opened 2015-05-19
PSR-7.

simensen:
For when this is accepted tonight.


weierophinney:
:+1: - all changes look good to me, and the naming after moving of the files is correct per the other accepted proposals.

PR #558, opened 2015-05-19
[PSR-7] Fixed quotes

hason:


ockcyp:
:+1:

samdark:
:+1:

Tobion:
:+1:

simensen:
I need to make sure that we can fix this type of thing in accepted PSR's. One would think we should be able to do so but I don't recall exactly what we are allowed to do. I know there are a few of these that have stacked up so hopefully we can get around to it in the next week or so.

pmjones:
Changes of no substance, such as typos that do not change meaning, are fair game.

simensen:
@pmjones so you'd feel comfortable merging this type of request? if so, i'd say let's merge it.

PR #566, opened 2015-06-01
[PSR-7] Fix title heading level

gaomd:


michaelcullum:
@weierophinney Is this valid?

I believe this type of change is allowed even after acceptance vote as it's a typo/formatting change.

PR #577, opened 2015-06-15
duplicate rows

godem111:
duplicate rows


bocharsky-bw:
Yes, it's look like a typo :+1:

PR #592, opened 2015-07-29
Reverse order of PSR Index by Status

colinodell:
I'd like to propose reversing the order of the Index by Status section.

The PHP-FIG works hard to create and approve these PSRs, so I feel the Accepted ones should be presented front-and-center, not buried half-way down the page.  I also think most visitors interested in implementing these PSRs would expect them to be displayed prominently, instead of seeing the incomplete works first.  This change is analogous to PSRs becoming "promoted" to a "higher" status.

I doubt that this change would negatively-impact those who are (or wish to become) more-involved in FIG.  These visitors will likely be familiar with the process and statuses, and will therefore know to scroll past the Accepted PSRs to locate those being worked on.

TL;DR: I think this change makes the index slightly more community-friendly.

### DIFF
--- a/index.md
+++ b/index.md
@@ -6,26 +6,6 @@ Unless a PSR is marked as "Accepted" it is subject to change. Draft can change d
 
 ## Index by Status
 
-### Draft
-
-| Num | Title                          | Editor                  |  Coordinator   | Sponsor        |
-|:---:|--------------------------------|-------------------------|----------------|----------------|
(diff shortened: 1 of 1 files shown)


samdark:
Agree. Please announce it at mailing list: https://groups.google.com/forum/#!forum/php-fig

reinink:
Yes, great! This will also work better on the new website, which uses this markdown document.

Crell:
:+1: 

But let's wait for the discussion on list before merging.  (I just linked to this PR from the website thread.)

colinodell:
Thanks @Crell!

PR #595, opened 2015-08-04
[PSR-6] Changes for consistency

hannesvdvreken:
- Moved around some spaces in docblocks to match other docblock's style
- All occurrences of `boolean` or `bool` are now `bool`.
- Replaced 'if one is not specified' with 'if none is specified', afaik this is equal.
- Replaced 'original' with 'originally' as (I think) this is the right variation of the word to use in this sentence.
- Made all 'True if `...`. False otherwise.' sentences equal.
- Empty classes have braces at the right place.

### DIFF
--- a/proposed/cache.md
+++ b/proposed/cache.md
@@ -175,7 +175,7 @@ interface CacheItemInterface
     /**
      * Retrieves the value of the item from the cache associated with this object's key.
      *
-     * The value returned must be identical to the value original stored by set().
+     * The value returned must be identical to the value originally stored by set().
      *
      * if isHit() returns false, this method MUST return null. Note that null
      * is a legitimate cached value, so the isHit() method SHOULD be used to
@@ -193,13 +193,14 @@ interface CacheItemInterface
      * although the method of serialization is left up to the Implementing
      * Library.
      *
-     * Implementing Libraries MAY provide a default TTL if one is not specified.
+     * Implementing Libraries MAY provide a default TTL if none is specified.
      * If no TTL is specified and no default TTL has been set, the TTL MUST
      * be set to the maximum possible duration of the underlying storage
      * mechanism, or permanent if possible.
      *
      * @param mixed $value
      *   The serializable value to be stored.
+     *
      * @return static
      *   The invoked object.
      */
@@ -211,7 +212,7 @@ interface CacheItemInterface
      * Note: This method MUST NOT have a race condition between calling isHit()
      * and calling get().
      *
-     * @return boolean
+     * @return bool
      *   True if the request resulted in a cache hit. False otherwise.
      */
     public function isHit();
@@ -223,8 +224,8 @@ interface CacheItemInterface
      * reasons, which could result in a race condition between exists() and get().
      * To avoid that potential race condition use isHit() instead.
      *
-     * @return boolean
-     *  True if item exists in the cache, false otherwise.
+     * @return bool
+     *  True if item exists in the cache. False otherwise.
      */
     public function exists();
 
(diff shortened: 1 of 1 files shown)


dragoonis:
@hannesvdvreken we don't put the FQCN in the phpdoc output because the full class is in the 'use' statement at the top. Please revert your commit to remove the entries of FQCN definitions.

hannesvdvreken:
Alright, I will. I'll notify you when it's done.

hannesvdvreken:
@dragoonis Updated it. No need for `use` statements though.

tedivm:
:+1:

dragoonis:
@simensen @Crell - what's your thoughts on this?

proposed/cache.md
```diff
@@ -293,11 +293,13 @@ interface CacheItemPoolInterface
...
-     * @return \Psr\Cache\CacheItemInterface
-     *   The corresponding Cache Item.
-     * @throws \Psr\Cache\InvalidArgumentException
+     *
+     * @throws InvalidArgumentException
      *   If the $key string is not a legal value a \Psr\Cache\InvalidArgumentException
      *   MUST be thrown.
+     *
+     * @return CacheItemInterface
+     *   The corresponding Cache Item.
```
Crell: Why rearrange these?  I'm not aware of any general standard on the order here, and more often than not what I see in practice is @throws at the very end.

Does PSR-5 have a statement on the matter?

proposed/cache.md
```diff
@@ -375,7 +379,9 @@ namespace Psr\Cache;
  * Any time an invalid argument is passed into a method it must throw an
  * exception class which implements Psr\Cache\InvalidArgumentException.
  */
-interface InvalidArgumentException { }
+interface InvalidArgumentException
+{
+}
```
Crell: Is this a PSR-2 thing?  Really?  I hate PSR-2... :frowning:

proposed/cache.md
GrahamCampbell: PSR-5 examples always include throws NOT at the end.

I think this makes sense because you throw before you returm.

GrahamCampbell: PSR-5 examples are ordered: param, throws, return.

proposed/cache.md
GrahamCampbell: Yeh, this is required by psr2.

proposed/cache.md
GrahamCampbell: The PSR-5 spec doesn;t say the order matters though. The spec defines what the annotations mean, not the code style.

proposed/cache.md
```diff
@@ -193,13 +193,14 @@ interface CacheItemInterface
      * although the method of serialization is left up to the Implementing
      * Library.
      *
-     * Implementing Libraries MAY provide a default TTL if one is not specified.
+     * Implementing Libraries MAY provide a default TTL if none is specified.
```
Crell: I've no idea what language is more common.  Which version did PSR-7 use? Let's be consistent with that, whatever it is.

proposed/cache.md
Crell: What do PSR-3 and PSR-7 do?  I think consistency matters most here.

Crell:
These are all in-scope-for-Review changes IMO. Discussion above.

proposed/cache.md
tedivm: If PSR-5 is likely to define it this way then lets just put it this way. This is trivial from a standards perspective while still being consistent. Plus I have to agree with Graham, I always see return as the last position.

proposed/cache.md
Crell: I just checked PSR-7, and "if none" is used consistently there.  So +1 to this adjustment for consistency.

Crell:
I was the only blocking dissent here for a month and a half, so let's merge these now that my hangups are satisfied.

hannesvdvreken:
Thanks for merging.

PR #596, opened 2015-08-04
[PSR-6] Fix exceptions so that they are throwable

hannesvdvreken:
If not, if you'd throw an object that implements these, it will generate a PHP fatal error.

### DIFF
--- a/proposed/cache.md
+++ b/proposed/cache.md
@@ -372,10 +372,12 @@ namespace Psr\Cache;
 /**
  * Exception interface for invalid cache arguments.
  *
- * Any time an invalid argument is passed into a method it must throw an
- * exception class which implements Psr\Cache\InvalidArgumentException.
+ * Any time an invalid argument is passed into a method it must throw a
+ * \Psr\Cache\InvalidArgumentException.
  */
-interface InvalidArgumentException { }
+class InvalidArgumentException extends \InvalidArgumentException
+{
+}
 ```
 
 ### CacheException
@@ -390,7 +392,9 @@ Any exception thrown by an Implementing Library MUST implement this interface.
 namespace Psr\Cache;
 
 /**
- * Exception interface for all exceptions thrown by an Implementing Library.
+ * Base exception for all exceptions thrown by an Implementing Library.
  */
-interface CacheException {}
+class CacheException extends \Exception
+{
+}
 ```


dragoonis:
@Crell @simensen - legitimate minor fixes, so happy to merge this. Do communicate if you see anything wrong with these minor corrections.

simensen:
@dragoonis Looks good to me, thanks!

dragoonis:
:+1:

tedivm:
I see a problem. By merging this the standard is now impossible to implement in it's current form, as we require *all* exceptions thrown by caching libraries to implement the CacheException interface-

> Any exception thrown by an Implementing Library MUST implement this interface.

By turning both of them into classes instead of interfaces you make this literally impossible, as you can't inherit from multiple classes. So this change is breaking the standard.

In addition, we don't implement *any* classes in this standard. Adding two exceptions in should not be the exception (lol). Unfortunately PHP fails by not providing an exception interface, so I'm not sure what the cleanest way to do this is. I would stick with leaving them as interfaces and requiring people to handle extending the /Exception class as well as requiring these exceptions.

Crell:
As @tedivm noted, this PR is completely and totally broken.  The whole point of providing *interfaces* for extensions is to make them flexible enough to allow implementers to implement it on their own classes.

Also, changes *need* to be discussed on list.  GitHub notifications are a total fail, so whether or not someone knows this thread is even here is debatable. I didn't know it existed until I happened to see the commit happen in IRC, by pure luck.  Asking for feedback in a mere 2 hours is absolutely insufficient.  I routinely have meetings longer than that and most people don't camp out waiting for GitHub notifications (that don't come).

This needs to be rolled back (the whole commit purged, not just reverted) immediately.

tedivm:
Purging versus reverting is irrelevant, and purging can break other people's repositories. We don't need to wipe it from history and salt the earth behind us, just fix the standard so it works.

simensen:
My bad for the :+1: in this. I'll revert it shortly!

hannesvdvreken:
@crell My appologies. It was not my intention to break things. I just wanted to spawn a discussion with a clear example of what could change. I didn't ask for it to be merged without discussing if first.

The use of concrete exception classes over interfaces is a breaking change, granted. I doubted why this couldn't be concrete classes so I went looking for a reference. As an example I looked at the InvalidArgumentException that comes with psr-3 https://github.com/php-fig/log/blob/master/Psr/Log/InvalidArgumentException.php It's a concrete class as well.

Also: mind the strong language "asking for feedback in a mere 2 hours is absolutely insufficient.", and "this PR is completely and totally broken."

tedivm:
@hannesvdvreken it's not your fault, things happen. You are right that this is inconsistent, and it does make me wonder if the FIG group should perhaps come up with a standard around exceptions (at least an internal one that can be applied to future PSRs).

On the plus side going with interfaces doesn't disallow us from shipping exception classes at some point, whereas going in the other direction does. We can add new classes in the future that implement these interfaces, and then encourage people to test against the interfaces. That gives us the best of all worlds.

Crell:
> it does make me wonder if the FIG group should perhaps come up with a standard around exceptions (at least an internal one that can be applied to future PSRs).

:+1:

PR #604, opened 2015-08-21
Remove getExpiration() from cache items

Crell:
As discussed on the mailing list.

### DIFF
--- a/proposed/cache.md
+++ b/proposed/cache.md
@@ -49,7 +49,7 @@ by an integer representing time in seconds, or a DateInterval object.
 typically calculated by adding the TTL to the time when an object is stored, but
 may also be explicitly set with DateTime object.
 
-    An item with a 300 second TTL stored at 1:30:00 will have an expiration at
+    An item with a 300 second TTL stored at 1:30:00 will have an expiration of
     1:35:00.
 
     Implementing Libraries MAY expire an item before its requested Expiration Time,
--- a/proposed/cache.md
+++ b/proposed/cache.md
@@ -255,16 +255,6 @@ interface CacheItemInterface
      */
     public function expiresAfter($time);
 
-    /**
-     * Returns the expiration time of a not-yet-expired cache item.
-     *
-     * If this cache item is a Cache Miss, this method MAY return the time at
-     * which the item expired or the current time if that is not available.
-     *
-     * @return \DateTime
-     *   The timestamp at which this cache item will expire.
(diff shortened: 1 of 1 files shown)


proposed/cache.md
```diff
@@ -255,16 +255,6 @@ interface CacheItemInterface
      */
     public function expiresAfter($time);
```
simensen: I don't recall if extra blank lines in interfaces is against PSR-2 or not, but I wanted to mention it in case someone else thinks it might. I also understand this is inline code so not sure how picky we need to be following PSR-2 for this.

proposed/cache.md
```diff
@@ -49,7 +49,7 @@ by an integer representing time in seconds, or a DateInterval object.
 typically calculated by adding the TTL to the time when an object is stored, but
 may also be explicitly set with DateTime object.
 
-    An item with a 300 second TTL stored at 1:30:00 will have an expiration at
+    An item with a 300 second TTL stored at 1:30:00 will have an expiration of
```
simensen: What about:

> An item with a 300 second TTL stored at 1:30:00 will expire at 1:35:00.

Probably doesn't matter too much and if this change hadn't been in this PR it probably wouldn't have even come up so I'm cool w/ leaving it with the new way suggested in this PR as well. :)

simensen:
:+1: I'd like to get feedback from @dragoonis & @tedivm before accepting this. By accepting this, I mean dropping PSR-6 from [REVIEW] long enough to merge these and put it back up for [REVIEW] again.

I probably should not have mentioned the thing up there w/ the alternate wording. Totally unimportant. Sorry @Crell. :) Don't bother responding to it.

proposed/cache.md
Crell: Um. I have no idea. :smile: I can remove whitespace if we really want.

simensen:
@dragoonis @tedivm Can I get a yay or nay on this one?

tedivm:
Yay.

I'm going to be off the grid for the next two weeks so just refer to Paul's judgement on this.

Rob

dragoonis:
@tedivm - I am happy with such change. "less is more" :-)

dragoonis:
Sorry I meant to tag @simensen* also.*

Crell:
Everyone seems on board here, so to keep things moving I'm merging this.

dragoonis:
:+1:

PR #611, opened 2015-08-24
Extended Coding Style Guide PSR

michaelcullum:
Pre-draft (So just introduction to what the PSR is about, no content as of yet). Typos and language might be adjusted at a later date but this is the general gist.

Entrance vote: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/php-fig/P9atZLOcUBM/_jwkvlYKEAAJ

Editor: Michael Cullum @michaelcullum
Coordinator: Korvin Szanto @KorvinSzanto 
Sponsor: Alexander Makarov @samdark


proposed/extended-coding-style-guide-meta.md
```diff
@@ -0,0 +1,96 @@
+Extended Coding Style Guide Meta Document
+====================================
```
GrahamCampbell: are we missing some `=` here?

samdark: Yes but it doesn't really matter for final rendered HTML.

GrahamCampbell: yeh, sure ;)

proposed/extended-coding-style-guide-meta.md
```diff
@@ -0,0 +1,96 @@
...
+
+2. Why Bother?
+--------------
+
+PSR-2 was accepted in 2012 and since then a number of changes have been made to PHP,
+most notably recent changes for PHP 7, which have implications for coding style
+guidelines. Whilst PSR-2 is very comprehensive of PHP functionality that existed at
+the time of writing, new functionality is very open to interpretation. PSR-N seeks
+to provide a set way that both coding style tools can implement, projects can declare
+adhearance to and developers can easily relate on between different projects for these
```
mwhitneysdsu: adhearance -> adherence (I even checked to make sure this wasn't a difference due to UK/US English)

mwhitneysdsu: Also, "relate to" was more correct than "relate on", but maybe "rely on" would be better.

KorvinSzanto:
Thanks @mwhitneysdsu, fixed that one and another one.

proposed/extended-coding-style-guide-meta.md
michaelcullum: Yeah I wasn't sure about the relate bit. I think this whole clause could be reworded potentially.

mwhitneysdsu: Maybe it would be better to just crib from the PSR-2 meta: "The intent of this guide is to reduce cognitive friction when scanning code from different authors. It does so by enumerating a shared set of rules and expectations about how to format PHP code."

michaelcullum: We do say that, just elsewhere.

Issue #609, opened 2015-08-22
[PSR-6] Decoupling CacheItem from CacheItemPool


rindeal:
As already pointed out in #389, method `exists()` is the only method in `CacheItemInterface` which requires to know which `CacheItemPool` the item belongs to.

When trying to figure out, why `exists()` exists at all, I stumbled upon [this article](http://blog.ircmaxell.com/2014/10/a-followup-to-open-letter-to-php-fig.html), which explains that the reason is memory efficiency for someone who just needs to peek into the cache to see if the item exists without caring what it's value is. To avoid race conditions the caller will also likely not use any other method provided in `CacheItemInterface`, unless he calls `getItem()` again.

To sum it up, I see no reason why `exists()` shouldn't be moved to `CacheItemPoolInterface`.

tedivm:
This is an excellent point, but it should be brought up on the mailing list.

rindeal:
Thanks. [Here is the link to mailing list conversation](https://groups.google.com/forum/#!topic/php-fig/NKaQhF98TVI)

dragoonis:
Closing this in favour of #612

Issue #389, opened 2014-12-04
PSR-6: List of implementation problems


milesj:
I decided to update one of my codebases in preparation for PSR-6 by copying the current proposal and modifying my code accordingly. The code can be found here: https://github.com/titon/framework/tree/master/src/Titon/Cache (Do note that the project is in Hack and that the Storage class is the pool).

While implementing the code, I found some weird implementations that don't make much sense, so I thought I would point them all out in a giant ticket.

## Item::exists() should be removed ##

This method should return a true or false if the item exists. An item is either a hit (it exists), or a miss (it doesn't exist), so why do we have a method to check it again? To see if it expired in that short timeframe? I can't seem to find a use case.

Honestly, the pool should handle this as the pool is the abstract layer and supports multiple implementations (memcache, redis, filesystem, etc). The item has no concept of the pools implementation, so can not successfully check if the item exists as it doesn't know which engine it is communicating with (memcache, etc).

I could find no way to successfully implement this without passing the pool into the item, which I do not want to do. For now, I'm just throwing an exception if you call the method. https://github.com/titon/framework/blob/master/src/Titon/Cache/Item.hh#L66

## Item::get() should not return null for misses ##

According to the docblocks of this method "if isHit() returns false, this method MUST return null.". The problem with this statement is that you can no longer set a value on a missed item and re-save it. Take the following example from the meta docs:

```pph
function get_widget_list()
{
    $pool = get_cache_pool('widgets');
    $item = $pool->getItem('widget_list');
    if (!$item->isHit()) {
        $value = compute_expensive_widget_list();
        $item->set($value);
        $pool->save($item);
    }
    return $item->get();
}
```

The `$pool->save($item);` line would fail as `$item->get()` would always return `null` and would not represent the new `$value` (the return statement would also return `null` instead of `$value`). If our `Item` implementation changes `isHit()` to be true when we set a new value, the state of this item is no longer a true representation.

I believe the solution is to have `get()` return whatever the default item value is, usually null. This is acceptable since a missed item wouldn't have a value anyways.

## Item::getExpiration() returning the expires for missed items ## 

According to the docblocks of this method "If this cache item is a Cache Miss, this method MAY return the time at which the item expired or the current time if that is not available.". Missed item's don't have expiration times as they no longer exist, or am I wrong on this? I have yet to see/find a way to implement this, so any guidance would be helpful.

I believe this method should return null for misses/hits and should only return a value when something is going to be saved.

[...]

PR #612, opened 2015-08-25
[PSR-6] move exists() from Item to ItemPool

rindeal:
related issues: #609, #389 
related discussion: https://groups.google.com/forum/#!topic/php-fig/NKaQhF98TVI
ping: @Crell  @simensen 

Wording will likely need to change as well.

### DIFF
--- a/proposed/cache.md
+++ b/proposed/cache.md
@@ -67,12 +67,8 @@ supported by implementing libraries: `{}()/\@:`
 
 *    **Hit** - A cache hit occurs when a Calling Library requests an Item by key
 and a matching value is found for that key, and that value has not expired, and
-the value is not invalid for some other reason.
-
-*    **Exists** - When the item exists in the cache at the time of calling
-hasItem(). As this is separate from isHit() there's a potential race condition
-between the time hasItem() is called and get() being called so Calling Libraries
-SHOULD make sure to verify isHit() on all of the get() calls.
+the value is not invalid for some other reason. Calling Libraries SHOULD make
+sure to verify isHit() on all of the get() calls.
 
 *    **Miss** - A cache miss is the opposite of a cache hit. A cache miss occurs
 when a Calling Library requests an item by key and that value not found for that
--- a/proposed/cache.md
+++ b/proposed/cache.md
@@ -309,10 +309,12 @@ interface CacheItemPoolInterface
      * reasons, which could result in a race condition between exists() and get().
      * To avoid that potential race condition use isHit() instead.
      *
+     * @param string $key
+     *    The key for which to check existance.
      * @return boolean
      *  True if item exists in the cache, false otherwise.
      */
-    public function itemExists();
+    public function itemExists($key);
 
     /**
      * Deletes all items in the pool.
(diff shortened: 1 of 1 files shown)


proposed/cache.md
```diff
@@ -315,6 +303,18 @@ interface CacheItemPoolInterface
...
+     * Confirms if the cache item exists in the cache.
+     *
+     * Note: This method MAY avoid retrieving the cached value for performance
+     * reasons, which could result in a race condition between exists() and get().
+     * To avoid that potential race condition use isHit() instead.
+     *
+     * @return boolean
+     *  True if item exists in the cache, false otherwise.
+     */
+    public function itemExists();
```
joshdifabio: Don't we need to pass the key in here?

rindeal: Maybe... :D

proposed/cache.md
```diff
@@ -315,6 +303,20 @@ interface CacheItemPoolInterface
     public function getItems(array $keys = array());
 
     /**
+     * Confirms if the cache item exists in the cache.
+     *
+     * Note: This method MAY avoid retrieving the cached value for performance
+     * reasons, which could result in a race condition between exists() and get().
+     * To avoid that potential race condition use isHit() instead.
```
Crell: The docblock also needs to be updated to clarify that we're talking about get() on Item, not on Pool.

simensen:
@Crell @dragoonis @tedivm Are we considering this change for when we do the review/draft/review switcharoo? Do we need to discuss it more / get more feedback?

dragoonis:
@simensen yes, no further discussion needed, I believe. Rob and myself are happy with this. I don't recall see any push back from Larry either.

Merge away!

simensen:
@dragoonis Indeed, @Crell has actually contributed to this PR. That said, I'd still like a :+1: from @Crell but I expect that will probably be the case. Thanks for providing feedback on this!

Crell:
The latest version above removes the race condition mention entirely, which is not what I was asking for. :smile: I said it needs to be revised to make it clear what call was being referred to.

We also should check the rest of the spec and make sure nothing else needs to be updated as well, such as the mentions of exists in the prose section.

I'm not opposed in concept, just want to make sure it's fully polished.

Crell:
I went through the doc and didn't see anywhere else that needs updating as a result of this change.  It sounds like everyone is on board, so merging.

dragoonis:
@rindeal ping

Crell:
@dragoonis Can you rebase this branch to a new PR, preserving @rindeal's commit credits? I can merge it then.

dragoonis:
@Crell I'm slammed today, totally, as I'm about to go on vacation for my birthday. Would you be able to rebase this, on this occasion? I'd appreciate it, so we can get this pushed through. I think it's the last outstanding PR.

rindeal:
Sorry for the delay. I could do it if you tell me which commands to issue. I'm not so proficient in merging conflicts and wouldn't like to do more harm than good.

Crell:
Merged master, filed new PR: #641.  @dragoonis, please verify and merge ASAP.

PR #620, opened 2015-09-03
PSR Config

sandrokeil:
Please take a look at the [config-meta.md](https://github.com/sandrokeil/fig-standards/blob/config/proposed/config-meta.md) and [config.md](https://github.com/sandrokeil/fig-standards/blob/config/proposed/config.md) for details. This is a draft, so if you have any suggestions, please leave a comment.

### DIFF
--- a/proposed/config-meta.md
+++ b/proposed/config-meta.md
@@ -22,18 +22,17 @@ It can also be used to build the name of the *Container* entry.
 * It's not a goal to validate the configuration option values.
 
 ## 4. History
-Before submitting the Config PSR to the PHP-FIG, the interfaces was first proposed in a project named 
+Before submitting the Config PSR to the PHP-FIG, the interfaces were first proposed in a project named 
 [interop-config](https://github.com/sandrokeil/interop-config).
 
-The goal of the project was to provide a test-bed for implementing the interfaces and implementation and to pave the way 
+The goal of the project was to provide a test-bed for implementing the interfaces and to pave the way 
 for the Config PSR.
 
 ## 5. Interfaces
-There are six interfaces and some interfaces extends from another interface to ensure the uniform configuration structure. 
-If you confused about the naming of the interfaces or methods, note that I have used the Domain Driven Design approach.
+This specification defines four interfaces. The naming of the interfaces and methods follows the Domain-Driven Design approach.
 
 ### 5.1 RequiresConfig
-This is the main interface and describes the default configuration structure. Since we have Composer more and more 
+This is the main interface and describes the default configuration structure. Since we have Composer, more and more 
 libraries can be combined to create individual projects. So it is only logical to start the configuration structure 
 with a `vendorName`. A vendor has multiple packages so the next structure depth is `packageName`.
 
--- a/proposed/config-meta.md
+++ b/proposed/config-meta.md
@@ -8,7 +8,7 @@ The word `Container` in this document is to be interpreted as the `ContainerInte
 
 ## 2. Why bother?
 The configuration of instances via factories is not uniform, uses different ways to retrieve and check the needed 
-options and it's not clear which configuration options needs the factory to create the instance.
+options and it's not clear which configuration options the factory needs to create the instance.
 
 ## 3. Scope
 
--- a/proposed/config-meta.md
+++ b/proposed/config-meta.md
@@ -13,9 +13,10 @@ options and it's not clear which configuration options the factory needs to crea
 ## 3. Scope
 
 ### 3.1 Goals
-The goal set by the Config PSR is to standardize how factories uses a configuration to create instances, support for 
+* The goal set by the Config PSR is to standardize how factories uses a configuration to create instances, support for 
 auto discovery of needed configuration, to reduce boilerplate code and to make it more readable and easier to understand. 
-It can also be used to build the name of the *Container* entry.
+* Validates the configuration structure depending on the implemented interfaces.
+* It can also be used to build the name of the *Container* entry.
 
 ### 3.2 Non-goals
 * It's not a goal to define a whole project/library configuration structure.
--- a/proposed/config-meta.md
+++ b/proposed/config-meta.md
@@ -32,7 +32,7 @@ for the Config PSR.
 There are six interfaces and some interfaces extends from another interface to ensure the uniform configuration structure. 
 If you confused about the naming of the interfaces or methods, note that I have used the Domain Driven Design approach.
 
-### 5.1 HasConfig
+### 5.1 RequiresConfig
 This is the main interface and describes the default configuration structure. Since we have Composer more and more 
 libraries can be combined to create individual projects. So it is only logical to start the configuration structure 
 with a `vendorName`. A vendor has multiple packages so the next structure depth is `packageName`.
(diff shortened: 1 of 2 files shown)


sandrokeil:
Replacing *component* with *package* makes more sense, because it's more like Java.

Any thoughts on this?

sandrokeil:
I've added two new interfaces `HasDefaultOptions` and `HasOptionalOptions`. They are useful for auto discovery of factories and for default values of options.

sandrokeil:
I've renamed the interfaces for better understanding and reduced the complexity by removing `ÒbtainsOptions` and `HasDefaultOptions` interface.

The `PSR\Config\RequiresConfig` is the main interface and `PSR\Config\RequiresContainerId`, `PSR\Config\RequiresMandatoryOptions` and `PSR\Config\ProvidesDefaultOptions` are optional.

There is a [quick start guide](https://github.com/sandrokeil/interop-config/blob/master/doc/book/quick-start.md) which also explains the need.

proposed/config-meta.md
```diff
@@ -0,0 +1,134 @@
...
+
+## 1. Introduction
+This document describes the process and discussions that led to the Config PSR. It's goal is to explain the reasons 
+behind each decision.
+
+The word `Container` in this document is to be interpreted as the `ContainerInterface` of the PSR Container proposal.
+
+## 2. Why bother?
+The configuration of instances via factories is not uniform, uses different ways to retrieve and check the needed 
+options and it's not clear which configuration options needs the factory to create the instance.
```
samdark: the factory needs?

proposed/config-meta.md
```diff
@@ -0,0 +1,134 @@
...
+## 3. Scope
+
+### 3.1 Goals
+The goal set by the Config PSR is to standardize how factories uses a configuration to create instances, support for 
+auto discovery of needed configuration, to reduce boilerplate code and to make it more readable and easier to understand. 
+It can also be used to build the name of the *Container* entry.
+
+### 3.2 Non-goals
+* It's not a goal to define a whole project/library configuration structure.
+* It's not a goal to validate the configuration option values.
```
samdark: Is it a goal to validate structure?

proposed/config-meta.md
```diff
@@ -0,0 +1,134 @@
...
+
+## 4. History
+Before submitting the Config PSR to the PHP-FIG, the interfaces was first proposed in a project named
```
samdark: was → were?

proposed/config-meta.md
```diff
@@ -0,0 +1,134 @@
...
+
+### 3.2 Non-goals
+* It's not a goal to define a whole project/library configuration structure.
+* It's not a goal to validate the configuration option values.
+
+## 4. History
+Before submitting the Config PSR to the PHP-FIG, the interfaces was first proposed in a project named 
+[interop-config](https://github.com/sandrokeil/interop-config).
+
+The goal of the project was to provide a test-bed for implementing the interfaces and implementation and to pave the way
```
samdark: implementing the interfaces and implementation → implementing the interfaces

proposed/config-meta.md
```diff
@@ -0,0 +1,134 @@
...
+## 4. History
+Before submitting the Config PSR to the PHP-FIG, the interfaces was first proposed in a project named 
+[interop-config](https://github.com/sandrokeil/interop-config).
+
+The goal of the project was to provide a test-bed for implementing the interfaces and implementation and to pave the way 
+for the Config PSR.
+
+## 5. Interfaces
+There are six interfaces and some interfaces extends from another interface to ensure the uniform configuration structure. 
+If you confused about the naming of the interfaces or methods, note that I have used the Domain Driven Design approach.
```
samdark: you *are* confused

proposed/config-meta.md
```diff
@@ -0,0 +1,134 @@
...
+
+### 5.1 RequiresConfig
+This is the main interface and describes the default configuration structure. Since we have Composer more and more
```
samdark: Missing `,` after "Since we have Composer"

proposed/config-meta.md
```diff
@@ -0,0 +1,134 @@
...
+for the Config PSR.
+
+## 5. Interfaces
+There are six interfaces and some interfaces extends from another interface to ensure the uniform configuration structure. 
+If you confused about the naming of the interfaces or methods, note that I have used the Domain Driven Design approach.
+
+### 5.1 RequiresConfig
+This is the main interface and describes the default configuration structure. Since we have Composer more and more 
+libraries can be combined to create individual projects. So it is only logical to start the configuration structure 
+with a `vendorName`. A vendor has multiple packages so the next structure depth is `packageName`.
```
samdark: It's a big hassle to start it with a vendor name when most of the packages are provided by framework and the vendor is the same. It feels like unnecessary complication in this case.

proposed/config-meta.md
```diff
@@ -0,0 +1,134 @@
...
+
+
+The `canRetrieveOptions` method:
+
+* checks if options are available depending on implemented interfaces and checks that the retrieved options are an array 
+or have implemented \ArrayAccess. The `RequiresContainerId` interface is optional but MUST be supported.
+
+### 5.2 RequiresContainerId
+Factories allow to create more than one instance of the same class with another configuration. This can be achieved with 
+the third structure depth `containerId`. Now the structure is fully expanded.
```
samdark: Couldn't it be achieved by using another package name for the same class instead?

samdark: It would look much more compact.

proposed/config-meta.md
sandrokeil: Fixed, thanks.

proposed/config-meta.md
sandrokeil: Yes, it validates the configuration structure depending on the implemented interfaces.

proposed/config-meta.md
sandrokeil: Fixed, thanks.

proposed/config-meta.md
sandrokeil: Fixed, thanks.

proposed/config-meta.md
sandrokeil: Yes, sorry. I've fixed it.

proposed/config-meta.md
sandrokeil: Fixed, thanks.

proposed/config-meta.md
sandrokeil: [Zend Framework 2](http://framework.zend.com/blog/zend-framework-2-5-0-released.html) was split into separate repositories and there are many other good PHP libraries. I don't think it is complicated and works also for monolithic frameworks. The config entries should depends on the package name and you find the options for a package easier. Take a look at this example.

```php
return [
    // vendor
    'zend' => [
        // package
        'service-manager' => [
            // options
            'factories' => [ /* ... */],
        ],
        // package
        'navigation' => [
            // container id
            'default' => [ /* ... */],
        ],
        // package
        'view_manager' => [
            // container id
            'template-map-resolver' => [ /* ... */],
        ],
    ],
    // vendor
    'symfony' => [
        // package
        'services' => [
            /* ... */
        ],
    ],
];
```

proposed/config-meta.md
sandrokeil: This is possible, but it makes the config more unreadable and makes no sense. I think the uniform structure is missing in this case, beause you are mixing the package name with an id or something else. It's up to you to use this interface. There is no check, if the vendor/package name matches the current factory. Only the configuration array is checked for vendor/package.

The `RequriesContainerId` interface is only needed for factories which creates multiple instances with different configuration e.g. database connections (read/write). There are many PHP libraries which use this approach.

* [DoctrineORMModule](https://github.com/doctrine/DoctrineORMModule/blob/master/docs/configuration.md)
* [Prooph Service Bus](https://github.com/prooph/service-bus/blob/develop/config/prooph_service_bus.config.php)
* Zend Framework 2 and other frameworks in some cases

proposed/config-meta.md
samdark: It's a good to mention it.

michaelcullum:
We discuss new proposals for PSRs on the mailing list first normally. If you'd like to continue with this, please suggest it on there where we can discuss it going to a potential entrance vote.

https://groups.google.com/forum/#!forum/php-fig

PR #624, opened 2015-09-09
PSR2 Coding Style Typo Fix

JulienRamel:
The keyword `function` should not be translated to `fonction` in french (only in the text but not as a keyword).


samdark:
Thanks!

Issue #615, opened 2015-08-28
[PSR-6] Typo


rindeal:
In file cache.md on line 180, `if isHit()` should be `If isHit()`

Crell:
Thanks.  See PR.

PR #626, opened 2015-09-17
Minor typo fix

Crell:
As noted by #615


Crell:
I see no way that this could be controversial, so...

tedivm:
How dare you capitalize that "if" without at least two months of conversation and at least two discussions on the mailing list! Think of the children!

PR #628, opened 2015-09-21
PSR version

Crell:
Pre-draft version of the PSR-Version, for voting.

Please merge only when it passes an Entrance Vote.

### DIFF
--- a/proposed/version-support-meta.md
+++ b/proposed/version-support-meta.md
@@ -16,12 +16,22 @@ lag far behind PHP itself in terms of the languages features they adopt.
 2. Why Bother?
 --------------
 
+"What version can we support?" is a frequent challenge for PHP projects. That
+debate can eat up considerable development time, and fear of alienating users
+on older PHP versions can prevent a project from adopting new language features 
+that would benefit the project and its users.  That reluctance, in turn, means
+there is no incentive for public hosts to offer newer PHP versions that would 
+resolve that issue.
 
-Pros:
-* 
+This standard draws inspiration from the GoPHP5 initiative from 2007. In that 
+case, numerous projects and hosts collectively agreed on a drop-dead date for
+all PHP 4 versions, and that future releases would standardize on at least PHP 5.2.
+That initiative was successful in breaking the log-jam that kept PHP 4 alive well
+past its retirement and delayed the implementation of PHP 5.
 
-Cons:
-* 
+Rather than periodic "surge" efforts, this standard creates a rolling-GoPHP process
+that both projects and hosts can build into their planning.
+
 
 3. Scope
 --------
--- a/proposed/version-support-meta.md
+++ b/proposed/version-support-meta.md
@@ -49,31 +49,61 @@ they can factor into their normal upgrade schedules.
 This document does not seek to impose any architectural decisions on participating
 projects, nor to dictate what language features they use.
 
-4. Participating projects
+
+4. Example release schedule
+---------------------------
+
+The following examples help illustrate with examples how a participating project
+and host could operate with respect to PHP releases.
+
+## 4.1 For projects
+
+### 4.1.1 PHP minor releases.
+
+Flubber 4.2 is the current release.  PHP 9.2.3 is the current stable release.
+
+On 1 March 2020, PHP 9.3.0 is released.
+
+Flubber 4.3 is released on 20 March. It is unaffected, because PHP 9.3 has been
+out for less than 90 days.
+
+Flubber 4.4 is released on 14 July. It is required to support PHP 9.3 because
+it has been out for more than 90 days.
+
+On 1 March 2021, PHP 9.4. is released.
+
+Flubber 5.0 is released on 20 March. It is unaffected, because PHP 9.4 has been
+out for less than 6 months.
+
+
+
+## 4.2
+
+5. Participating projects
 -------------------------
 
 
-5. Participating hosts
+6. Participating hosts
 ----------------------
 
 
-6. People
+7. People
 ---------
 
-### 5.1 Editor
+### 7.1 Editor
 
 * Larry Garfield
 
-### 5.2 Sponsors
+### 7.2 Sponsors
 
 * Korvin Szanto, Concrete5 (Coordinator)
 * Cal Evans, Community
 
 
-7. Votes
+8. Votes
 --------
 
 
-8. Relevant Links
+9. Relevant Links
 -----------------
 
--- a/proposed/version-support-meta.md
+++ b/proposed/version-support-meta.md
@@ -4,7 +4,13 @@ PSR-Version Support Meta Document
 1. Summary
 ----------
 
-
+PHP has a long history of slow adoption of new versions of the language. A major
+factor in that slow adoption curve is the reluctance of public web hosts to upgrade
+their offerings without pressure from their customers, while major PHP projects
+are reluctant to adopt new PHP versions while few public hosts support it. Few
+projects wish to be the first to drop old versions for fear of alienating their
+user base. This chicken-and-egg problem is one of the reasons why many PHP projects
+lag far behind PHP itself in terms of the languages features they adopt.
 
 
 2. Why Bother?
@@ -22,17 +28,26 @@ Cons:
 
 ## 3.1 Goals
 
-* 
-* 
+This standard serves as a "community covenant" for projects and hosts to agree on
+an approximate schedule for when they will adopt and drop support for PHP versions.
+By synchronizing release schedules, no project is overly penalized for being first
+to drop support for a given PHP version and public hosts have a clear timeline
+they can factor into their normal upgrade schedules.
 
 ## 3.2 Non-Goals
 
+This document does not seek to impose any architectural decisions on participating
+projects, nor to dictate what language features they use.
 
-4. Approaches
--------------
+4. Participating projects
+-------------------------
 
 
-5. People
+5. Participating hosts
+----------------------
+
+
+6. People
 ---------
 
 ### 5.1 Editor
@@ -45,10 +60,10 @@ Cons:
 * Cal Evans, Community
 
 
-6. Votes
+7. Votes
 --------
 
 
-7. Relevant Links
+8. Relevant Links
 -----------------
 
--- a/proposed/version-support-meta.md
+++ b/proposed/version-support-meta.md
@@ -1,0 +1,54 @@
+PSR-Version Support Meta Document
+=================================
+
+1. Summary
+----------
+
+
+
+
+2. Why Bother?
+--------------
+
+
+Pros:
+* 
+
+Cons:
+* 
+
+3. Scope
+--------
+
+## 3.1 Goals
+
+* 
+* 
+
+## 3.2 Non-Goals
+
+
+4. Approaches
+-------------
+
+
+5. People
+---------
+
+### 5.1 Editor
+
+* Larry Garfield
+
+### 5.2 Sponsors
+
+* Korvin Szanto, Concrete5 (Coordinator)
+* Cal Evans, Community
+
+
+6. Votes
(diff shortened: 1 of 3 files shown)


GrahamCampbell:
This is such a good idea. :+1:

proposed/version-support.md
```diff
@@ -0,0 +1,113 @@
...
+hosting.  This standard applies to all offerings that include the ability to
+execute PHP code, and applies to all offered PHP SAPI configurations.  (For instance,
+it applies to both  command line PHP and web scripts.)
+
+
+## Expectations of projects
+
+A Major Release of a participating project MUST:
+
+* support the most recent stable Minor release of PHP as of thirty (30) days prior
```
KorvinSzanto: I had to read this a couple of times to understand what "its" in this context meant, maybe "Support the most recent stable minor PHP release older than 30 days."

proposed/version-support.md
```diff
@@ -0,0 +1,111 @@
...
+hosting.  This standard applies to all offerings that include the ability to
+execute PHP code, and applies to all offered PHP SAPI configurations.  (For instance,
+it applies to both command line PHP and web scripts.)
+
+
+## Expectations of projects
+
+A Major Release of a participating project MUST:
+
+* support the most recent stable Minor PHP release older than thirty (30) days.
```
Ocramius: This denies folks from trying out the new stuff.

proposed/version-support.md
```diff
@@ -0,0 +1,111 @@
...
+
+* support the most recent stable Major release older than six (6) months.
```
Ocramius: This denies folks from trying out the new stuff.

Crell: How? It is the opposite.  You need to run on a new release within a certain amount of time after it comes out.  EG, once PHP 7 comes out a project has 6 months to work on it, and when 7.1 comes out they have 30 days to work on it.  That clause says nothing about when they stop working on 5.4 or 5.5.

Ocramius: Then I am misunderstanding the wording. Doesn't this line force me to support PHP 5.6, even though I built for 7.0?

michaelcullum: No, it's saying you'll support a version within 6 months of it being released.

Crell: It's saying you have to support PHP 7 once it comes out.  You can stop supporting 5.6 whenever you want, but no later than X time after Internals drops support for it.

Crell: On the flipside, though, one of the Drupal release managers actually asked to include the inverse. Mainly, because there is no PHPUnit release now that is supported on both 5.5 and 7, which is a major epic fail for any project that wants to work on 5.5, 5.6, and 7 (like, say, Drupal).  I don't know that this document is the right way to solve that, but it's a legit concern.

michaelcullum: What do you mean by your last comment? I'm confused as to what 'the inverse' would be.

Crell: The Drupal RM, catch, is currently trying to deal with the fallout from PHPUnit having no 5.5-through-7 supported version.  He expressed interest in adding a clause to this document along the lines of "a project may not drop support for a supported PHP release in a minor or patch version", or something, to address that.  I am unsure about that myself.

Either way, that would be independent of the line we're discussing on, which says nothing of that sort. :smile:

KorvinSzanto: @Crell I wouldn't block it, but I definitely would not vote for this if we required KEEPING support for all supported PHP versions. Wouldn't that prevent people from releasing PHP 7 projects that adhere to this PSR?

michaelcullum: I would support not dropping support for a x.y PHP version in an maintenance/patch release however I can see it might be necessary to require a new version within a PHP branch (e.g. PHP x.y.z -> x.y.z+1) update in a patch/maintenance release in order to fix a critical bug.

proposed/version-support-meta.md
```diff
@@ -0,0 +1,130 @@
...
+PHP 9.0 security support ends entirely on 1 October 2020.  Flubber Hosting has
+until 1 October 2022 to cease offering PHP 9.0.x.
+
+PHP 10.0 is released on 1 April 2021.  Flubber Hosting has until 1 April 2022
+to offer it for both new and old accounts. It may be the default or not.  If
+a new minor or patch release comes out before then, it should be treated
+accordingly.
+
+
+5. Participating projects
```
michaelcullum: Do we really want to keep a list of participating projects/hosts within the meta document? I think this would be better on the website or something if we want to centralise the keeping of a list but either way it adds a burden of ensuring that 'participating projects' are keeping to it on the FIG.

Maybe it might be best to leave it up to a 3rd party to maintain a list or, as with other PSRs, projects/hosts simply declare compliance with the standard.

proposed/version-support-meta.md
```diff
@@ -0,0 +1,130 @@
...
+* Cal Evans, Community
+
+
+8. Votes
+--------
+
+
+9. Relevant Links
+-----------------
+
```
michaelcullum: Initial ML discussion topics?

proposed/version-support.md
```diff
@@ -0,0 +1,111 @@
...
+
+This standard serves as a "community covenant" for projects and hosts to agree on
+an approximate schedule for when they will adopt and drop support for PHP versions.
+By synchronizing release schedules, no project is overly penalized for being first
+to drop support for a given PHP version and public hosts have a clear timeline
+they can factor into their normal upgrade schedules.
+
+## Definitions
+
+* **Major version** - A major version of a project is a release that is documented
```
michaelcullum: Do we need to redefine semantic versioning or can we just reference it?

proposed/version-support.md
```diff
@@ -0,0 +1,111 @@
...
+
+* **Participating Host** - A *participating host* is a company that offers web
+hosting to customers that includes the ability to execute PHP code and openly and
+publicly declares its intent to comply with the expectations of this standard. A
+participating host MUST make reasonable effort to comply with the expecatations of
+this standard.
+
+Such hosts MAY be specific to a particular project or projects, or MAY offer
+general hosting for arbitrary PHP code. Such hosts MAY offer "shared hosting",
+virtual machines, dedicated hardware, cloud configurations, or any other form of
```
michaelcullum: For VMs and dedicated hardware is this not irrelevant as the management of the php version lays in the hands of the client?

proposed/version-support.md
```diff
@@ -0,0 +1,111 @@
...
+hosting.  This standard applies to all offerings that include the ability to
+execute PHP code, and applies to all offered PHP SAPI configurations.  (For instance,
+it applies to both command line PHP and web scripts.)
+
+
+## Expectations of projects
+
+A Major Release of a participating project MUST:
```
michaelcullum: What about: Not support PHP release that has been EOL'd for longer than 24 months?

proposed/version-support.md
```diff
@@ -0,0 +1,111 @@
...
+
+* support the most recent stable Minor PHP release older than thirty (30) days.
+
+* support the most recent stable Major release older than six (6) months.
+
+A Minor Release of a participating project MUST:
```
michaelcullum: SHOULD as the exception might be where BC could not be maintained?

proposed/version-support.md
```diff
@@ -0,0 +1,111 @@
...
+
+* support the most recent stable Minor release of PHP older than ninety (90) days.
+
+* support the most recent stable Major release of PHP older than twelve (12) months.
+
+
```
michaelcullum: What about patch versions should attempt to support the most recent stable minor release of PHP older than ninety days?

proposed/version-support-meta.md
Crell: I deliberately didn't reference SemVer, as not all projects use semver numbering.  Drupal, for instance, didn't pre-8.0.  (It basically skipped the Minor release.)

proposed/version-support-meta.md
Crell: Or something. :smile: Both of those sections were left in when I copied the meta doc from a previous PSR. I figured we can always remove them later if appropriate.

proposed/version-support.md
Crell: I deliberately didn't reference SemVer, as not all projects use semver numbering.  Drupal, for instance, didn't pre-8.0.  (It basically skipped the Minor release.)

proposed/version-support-meta.md
Crell: I'm fine with leaving it out here.

proposed/version-support.md
Crell: I suppose that varies by the host. I was talking to one in Bulgaria that stopped at the kernel, basically, so how this would apply to them I'm not sure. Probably we should tweak the language to say it applies only to hosts that provide a PHP version rather than simply an empty OS image.  (Hosts that don't provide/specify a PHP version are rather irrelevant to this discussion, as they are not holding back language progress.)

proposed/version-support.md
Crell: Yeah, need to add that.

proposed/version-support.md
catch56: So Drupal 8 will release with support for 5.5, 5.6 and 7.0.

We might eventually drop support for 5.5, but this is very unlikely to be before PHP itself does.

However, phpunit 4.8 does not support PHP 7, and phpunit 5.0 does not support PHP 5.5 - see https://github.com/sebastianbergmann/phpunit/wiki/Release-Announcement-for-PHPUnit-4.7.0

This means any project with a phpunit dependency, is itself unable to fully support the range of 5.5-7.0 which is what PHP itself does (for security at least).

This document as written appears to encourage that behaviour - new major versions are expected to adopt the latest PHP release and drop support for any release they like.

There's no requirement for the previous minor branch to support the latest PHP release, or to otherwise bridge the gap between versions.

Given different projects have different release cycles, you could end up with a situation where the only support release of library A doesn't support PHP 7 yet, and the only supported release of library B has dropped support for PHP 5.6, so that the two can't co-exist at all.

michaelcullum: @catch56 That link you said doesn't say PHP Unit 4.8 doesn't support PHP 7, only that it will be the last release to support 5.3, 5.4 and 5.5.

@sebastianbergmann Could you confirm?

Although I do agree wit the point being made that this situation could occur.

catch56: See https://github.com/sebastianbergmann/phpunit/issues/1882 and https://github.com/symfony/symfony/pull/16041

proposed/version-support-meta.md
```diff
@@ -0,0 +1,130 @@
...
+Flubber Hosting has until 1 June 2020 to offer PHP 9.2.3.  When it does so it
+replaces all its 9.2.1 offerings with 9.2.3.
+
+PHP 9.3.0 is released on 1 July 2020.  Flubber Hosting has until 1 January 2021
+to offer 9.3.0.  It may make it the default for new accounts or not, but both
+new and old accounts must have the option to use it.
+
+PHP 9.0 security support ends entirely on 1 October 2020.  Flubber Hosting has
+until 1 October 2022 to cease offering PHP 9.0.x.
```
carnage: Perhaps also add a clause suggesting that the default version offered should be one with _at a minimum_ active security support.

proposed/version-support.md
GrahamCampbell: All this stuff is inconsistent with semver anyway, because in semver, 0.x releases can contain BC breaks in minor releases.

michaelcullum:
@Crell What are your intentions with this now? I'd love to still see this as a PSR but if you're now no longer thinking of doing so, can we close this?

Crell:
I still have the text in my fork.  At this time the proposal seems dead for FIG purposes, so we can likely close this.

PR #629, opened 2015-09-22
PSR Storage

YP28:
It's been a while since I've worked on this, but I wanted to share it anyway to get things going. 

Feedback would be appreciated.


mmoreram:
@YP28 Can you please rebase all commits in a single one?
Some ideas here:

* exists(key) - Return if a certain key exists
* move(oldkey, newkey) - Moves a key value to another key

some of them can be added as well, but certainly should be discussed if they have really sense.

michaelcullum:
This should be discussed on the mailing list first really

mmoreram:
@michaelcullum sure :)

michaelcullum:
@mmoreram @YP28 If you still want to continue with this, please suggest it on the mailing list where we can discuss it going to a potential entrance vote.

PR #634, opened 2015-09-29
Added method Pool::resolveItem() to PSR-6 cache proposal

AndrewCarterUK:
Discussion thread: https://groups.google.com/d/msg/php-fig/l2Cl34zkAJ8/K7m6j-8UBgAJ

This method would allow for code to be rewritten from:

```php
function getThingyValue() {
    $item = $pool->getItem('thingy');

    if(!$item->isHit()){
        $item->set(computeThingy());
        $item->expireAfter(300);
        $item->addTag('foo');
        $item->addTag('bar');
        $pool->save($item);
    }

    return $item->get();
}
```

To

```php
function getThingyValue() {
    return $pool->resolveItem('thingy', 'computeThingy', function (CacheItemInterface $item) {
        $item->expireAfter(300);
        $item->addTag('foo');
        $item->addTag('bar');
    })->get();
}
```

### DIFF
--- a/proposed/cache.md
+++ b/proposed/cache.md
@@ -294,6 +294,35 @@ interface CacheItemPoolInterface
     public function getItem($key);
 
     /**
+     * Returns a Cache Item representing the specified key that is a cache hit.
+     *
+     * In the case of a cache miss, the Implementation Library will use the callback
+     * parameters to generate the Cache Item. The Cache Item will then be persisted
+     * and returned.
+     *
+     * @param string $key
+     *   The key for which to return the corresponding Cache Item.
+     *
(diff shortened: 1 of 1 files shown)


joshdifabio:
Would someone be so kind as to present a use case for `resolveItem()` returning a `CacheItemInterface` as opposed to the unwrapped value?

AndrewCarterUK:
@joshdifabio - In theory the method could return the item value (the implicit get()).

The proposal states that the CacheItemInterface object should encapsulate the storage and retrieval of cache items. I think an implicit get() might confuse this principal - but I'd be interested to hear the thoughts of the people who actually wrote the proposal.

PR #635, opened 2015-09-29
add SHOULD provide PGP key

lsmith77:


KorvinSzanto:
Should this also be in the XML?

quantumpacket:
How would someone know if the key/fingerprint being provided can be trusted to be the correct one? Correct me if I am wrong, as of now the spec does not enforce using a secure connection for the security domain, as all the example URLs are HTTP.

The public key fingerprint is provided for fetching via a keyserver,  or there is a link for downloading the key from the domain, this is all done over an insecure connection. So using those keys without knowing if they are not tampered/replaced with those of unknown parties would be just as bad as not using keys at all. HTTPS requires the cert to be signed by a trusted 3rd party. Similarly PGP keys can be signed by a trusted 3rd party, but that's impractical in this case.

Imagine a scenario where you have an open source project similar to Tor used for anonymous political activism, and you've found a vulnerability which you want to responsibly disclose to the developers. If you just grab the first key shown to you on their unencrypted security domain it could be tampered with by the NSA for example so they could then read the email that you encrypt with their keys, and know about the vulnerability beforehand. Then encrypt the email using the real key and send it on its way. Governments and hackers could pull this off.

We should be enforcing the security domain to be on an encrypted connection using the current security recommendations at the time. This should be simple for people to do once https://letsencrypt.org/ is up and running. Not only to prevent tampering of key information, but also the page contents. Encryption is not just about privacy, but also ensuring data integrity.

> Projects SHALL NOT use contact forms

Can someone explain to me why that is so? Other than the fact the site/form could be using the vulnerable software in question. As using a form over an encrypted connection is far more accessible to even the most novice developer, who may not know how to properly use PGP keys. PGP keys in the hands of those who don't understand how the trust model works is as good as useless.

Thoughts? I apologize in advance if this has already been discussed elsewhere.

KorvinSzanto:
I believe anyone being fooled by a spoof attack is doomed to be sending the
details to the attacker. Could we just make the PGP key a url field and
suggest public key servers?

lsmith77:
@quantumpacket good point. indeed a "SHOULD use SSL for all security relevant resources" would be good to add as well. aside from this I hope that this standard will encourage the creation of central services, ie. an index of relevant PGP keys, an aggregation of all advisories etc.

As for "no forms", yes its about the probability that people are using their own dog food either for the form itself or for other things on the same server. Also jn the spirit of this PR, who uses PGP to notify themselves for things entered into a web form? But we should maybe clarify the reasoning for this as you are not the first person to ask.

michaelcullum:
@php-fig/psr-9-10

vdelau:
As PSR-9/10 are abandoned, I intent to close this pull request, unless there is a reason to keep it.

PR #640, opened 2015-10-08
Switched 'he' to 'they'

AndrewCarterUK:


samdark:
Coordinator is a single person. Thus "he". Right?

quantumpacket:
Why was this closed? I have to agree with the usage of "they", or at least modify it so it's not specific to a male, unless for some odd reason the coordinator is a male only position.

samdark:
It's not about gender. It's about calling a single person "they".

quantumpacket:
https://www.oxforddictionaries.com/words/he-or-she-versus-they

michaelcullum:
+1 to the PR. 'they' can be used to refer to a singular person. Also, we can have more than one editor.

samdark:
OK. Wasn't aware of this way of using English. Thanks!

PR #643, opened 2015-10-12
Some stuff

michaelcullum:
We can do the heavy lifting in other branches on this repo then PR to task/psr-12 for us three to review and discuss, then task/psr-12 will be PR'd & merged to the main repo regularly (no need for review at that stage).


michaelcullum:
Opened on the wrong repo. Whoops.

PR #646, opened 2015-10-14
Fix docblock indentation

ajgarlag:


Crell:
PRs are not watched regularly.  Please inform the mailing list of any and all discussion.  Since this branch won't merge, though, I'm assuming it was already fixed.  (At least when I last inspected visually I think it was.)

PR #656, opened 2015-10-27
[PSR-12] Fix #655: Address unclear declare usage in proposed Extended Coding S…

### LINKED ISSUE #655: [PSR-12] Unclear `declare` usage in proposed *Extended Coding Style Guide*
The first line of "[Namespace, Strict Types and Use Declarations](https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#namespace-strict-types-and-use-declarations)" is unclear. It states the following:

> When present, there MUST be one blank line before the `namespace` declaration.

Many of the ensuing code examples then use the following structure:

```
<?php
namespace Vendor\Package;

use FooClass;
use BarClass as Bar;
use OtherVendor\OtherPackage\BazClass;

class ClassName extends ParentClass implements \ArrayAccess, \Countable
{
    // constants, properties, methods
}
```

The meaning is only clarified by the immediate example:

```
<?php
declare(strict_types=1);

namespace Vendor\Package;

use Vendor\Package\{ClassA as A, ClassB, ClassC as C};
use Vendor\Package\Namespace\ClassD as D;
use Vendor\Package\AnotherNamespace\ClassE as E;

use function Vendor\Package\{functionA, functionB, functionC};
use const Vendor\Package\{ConstantA, ConstantB, ConstantC};

class FooBar
{
    // ... additional PHP code ...
}
```

Right, so the "When present" in that first line is actually referring to the "Strict Types" declaration. After multiple readings, it seems like this prescription is trying to do too much by also including "Strict Types." However, because the prescription deals with the top-most section of a PHP file, it is natural to lump it with the other two. The solution, then, is more clarification.

**This is how the text should be clarified in writing:**

1. The section title should reflect the order that these PHP [keywords](http://php.net/manual/en/reserved.keywords.php) are declared. For example, "Strict Types, Namespace and Use Declarations," provides a more immediate hint of the prescribed order, just as [the original title from PSR-2](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#3-namespace-and-use-declarations) conveys.

2. Why is the first line talking about `namespace` at all? The second line does that. Why have two lines talking about `namespace` and the blank lines surrounding it? The first line should be prescribing how the `declare` keyword is used, thus handling its prescription of a blank line; it should not be offshoring this duty to `namespace`, which should not be tasked with handling the blank line prescriptions of its possibly present predecessor. Too much is being asked of `namespace`. A more clear first line is, "When present, there MUST be one blank line after the `declare` declaration," which is more meaningful. It also addresses one of the titular prescriptions, namely, "Strict Types," which is conspicuously absent from most of the section; it is only near the end where its prescribed usage is slightly more clarified, but by this point it is almost an afterthought, with whatever ideas in whatever order jotted down. The very body of a section should follow the ordering of its title. There is a necessary reading harmony that is not being observed. The additional meaning provided by the new first line from my example also means the second line regarding `namespace`, "When present, there MUST be one blank line after the namespace declaration," makes more sense in context.

3. The body should address the key ideas of the title in the same order they are written. This was mentioned in point (2) just above. Futhermore, the ideas should be written in a less thrown-together style.

4. This point will be moot if the first three are addressed. The following sentence is jarring, "There MUST not be a blank line before the strict types declaration," considering that it is not following the "MUST NOT" syntax.

danemacmillan:
…tyle Guide.

### DIFF
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -171,7 +171,7 @@ All files MUST declare strict types.
 Files containing only PHP MUST contain the strict type declarations on the
 first line preceeding the opening PHP tag.
 
-There MUST not be a blank line before the strict types declaration.
+There MUST NOT be a blank line before the strict types declaration.
 
 For example:
 
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -113,12 +113,15 @@ PHP [keywords] MUST be in lower case.
 The PHP reserved words `int`, `true`, `object`, `float`, `false`, `mixed`,
 `bool`, `null`, `numeric`, `string` and `resource` MUST be in lower case
 
-Namespace, Strict Types and Use Declarations
+Strict Types, Namespace, and Use Declarations
 --------------------------------------------
 
-When present, there MUST be one blank line before the `namespace` declaration.
+When present, there MUST be one blank line after the `declare` declaration.
 
 When present, there MUST be one blank line after the `namespace` declaration.
+
+When present, the `namespace` declaration MUST go after the `declare`
+declaration.
 
 When present, all `use` declarations MUST go after the `namespace`
 declaration.
@@ -511,7 +514,7 @@ for ($i = 0; $i < 10; $i++) {
 ```
 
 ### `foreach`
-    
+
 A `foreach` statement looks like the following. Note the placement of
 parentheses, spaces, and braces.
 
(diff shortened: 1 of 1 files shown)


samdark:
:+1:

samdark:
#655

michaelcullum:
:+1: Could someone merge this please?

samdark:
Done.

PR #657, opened 2015-10-27
Typo fix

marcusmoore88:
It seems like Paul M. Jones is missing his M. in a few spots.

Unless there is another Paul Jones I am ignorantly unaware of.


pmjones:
:+1: and thanks :-)

PR #659, opened 2015-10-28
Updating the return value for Pool::save() to be a boolean

dragoonis:


proposed/cache.md
```diff
@@ -332,8 +332,8 @@ interface CacheItemPoolInterface
      * @param CacheItemInterface $item
      *   The cache item to save.
      *
-     * @return static
-     *   The invoked object.
+     * @return boolean
```
GrahamCampbell: should be `bool` to match the rest of the PSRs

dragoonis: cheers. fixed.

PR #664, opened 2015-11-01
[PSR-6] Adding in missing singular deleteItem() and multiple saveItems()

dragoonis:


proposed/cache.md
```diff
@@ -348,6 +359,17 @@ interface CacheItemPoolInterface
     public function save(CacheItemInterface $item);
 
     /**
+     * Persists multiple cache items immediately.
+     *
+     * @param array|\Traversable $items
+     *   A traversable collection or array of CacheItemInterface[] items
```
xabbuh: Shouldn't this be `array of CacheItemInterface items`? Each is a cache item and not a collection of cache items.

dragoonis: @xabbuh thanks for your feedback. The intent here is to say it's an array of CacheItemIterface elements, or a Traverable collection of CacheItermInterface items. As per the function above getItems(). Do you want to suggest an alternative description?

xabbuh: I would keep it as is but simply remove the square brackets after the interface name:

> A traversable collection or array of CacheItemInterface items

dragoonis:
@xabbuh updated, as per your feedback. Thanks.

PR #665, opened 2015-11-01
[PRSR-6] fix return type

xabbuh:


samdark:
Why? `boolean` is valid type, isn't it?

xabbuh:
For two reasons: `bool` is used everywhere else in this document and `bool` is what you can use for type hints in PHP 7.

samdark:
You can use `boolean` for both type hinting and type casting in PHP.

milesj:
@samdark No you can't? http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration

samdark:
Wat... it could be used for type casting but not for scalar type hinting in PHP 7. I think it's a bug. Will report it to PHP team...

xabbuh:
IIRC the initial proposal wanted to support both, but current implementations don't seem to support that: https://3v4l.org/qXBeR

samdark:
https://bugs.php.net/bug.php?id=70833

dragoonis:
I'm merging this only because the rest of the PSR-6 document has 'bool' in it. Nothing to do with PHP7.

samdark:
Yes. Makes sense.

PR #667, opened 2015-11-03
[PSR-6] Adding Pool::deleteItem()

dragoonis:


tedivm:
Is there an emoticon for "I don't care either way so lets just merge this thing and pass us a standard"?

PR #669, opened 2015-11-03
[PSR-6] Adding in @throws blocks to any functions that receive invalid keys

dragoonis:


proposed/cache.md
```diff
@@ -300,6 +300,10 @@ interface CacheItemPoolInterface
      * @param array $keys
      * An indexed array of keys of items to retrieve.
      *
+     * @throws InvalidArgumentException
+     *   If any of the keys in $keys are not a legal value a \Psr\Cache\InvalidArgumentException
```
samdark: What's "not a legal value" in this case?

dragoonis: @samdark 
https://github.com/dragoonis/fig-standards/blob/cache-invalid-keys-throw/proposed/cache.md#definitions
-> definition of key

samdark: OK.

PR #670, opened 2015-11-06
Adding documentation for trait uses

rafreis:


proposed/extended-coding-style-guide.md
```diff
@@ -247,6 +247,87 @@ class ClassName extends ParentClass implements
...
+class ClassName
+{
+    use FirstTrait,
+        SecondTrait,
+        ThirdTrait,
+        FourthTrait,
+        FivethTrait,
+        SixthTrait,
+        SeventhTrait,
+        ighthTrait,
```
samdark: `E` is missing.

rafreis:
@samdark is not missing anymore.

samdark:
Overall it looks OK to me.

michaelcullum:
Traits existed in PHP 5.4 and there is nothing different from classes beyond changing the word class for trait. This is an extreme over complication of adding traits.

michaelcullum:
Ah it appears I misread this at first glance and this is about the use of the use statement within the class.

proposed/extended-coding-style-guide.md
```diff
@@ -247,6 +247,89 @@ class ClassName extends ParentClass implements
...
+When the `use` declaration exceeded 120 lines, use muiltiple lines for
+implement the traits.
+
+```
+<?php
+namespace Vendor\Package;
+
+use Vendor\Package\FirstTrait;
+use Vendor\Package\SecondTrait;
+use Vendor\Package\ThirdTrait;
```
michaelcullum: Three is plenty I think

proposed/extended-coding-style-guide.md
```diff
@@ -247,6 +247,89 @@ class ClassName extends ParentClass implements
...
+use Vendor\Package\FourthTrait;
```
michaelcullum: Two or three is plenty

proposed/extended-coding-style-guide.md
```diff
@@ -247,6 +247,89 @@ class ClassName extends ParentClass implements
...
+
+use Vendor\Package\FirstTrait;
+
+class ClassName
+{
+    use FirstTrait;
+}
+```
+
+When the `use` declaration exceeded 120 lines, use muiltiple lines for
```
michaelcullum: 120 lines?

proposed/extended-coding-style-guide.md
```diff
@@ -247,6 +247,89 @@ class ClassName extends ParentClass implements
 }
 ```
 
+### Using traits
+
+The `use` keyword used inside the classes to implement traits MUST be
+declared one line after the classe's opening brace, and MUST be indented
+with 4 space.
```
michaelcullum: The indentation note is unnecessary. Look at how we say similar things for classes (`Note indentation from the example`)

proposed/extended-coding-style-guide.md
```diff
@@ -247,6 +247,89 @@ class ClassName extends ParentClass implements
 }
 ```
 
+### Using traits
+
+The `use` keyword used inside the classes to implement traits MUST be
+declared one line after the classe's opening brace, and MUST be indented
```
michaelcullum: on the next line after the opening brace. Not one line after.

proposed/extended-coding-style-guide.md
michaelcullum: Follow through from how we do this elsewhere in PSR-12 and PSR-2 such as when we have arguments we say:
```
Argument lists MAY be split across multiple lines, where each subsequent line
 is indented once. When doing so, the first item in the list MUST be on the
 next line, and there MUST be only one argument per line.
 
 When the argument list is split across multiple lines, the closing parenthesis
 and opening brace MUST be placed together on their own line with one space
 between them.
```

Then we have a short example. There is no need to set the line limit again here, we already do that elsewhere.

proposed/extended-coding-style-guide.md
```diff
@@ -247,6 +247,89 @@ class ClassName extends ParentClass implements
...
+        SixthTrait,
+        SeventhTrait,
+        EighthTrait,
+        NinethTrait,
+        TenthTrait;
+}
+```
+
+When the class has nothing after the `use` declaration, the class
+closing brace MUST be one line after the `use` declaration.
```
michaelcullum: Again `one line` language

proposed/extended-coding-style-guide.md
```diff
@@ -247,6 +247,89 @@ class ClassName extends ParentClass implements
...
+
+use Vendor\Package\FirstTrait;
+
+class ClassName
+{
+    use FirstTrait;
+}
+```
+
+When the class has things after the `use` declaration, you MUST skip
```
michaelcullum: Let's not use the word 'things' and change skip one line to have a blank line after. See earlier in the spec (with namespaces) for correct and consistent language to use.

proposed/extended-coding-style-guide.md
KorvinSzanto: I don't like the idea of requiring that a project put the use statement on the line immediately following the opening brace. My own code looks like this:

```php
<?php
namespace Vendor\Package\Feature;

use Vendor\Package\OtherFeature\FeatureFactory;

/**
 * Class docblock
 */
class Feature implements FeatureInterface
{

    use FeatureTrait;

    /**
     * Method docblock
     * @return string
     */
    public function doFeature()
    {
        return 'Done';
    }

}
```

I like to have a single blank line before my use statement and another after my method declarations to give my class brackets a little bit of padding. I don't see any reason we should disallow this case, we should instead say something like "MUST NOT have more than one blank linefeed, MAY omit blank line completely."

proposed/extended-coding-style-guide.md
KorvinSzanto: +1

proposed/extended-coding-style-guide.md
KorvinSzanto: I'm not sure how much we should make a judgement on line length at all. The length requirements as written pretty much say  "Lines MAY be any length and MUST NOT be limited." with some extra wording around how and when analysis tools warn you. It's really a bunch of "SHOULD"s with the only requirement being that compliant code bases "MUST" not have a hard limit.  
Other than that, it reads more like it's directed at code analysis tools and not at code style compliance, and I think we should stay away from that.

rafreis:
@michaelcullum, @KorvinSzanto I guess that I had fixed all notes.

michaelcullum:
Let's merge this. We can always tidy up wording later on but the content here is good I think. @samdark please

samdark:
Done. Thank you @rafreis

NHoeller:
I just created a multiple (static only) autoloader class to load files first by using psr-4 convention. if this fails psr-0 convention is used. if thes both failed a namespace_like_class loader is used. i tried to follow also the other psr-standards.
if you want to i can try to upload my static class (and it's simple testing).

lacek:
The [`insteadof`](http://php.net/manual/en/language.oop5.traits.php#language.oop5.traits.conflict) and [`as`](http://php.net/manual/en/language.oop5.traits.php#language.oop5.traits.visibility) operators for using trait are not covered.

samdark:
https://github.com/php-fig/fig-standards/issues/738

PR #674, opened 2015-11-19
Highlight code in the proposed cache text

kabirbaidhya:
Highlight code in the text eg: Class/Interface names, function names, expressions etc


localheinz:
@kabirbaidhya 

:+1:

Crell:
I didn't see this until after the vote had already passed, so we can no longer make changes to the spec text.  Even for just formatting I'm very reluctant to touch a single byte.

Please, please, please, please post ANYTHING you do to the mailing list.  GitHub issues are *not* watched regularly, and GitHub's notification system is broken.  (I am subscribed to this whole repo but never get an email about it. I'm not the only one.)  Had this been spotted before we went to a vote I don't think anyone would have objected.

michaelcullum:
@Crell In the past we've always allowed formatting/typo fixes so long as they don't break any BC on the spec.

PR #682, opened 2015-11-25
Stream Interface

AndrewCarterUK:
There has been talk of a PSR for streams in the mailing list.

I'm going to use this pull request to attract sponsors for the proposal and then eventually an entrance vote.

Currently, these documents are a stripped down version of PSR-7. Proper attribution will need to be given to those who worked on this part of that proposal.

### DIFF
--- a/proposed/stream.md
+++ b/proposed/stream.md
@@ -1,11 +1,12 @@
-# Stream interface
+# Stream Interface
 
 This document describes a common interface for representing streams.
 
 Input and output streams are a fundamental concept in programming. Typically
-in PHP streams are represented as resources but there are benefits to a more
-object oriented approach. These benefits include exception handling on failure
-and more testable code.
+in PHP streams are represented as resources, but there are benefits to taking an
+object oriented approach. This specification seeks to improve interopability by
+providing a generic interface that PHP software utilising streams can depend
+upon without concern for the underlying implementation.
 
 ## 1. Specification
 
--- a/proposed/stream.md
+++ b/proposed/stream.md
@@ -1,0 +1,193 @@
+# Stream interface
+
+This document describes a common interface for representing streams.
+
+Input and output streams are a fundamental concept in programming. Typically
+in PHP streams are represented as resources but there are benefits to a more
+object oriented approach. These benefits include exception handling on failure
+and more testable code.
+
+## 1. Specification
+
+`StreamInterface` exposes several methods that enable streams to be read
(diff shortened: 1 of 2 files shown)


AndrewCarterUK:
https://groups.google.com/forum/#!topic/php-fig/LGYQ5x9L9-k

proposed/stream.md
```diff
@@ -0,0 +1,194 @@
...
+     * Seek to the beginning of the stream.
+     *
+     * If the stream is not seekable, this method will raise an exception;
+     * otherwise, it will perform a seek(0).
+     *
+     * @see seek()
+     * @see http://www.php.net/manual/en/function.fseek.php
+     * @throws \RuntimeException on failure.
+     */
+    public function rewind();
```
Chippiewill: Given that `seek(0)` is shorter than `rewind()` I am not sure of the benefit of this method, particularly when it enforces in it's description that it will perform seek(0)

proposed/stream.md
```diff
@@ -0,0 +1,194 @@
...
+     * @return bool
+     */
+    public function eof();
+
+    /**
+     * Returns whether or not the stream is seekable.
+     *
+     * @return bool
+     */
+    public function isSeekable();
```
Chippiewill: I feel like it would be better to spin off the seekable methods into a separate interface so that it's easier to enforce on typehints. Ditto with readable and writable.

Jean85:
This is stale, and it's from before the transition to the new FIG 3.0 process.

Closing, feel free to reach out to us if you're willing to restart the effort for this new PSR.

proposed/stream.md
azjezz: i highly agree with @Chippiewill here, the API would look something in lines of : 

```php
interface StreamInterface {}

interface ReadStreamInterface extends
  StreamInterface {}
interface WriteStreamInterface extends
  StreamInterface {}

interface DuplexStreamInterface extends
  ReadStreamInterface, WriteStreamInterface {}

interface SeekableStreamInterface extends
  StreamInterface {}

interface SeekableReadStreamInterface extends
  SeekableStreamInterface, ReadStreamInterface {}
interface SeekableWriteStreamInterface extends
  SeekableStreamInterface, WriteStreamInterface {}

interface SeekableDuplexStreamInterface extends
  DuplexStreamInterface, SeekableReadStreamInterface, SeekableWriteStreamInterface {}
````

PR #686, opened 2015-11-30
Update CONTRIBUTING.md

timwangdev:
Fix broken link


samdark:
Thanks!

KorvinSzanto:
Is this something that changed with the last site update? Should we 301 redirect `/faq/` to `/faqs/`?

samdark:
I have no idea how it was before but now it's like that so the change is definitely valid.

PR #688, opened 2015-12-02
Update the status of PSR-6

mnapoli:
If I'm not mistaken it is currently in voting phase.


samdark:
It is.

michaelcullum:
Voting is not a stage. It's Review then Accepted.
https://github.com/php-fig/fig-standards/blob/master/bylaws/004-psr-workflow.md#23-review

--
Michael C

samdark:
According to bylaws, yes. But see the footer of that page. There's `V = Voting` and it's actually what happens now.

mnapoli:
Yep that was confusing on the website (see http://www.php-fig.org/psr/#numerical-index)

samdark:
I think it's an important info for community watching php-fig and should be like that despite our internal bylaws.

PR #690, opened 2015-12-03
Dont always require a new blank line after opening php tag

michaelcullum:
@KorvinSzanto @samdark Could this be merged ASAP please

### DIFF
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -128,8 +128,8 @@ opening tag (which must be on the first line when declare statement(s) are prese
 
 Each declare statement (e.g. `declare(ticks=);`) MUST be on its own line.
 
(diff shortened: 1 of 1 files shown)


samdark:
Done.

ChristianVermeulen:
What is the status on having a blank line after the opening `<?php`? There was a **[voting in PSR-2](http://www.php-fig.org/psr/psr-2/#a3-survey-results)** but it was never implemented. When looking at the PSR-12 proposal **[there is a statement](https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#3-declare-statements-namespace-and-use-declarations)** about it, but the code examples don't seem to follow it. Is there any clarity on this yet?

samdark:
There's no statement about it.

PR #692, opened 2015-12-04
Remove confusing "non-blank" phrase

shadowhand:
Trailing whitespace is not allowed in any situation. A LF does not
constitute whitespace.


okj579:
Just to point it out: the "non-blank" does have an important purpose. It allows lines with _only_ whitespace, so blank lines can be indented:
```
{¶
␠␠␠␠//some code¶
␠␠␠␠¶
␠␠␠␠//some code¶
}¶
```
I don't like it personally, but opinions can differ on that.

PR #693, opened 2015-12-05
Slovak translation of accepted PSRs

skecskes:
Hi,

I spent my last 2 weeks of free time translating this to Slovak language. If nothing else, at least I learned a lot about PSRs. :) Only accepted PSRs are translated for now.

Thanks,
Stefan


samdark:
Great. Can you get reviews from Slovak developers here?

PR #695, opened 2015-12-06
Switched from specifying HTML to markup in general.

AndrewCarterUK:
PHP can be used to generate much more than HTML.

'Files containing **HTML** outside PHP opening and closing...'

Should be more general:

'Files containing **markup** outside PHP opening and closing...'

ping @michaelcullum


michaelcullum:
This is a good change. @samdark Could this be merged please?

samdark:
Great suggestion. Merged.

PR #697, opened 2015-12-07
Don't require a forward slash for exception example

michaelcullum:
Clarify that forward slash is not required.

### DIFF
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -193,9 +193,9 @@ use Vendor\Package\Namespace\{
 };
 ```
 
-Files containing markup outside PHP opening and closing tags MUST, on the first
-line, include an opening php tag, the strict types declaration and closing
-tag.
+When wishing to declare strict types in files containing markup outside PHP
+opening and closing tags MUST, on the first line, include an opening php tag,
+the strict types declaration and closing tag.
 
 For example:
 ```php
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -142,7 +142,7 @@ declaration.
 There MUST be one `use` keyword per declaration.
 
 You MUST NOT use `use` statements for classes in the root namespace. Therefore
-you should use `throw new \Exception();` instead of
+you should use `throw new \Exception();` or `throw new Exception();` instead of
 `use Exception; throw new Exception();`
 
 Use statements MUST be in blocks, grouped by varying entity (classes [inc. interfaces and traits],


michaelcullum:
@samdark @KorvinSzanto Merge ASAP please

KorvinSzanto:
I'm not sure that this resolves the point brought up on this list, why are we disallowing use statements for classes in the root namespace? https://3v4l.org/QSnaf

GrahamCampbell:
I think this example is meant to be when you're not in a namespace.

michaelcullum:
Indeed and it was due to comments of people saying they refuse to always use \Exception when they don't need to when actually the PSR never required the forward slash, it just disallowed the import.

`use Exception; throw new Exception();` is riddled with redundant/dead code and doesn't help clarify understanding at all, in fact it makes it possibly decreases readability as you have to check the two places whereas `throw new \Exception();` is so much easier to understand.

KorvinSzanto:
I'm not sure where I stand on the issue, it sounds a little heavy handed to me tbh. I just wanted to make sure we're really addressing the issue.

michaelcullum:
I've added a small clarification on here too. I didn't mean to add two things to the PR but oh well.

It's a common question and certainly within scope. Also most projects don't import root classes anyway.

simensen:
> Also most projects don't import root classes anyway.

I think this would be a case where it would actually make sense to do a survey. :) I don't know that I've seen consistency here anywhere and honestly I'm not sure that it matters too much.

For me, personally, I'd prefer people always `use` things and never have a leading `\` anywhere for the sake of consistency.

Requiring me to always `throw \RuntimeException` or worse yet `extend \RuntimeException` is probably going to get a "PSR-12 (without requiring the leading `\` for root classes)" label.

KorvinSzanto:
@simensen I agree. I'm not sure we need to make a judgement on this at all.

markstory:
@simensen I also think having explicit `use` gives code more consistency and makes the imports in any given file simpler to follow. 

Exempting root classes from `use` means the recommended standards are more complex as documenting when to use `use` is required.

pmjones:
@simensen Is exactly right -- this is not a PHP-7-specific requirement, and as such is amenable to survey.

judgej:
Personally I import any namespaces are are not in the classes namespace. That way you can see right at the top what namespaces that class uses. This includes \Exception, and project aliases such as Laravel's \Log, \Input etc.

Also - can we stop talking about *forward* slashes.

* / is a forward slash.
* \\ is a backward slash

judgej:
> When wishing to declare strict types in files containing markup outside PHP opening and closing tags MUST, on the first line, include an opening php tag, the strict types declaration and closing tag.

IMO this statement reads more like a work instruction, telling you what to do. It should just describe the end result, what the relationships are between constructs. TBH, I can't even follow what this sentence is saying anyway - it is just confusing even after reading five times. Can it be split up into smaller chunks?

PR #699, opened 2015-12-08
Preserve indentation whitespace in blank lines

dracony:
Editors such as atom remove all trailing whitespace by default, which is extremely annoying when when navigating code in some editors.


samdark:
IMO it should be always removed i.e. blank line should be just `\n` w/o any spaces.

michaelcullum:
Can we keep discussion on the ML please

GrahamCampbell:
> IMO it should be always removed i.e. blank line should be just \n w/o any spaces.

Totally agree. I don't know of anyone that puts trailing whitespace there. If this were to go through, nobody would adopt this standard. For example, failing to attract symfony and laravel would be a blow.

michaelcullum:
But for the record, huge :-1:

This is unnecessary, out of scope, most IDEs don't do this and most projects don't do this.

PR #700, opened 2015-12-08
Remove the root namespace statement

michaelcullum:
It's appearing controversial, which means it will hurt the upgrade path for enough people for it to not be worth it.

@samdark @KorvinSzanto Merge please?


samdark:
:+1:

DerManoMann:
:+1:

PR #701, opened 2015-12-08
Clarifying 4.2

skyzyx:


proposed/extended-coding-style-guide.md
```diff
@@ -296,9 +296,10 @@ class ClassName
 }
 ```
 
-Using multiple lines for traits, where each subsequent line is indented
-once. When doing so, the first item in the list MUST be on the same line
-of the `use` keyword, and there MUST be only one trait per line.
+Each individual Trait that is imported into a class MUST be included 
+one-per-line, and each inclusion MUST have its own `use` statement. 
+This matches the PSR-2 format for including Namespaces and Namespace
```
michaelcullum: This line and 302 are not needed. Please remove.

michaelcullum:
Good to merge once my comment has been addressed (Removal of `This matches the PSR-2 format for including Namespaces and Namespace Aliases`)

skyzyx:
Done. @michaelcullum

lcdss:
In this case, how can i indent this:

```php
use Authenticatable,
    Authorizable,
    CanResetPassword,
    SoftDeletes,
    EntrustUserTrait {
        EntrustUserTrait::can insteadof Authorizable;
        SoftDeletes::restore insteadof EntrustUserTrait;
    }
```

Like this?

```php
use Authenticatable;
use Authorizable;
use CanResetPassword;
use SoftDeletes;
use EntrustUserTrait {
    EntrustUserTrait::can insteadof Authorizable;
    SoftDeletes::restore insteadof EntrustUserTrait;
}
```

PR #702, opened 2015-12-08
Use "types and keywords" instead of "reserved words"

KorvinSzanto:
@ircmaxell mentioned that these aren't reserved words, I think "types and keywords" is a better thing to call em.


michaelcullum:
This is cool with me :+1: @samdark

PR #707, opened 2015-12-10
PSR-0 is no longer accepted as it's deprecated

philsturgeon:
https://groups.google.com/forum/#!topic/php-fig/LJP7dcgh7uQ

### DIFF
--- a/index.md
+++ b/index.md
@@ -6,11 +6,16 @@ As also described in the [PSR Workflow Bylaw][workflow]. The Editor, or editors,
 
 ## Index by Status
 
-### Accepted
+### Deprecated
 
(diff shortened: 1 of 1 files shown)


philsturgeon:
Thanks but that might have been a tad hasty. Looking at http://www.php-fig.org/psr/ and they have a whole different key, with `X = Rejected`.

That index.md file is being included from fig-standards on the website, so im not sure where the discrepancy is coming from.

philsturgeon:
@reinink halp

philsturgeon:
This should do it https://github.com/php-fig/php-fig.github.com/pull/177

PR #683, opened 2015-11-26
[PRE-DRAFT] Hypermedia Link PSR

Crell:
Proposed PSR for Hypermedia Link handling.  Currently in pre-draft state.  Merge only when it passes an entrance vote.

PR #717, opened 2015-12-29
Fix: Headline in PSR-13 proposal

localheinz:
This PR

* [x] fixes a headline using an invalid namespace, as pointed out by @renan

Follows https://github.com/php-fig/fig-standards/pull/683#discussion_r47165863.


Crell:
Already covered by #720.

PR #718, opened 2015-12-29
Markdown small fix

peterkokot:


keradus:
:+1:

evert:
Looks good to me. I have access to the merge button, but I'm not sure if I'm allowed to press it @michaelcullum  ?

michaelcullum:
@evert In general, I think the best policy will be to only have respective WGs touching their own specs and @php-fig/secretaries can pick up the slack.

PR #721, opened 2015-12-30
Expand and clarify attributes

Crell:
Suggested expansion/clarification of attribute handling.

### DIFF
--- a/proposed/links-meta.md
+++ b/proposed/links-meta.md
@@ -16,7 +16,6 @@ of the process of deciding what those links should be.
 The following questions are still outstanding, in the opinion of the Editor, and should be resolved.
 
 * LinkableInterface is a terrible name. Please suggest another one.
-* How do we support "empty" attributes, as HTML5 permits but few other systems do?
 * Should Href be a string, or can/should we use PSR-7 URI objects? I'm very very tempted to go with the latter.
 * Is there wording we should clean up around rel definitions?
 * Should the rel definition information move from the interfaces to the spec, or stay in the interface docblocks where
--- a/proposed/links.md
+++ b/proposed/links.md
@@ -24,7 +24,7 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
 
 ## 1. Specification
 
-### 1.1 
+### 1.1 Basic links
 
 A Hypermedia Link consists of, at minimum:
 - A URI representing the target resource being referenced.
@@ -33,6 +33,43 @@ A Hypermedia Link consists of, at minimum:
 Various other attributes of the Link may exist, depending on the format used. As additional attributes 
 are not well-standardized or universal, this specification does not seek to standardize them.
 
+For the purposes of this specification, the following definitions apply.
+
+*    **Implementing Object** -An object that implements one of the interfaces
+defined by this specification.
+
+*    **Serializer** - A library or other system that takes one or more Link objects
+and produces a serialized representation of it in some defined format.
+
+
+### 1.2 Attributes
+
+All links MAY include zero or more additional attributes beyond the URI and relationship.
+There is no formal registry of the values that are allowed here, and validity of values
+is dependant on context and often on a particular serialization format.  Commonly supported
+values include  'hreflang', 'title', and 'type'.
+
+Serializers MAY omit attributes on a link object if required to do so by the serialization
+format. However, serializers SHOULD encode all provided attributes possible in order to
+allow for user-extension unless prevented by a serialization format's definition.
+
+Some attributes (commonly hreflang) may appear more than once in their context. Therefore,
+an attribute value MAY be an array of values rather than a simple value. Serializers MAY
+encode that array in whatever format is appropriate for the serialized format (such
+as a space-separated list, comma-separated list, etc.).  If a given attribute is not
+allowed to have multiple values in a particular context, serializers MUST use the first
+value provided and ignore all subsequent values.
+
+If an attribute is boolean True, serializers MAY use abbreviated forms if appropriate
+and supported by a serialization format. For example, HTML permits attributes to
+have no value when the attribute's presence has a boolean meaning.  This rule applies
+if and only if the attribute is boolean True, not for any other "truthy" value
+in PHP such as integer 1.
+
+If an attribute is boolean False, serializers SHOULD omit the attribute entirely
+unless doing so changes the semantic meaning of the result. This rule applies if
+and only if the attribute is boolean False, not for any other "falsy" value in PHP
+such as integer 0.
 
 ## 2. Package
 
@@ -87,23 +124,10 @@ interface LinkInterface
     /**
      * Returns a list of attributes that describe the target URI.
      *
-     * The list should be specified as a key-value list.
-     *
-     * There is no formal registry of the values that are allowed here, and
-     * validity of values is dependant on context.
-     *
-     * Common values are 'hreflang', 'title', and 'type'. Implementors
-     * embedding a serialized version of a link are responsible for only
-     * encoding the values they support.
-     *
-     * Any value that appears that is not valid in the context in which it is
-     * used should be ignored.
-     *
-     * Some attributes, (commonly hreflang) may appear more than once in their
-     * context. Attributes such as those may be specified as an array of
-     * strings.
-     *
-     * @return string[]
+     * @return array
+     *   A key-value list of attributes, where the key is a string and the value
+     *  is either a PHP primitive or an array of PHP strings. If no values are
+     *  found an empty array MUST be returned.
      */
     public function getAttributes();
 }
--- a/proposed/links.md
+++ b/proposed/links.md
@@ -60,15 +60,15 @@ as a space-separated list, comma-separated list, etc.). If a given attribute is 
 allowed to have multiple values in a particular context, serializers MUST use the first
 value provided and ignore all subsequent values.
 
-If an attribute is boolean True, serializers MAY use abbreviated forms if appropriate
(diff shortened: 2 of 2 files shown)


proposed/links.md
```diff
@@ -33,6 +33,43 @@ A Hypermedia Link consists of, at minimum:
...
+
+If an attribute is boolean True, serializers MAY use abbreviated forms if appropriate
+and supported by a serialization format. For example, HTML permits attributes to
+have no value when the attribute's presence has a boolean meaning.  This rule applies
+if and only if the attribute is boolean True, not for any other "truthy" value
+in PHP such as integer 1.
+
+If an attribute is boolean False, serializers SHOULD omit the attribute entirely
+unless doing so changes the semantic meaning of the result. This rule applies if
+and only if the attribute is boolean False, not for any other "falsy" value in PHP
```
shadowhand: "falsey" might be more correct. I've seen both but prefer the "ey".

http://james.padolsey.com/javascript/truthy-falsey/

proposed/links.md
```diff
@@ -87,23 +124,10 @@ interface LinkInterface
...
-     *
-     * Any value that appears that is not valid in the context in which it is
-     * used should be ignored.
-     *
-     * Some attributes, (commonly hreflang) may appear more than once in their
-     * context. Attributes such as those may be specified as an array of
-     * strings.
-     *
-     * @return string[]
+     * @return array
```
shadowhand: Does it have to return an array, or can it return a `Traversable`?

proposed/links-meta.md
```diff
@@ -16,7 +16,6 @@ of the process of deciding what those links should be.
 The following questions are still outstanding, in the opinion of the Editor, and should be resolved.
 
 * LinkableInterface is a terrible name. Please suggest another one.
-* How do we support "empty" attributes, as HTML5 permits but few other systems do?
 * Should Href be a string, or can/should we use PSR-7 URI objects? I'm very very tempted to go with the latter.
```
shadowhand: Unrelated to this diff, but I suggest the recommendation be that it be a string or any object with a `__toString` method that converts the object to a URI, not just PSR-7 objects.

proposed/links.md
Crell: I considered that.  However, since attributes are always going to be fairly primitive, I don't think a traversable is necessary here.  The reason to use a traversable is when the dataset being returned could be very large or very slow to generate.  That could be the case for a set of Links, as they could be produced off of lazy DB queries, for instance, or the objects returned could be far more elaborate than just a trivial struct implementation of LinkInterface.  For attributes, since we're just returning, really, an array of strings (or an array of arrays of strings), neither of those situations applies.

proposed/links.md
Crell: Eh, sure. I have no strong preference so will add the e.

proposed/links-meta.md
Crell: We'll get there. :smile: One issue at a time.  (OK, two in some cases, but still.)

proposed/links.md
```diff
@@ -33,6 +33,43 @@ A Hypermedia Link consists of, at minimum:
 Various other attributes of the Link may exist, depending on the format used. As additional attributes 
 are not well-standardized or universal, this specification does not seek to standardize them.
 
+For the purposes of this specification, the following definitions apply.
+
+*    **Implementing Object** -An object that implements one of the interfaces
```
michaelcullum: I'm being pedantic but your wrapping is different from previously. Are you intending to standardise this at some point out of interest?

proposed/links.md
```diff
@@ -33,6 +33,43 @@ A Hypermedia Link consists of, at minimum:
...
+
+*    **Serializer** - A library or other system that takes one or more Link objects
+and produces a serialized representation of it in some defined format.
+
+
+### 1.2 Attributes
+
+All links MAY include zero or more additional attributes beyond the URI and relationship.
+There is no formal registry of the values that are allowed here, and validity of values
+is dependant on context and often on a particular serialization format.  Commonly supported
```
michaelcullum: dependant -> dependent

proposed/links.md
```diff
@@ -33,6 +33,43 @@ A Hypermedia Link consists of, at minimum:
...
+values include 'hreflang', 'title', and 'type'.
+
+Serializers MAY omit attributes on a link object if required to do so by the serialization
+format. However, serializers SHOULD encode all provided attributes possible in order to
```
michaelcullum: `MUST`, not `SHOULD` as you include the word `possible`?

In other words, It MUST include the ones that are possible, but MAY omit the ones a serializer does not have support for.

proposed/links.md
```diff
@@ -33,6 +33,43 @@ A Hypermedia Link consists of, at minimum:
...
+allow for user-extension unless prevented by a serialization format's definition.
+
+Some attributes (commonly hreflang) may appear more than once in their context. Therefore,
```
michaelcullum: backticks around `hreflang`? It looks odd in prose.

proposed/links.md
```diff
@@ -33,6 +33,43 @@ A Hypermedia Link consists of, at minimum:
...
+an attribute value MAY be an array of values rather than a simple value. Serializers MAY
+encode that array in whatever format is appropriate for the serialized format (such
+as a space-separated list, comma-separated list, etc.).  If a given attribute is not
```
michaelcullum: Double space

proposed/links.md
```diff
@@ -33,6 +33,43 @@ A Hypermedia Link consists of, at minimum:
...
+allowed to have multiple values in a particular context, serializers MUST use the first
+value provided and ignore all subsequent values.
+
+If an attribute is boolean True, serializers MAY use abbreviated forms if appropriate
```
michaelcullum: The capitalisation of `True` reads a bit oddly here.

proposed/links.md
michaelcullum: Is there a reason that `@return` is above the long description and not below it in the docblock?

proposed/links.md
```diff
@@ -87,23 +124,10 @@ interface LinkInterface
...
-     * Any value that appears that is not valid in the context in which it is
-     * used should be ignored.
-     *
-     * Some attributes, (commonly hreflang) may appear more than once in their
-     * context. Attributes such as those may be specified as an array of
-     * strings.
-     *
-     * @return string[]
+     * @return array
+     *   A key-value list of attributes, where the key is a string and the value
```
michaelcullum: Extra space indentation after the asterisk.

michaelcullum:
Otherwise :+1:

proposed/links.md
shadowhand: @michaelcullum that's the way @Crell and Drupal write inline descriptions. It's allowed by phpDoc.

proposed/links.md
shadowhand: Same thing here, the extra space denotes that it is a description.

proposed/links.md
shadowhand: :+1:

proposed/links.md
michaelcullum: Sure but it's inconsistent with the rest of the spec and other FIG specs.

proposed/links.md
michaelcullum: Same thing as above. It's not consistent with the rest of the spec or other PSRs.

proposed/links.md
shadowhand: I honestly had never noticed but you are quite right.

proposed/links.md
Crell: It's all done manually and I don't think we have a standard for where we wrap, so... meh. I hate manually wrapping text.  I hate Markdown.

proposed/links.md
Crell: I wasn't sure how strong to go here, since I don't know the rules of all of the various possible formats. I'm open to making it a MUST if others feel it is better.

proposed/links.md
Crell: After a period, double space As God Intended:tm:

proposed/links.md
Crell: I debated this one, too. All lowercase feels too informal, as we're talking about specific explicit values. All caps nobody uses but Drupal AFAIK.  So I settled on single capital to make it clear we're talking about the specific boolean values, not the concept of truthiness.

FIG really should develop a standard for that. :smile:

proposed/links.md
Crell: Yes, two spaces in from the @.  That's what I'm doing everywhere.

proposed/links.md
michaelcullum: My personal 2c is that 'SHOULD' should be avoided where possible in PSRs in general.

proposed/links.md
michaelcullum: The Drupal-way seems to make sense, and I see what you mean about formality. I'd be :+1: to all upper or backticks. Both are clear and probably look more... sensical?

proposed/links.md
michaelcullum: You're not in the rest of the spec (or other PSRs)?

proposed/links.md
shadowhand: I prefer wrapping in backticks whenever referring to programming elements.

Crell: *shrug*  I'll switch it to backticks.

proposed/links.md
Crell: PSR-6 used description-on-own-line as well.  It's considerably more readable.

Crell:
It looks like the remaining discussion here is all about whitespace alignment, not the text, so I'll go ahead and merge.  Start the new year with no open PRs and all of that. :smile: Thanks folks!

PR #728, opened 2016-01-04
There could be no list of extends

samdark:


donquixote:
Really? An interface can extend any number of parent interfaces.

samdark:
It's already back.

PR #730, opened 2016-01-05
Rewrite header section to be more readable

Crell:
As discussed on list.

### DIFF
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -120,42 +120,34 @@ The PHP types and keywords `array`, `int`, `true`, `object`, `float`, `false`, `
 3. Declare Statements, Namespace, and Use Declarations
 --------------------------------------------
 
-When present, there MUST be one blank line after the `declare` statement(s)
-e.g. `declare(ticks=);`
+The header of a PHP file may consist of a number of different blocks.  If present,
+each of the blocks below MUST be separated by a single blank line, and MUST NOT contain
+a blank line.  Each block MUST be in the order listed below, although blocks that are
+not relevant may be omitted.
 
-There MUST NOT be a blank line before declare statements such as those for strict
-types or ticks. They MUST be contained on the lines immediately following the
-opening tag (which must be on the first line when declare statement(s) are present).
+* Opening PHP tag <?php
+* File-level docblock.
+* One or more declare statements.
+* The namespace declaration of the file.
+* One or more class-based `use` statements.
+* One or more function-based `use` statements.
+* One or more `const`-based use statements.
+* The remainder of the code in the file.
 
-Each declare statement (e.g. `declare(ticks=);`) MUST be on its own line.
+When a file contains a mix of HTML and PHP, any of the above sections may still
+be used.  If so, they MUST be present at the top of the file, even if the
+remainder of the code consists a closing PHP tag and then a mixture of HTML and
+PHP.
 
 When the opening `<?php` tag is on the first line of the file, it MUST be on its
 own line with no other statements unless it is a file containing markup outside of PHP
 opening and closing tags.
 
-When present, there MUST be one blank line after the `namespace` declaration.
-
-When present, the `namespace` declaration MUST be on its own line.
-
-When present, all `use` declarations MUST go after the `namespace`
-declaration.
-
-There MUST be one `use` keyword per declaration.
-
-Use statements MUST be in blocks, grouped by varying entity (classes [inc. interfaces and traits],
-functions or constants). To elaborate, this means that any and all classes are in a block
-together; any and all functions are in a block together; and any and all constants must
-be grouped together. Within each block there MUST be no blank lines. If a block has
-multiple lines there MUST be a blank line before the first line and a blank line after
-the last line.
-
-The groups MUST be ordered such that classes (together with interfaces and traits) are first,
-followed by functions and then constants.
-
-Example of the above notices about namespace, strict types and use declarations:
+The following example illustrates a complete list of all blocks:
 
 ```php
 <?php
+
 declare(strict_types=1);
 
 namespace Vendor\Package;
@@ -165,7 +157,10 @@ use Vendor\Package\Namespace\ClassD as D;
 use Vendor\Package\AnotherNamespace\ClassE as E;
 
 use function Vendor\Package\{functionA, functionB, functionC};
-use const Vendor\Package\{ConstantA, ConstantB, ConstantC};
+use function Another\Vendor\function D;
+
+use const Vendor\Package\{CONSTANT_A, CONSTANT_B, CONSTANT_C};
+use const Another\Vendor\CONSTANT_D;
 
 class FooBar
 {
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -131,7 +131,7 @@ not relevant may be omitted.
 * The namespace declaration of the file.
 * One or more class-based `use` statements.
 * One or more function-based `use` statements.
-* One or more `const`-based use statements.
+* One or more constant-based `use` statements.
 * The remainder of the code in the file.
 
 When a file contains a mix of HTML and PHP, any of the above sections may still
@@ -147,6 +147,10 @@ The following example illustrates a complete list of all blocks:
 
 ```php
 <?php
+
+/**
+ * @file Contains an example of coding styles.
+ */
 
 declare(strict_types=1);
 


proposed/extended-coding-style-guide.md
```diff
@@ -120,42 +120,34 @@ The PHP types and keywords `array`, `int`, `true`, `object`, `float`, `false`, `
...
+a blank line.  Each block MUST be in the order listed below, although blocks that are
+not relevant may be omitted.
+
+* Opening PHP tag <?php
+* File-level docblock.
+* One or more declare statements.
+* The namespace declaration of the file.
+* One or more class-based `use` statements.
+* One or more function-based `use` statements.
+* One or more `const`-based use statements.
```
michaelcullum: Backticks around `use` and maybe say constant instead of `const` to be consistent here.

proposed/extended-coding-style-guide.md
```diff
@@ -120,42 +120,34 @@ The PHP types and keywords `array`, `int`, `true`, `object`, `float`, `false`, `
...
-opening tag (which must be on the first line when declare statement(s) are present).
-
-Each declare statement (e.g. `declare(ticks=);`) MUST be on its own line.
+The header of a PHP file may consist of a number of different blocks.  If present,
+each of the blocks below MUST be separated by a single blank line, and MUST NOT contain
+a blank line.  Each block MUST be in the order listed below, although blocks that are
+not relevant may be omitted.
+
+* Opening PHP tag <?php
+* File-level docblock.
```
michaelcullum: Do we want to enforce an empty line after the opening php tag and before the file-level docblock as is common practice? This would make sense but an exception to an otherwise clear rule would be a shame.

proposed/extended-coding-style-guide.md
```diff
@@ -120,42 +120,34 @@ The PHP types and keywords `array`, `int`, `true`, `object`, `float`, `false`, `
...
-
-There MUST NOT be a blank line before declare statements such as those for strict
-types or ticks. They MUST be contained on the lines immediately following the
-opening tag (which must be on the first line when declare statement(s) are present).
-
-Each declare statement (e.g. `declare(ticks=);`) MUST be on its own line.
+The header of a PHP file may consist of a number of different blocks.  If present,
+each of the blocks below MUST be separated by a single blank line, and MUST NOT contain
+a blank line.  Each block MUST be in the order listed below, although blocks that are
+not relevant may be omitted.
```
michaelcullum: Could we add a note to clarify that different statements must not be on the same line? (Such as namespace and opening php tag on the same line).

michaelcullum:
Otherwise :+1:

proposed/extended-coding-style-guide.md
Crell: I would much rather go with the easier-to-define, easier-to-remember, easier-to-write-a-parser-for approach.  I know that blank line is there in at least much of Drupal's code base and our doc parser handles it fine, so it wouldn't harm us in the least. :smile:

PowerKiKi: I would also prefer to enforce blank lines. The visual difference it may makes for reading is, IMHO, insignificant compared to the huge benefit of simplicity if we enforce it.

This PR makes the rules really trivial to understand and apply. And I believe it is something that should be aimed for.

teohhanhui: Is this an ordered list? If so, it's not clear from reading this.

If it's an ordered list, I think it'd be weird for the file-level docblock to go before the `declare` statement.

proposed/extended-coding-style-guide.md
```diff
@@ -120,42 +120,34 @@ The PHP types and keywords `array`, `int`, `true`, `object`, `float`, `false`, `
...
-together; any and all functions are in a block together; and any and all constants must
-be grouped together. Within each block there MUST be no blank lines. If a block has
-multiple lines there MUST be a blank line before the first line and a blank line after
-the last line.
-
-The groups MUST be ordered such that classes (together with interfaces and traits) are first,
-followed by functions and then constants.
-
-Example of the above notices about namespace, strict types and use declarations:
+The following example illustrates a complete list of all blocks:
```
teohhanhui: It's not complete as file-level docblock is not shown here.

proposed/extended-coding-style-guide.md
Crell: From the text: " Each block MUST be in the order listed below".

So, yes, it's an ordered list.

Crell:
Made a few small adjustments per discussion above. I think this is ready to merge now if the Editor agrees.

KorvinSzanto:
Looks like a good improvement to me, thank you!

le6o:
Most of the code samples need updating to follow these new rules.
e.g. Line 43 and 257 plus others.

proposed/extended-coding-style-guide.md
samdark: I was thinking about this for a long time:

1. I tend to handle opening PHP tag separately i.e. as non-block so we won't force a blank line after it.
2. "each of the blocks below MUST be separated by a single blank line" could be re-phrased to be "each of the blocks below MUST be separated by a single blank line from each other" to exclude separation from opening `<?php` tag.

teohhanhui: I thought the point of PSR-12 is to enforce more consistency... So forcing a blank line after the opening `<?php` tag is a good thing.

teohhanhui: I'd still argue for `declare` before the file-level docblock. It's something that changes the PHP compiler's behavior, and as such fits perfectly right after the `<?php` opening tag.

Example:

```php
<?php
declare(strict_types=1);

/*
 * This file is part of the Symfony package.
 *
 * (c) Fabien Potencier <fabien@symfony.com>
 *
 * For the full copyright and license information, please view the LICENSE
 * file that was distributed with this source code.
 */

namespace Symfony\Component\PropertyInfo\Extractor;

use Symfony\Component\PropertyInfo\PropertyListExtractorInterface;
use Symfony\Component\Serializer\Mapping\Factory\ClassMetadataFactoryInterface;

/**
 * Lists available properties using Symfony Serializer Component metadata.
 *
 * @author Kévin Dunglas <dunglas@gmail.com>
 */
class SerializerExtractor implements PropertyListExtractorInterface
{
    ...
}
```

samdark: Probably but that's not what's enforced currently and I'm not sure it should not be enforced the other way around. Some reasearch:

### No blank lines

- [Yii](https://github.com/yiisoft/yii2/blob/master/framework/caching/ApcCache.php)
- [Aura](https://github.com/auraphp/Aura.Di/blob/3.x/src/Container.php)
- [Zend](https://github.com/zendframework/zend-expressive/blob/master/src/Application.php)
- [Slim](https://github.com/slimphp/Slim/blob/3.x/Slim/App.php)
- [Propel](https://github.com/propelorm/Propel2/blob/master/src/Propel/Common/Exception/SetColumnConverterException.php)
- [PPI](https://github.com/ppi/framework/blob/master/src/Console/Application.php)
- [Doctrine](https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Event/ListenersInvoker.php)
- [CakePHP](https://github.com/cakephp/cakephp/blob/master/src/Collection/CollectionInterface.php)

### Blank lines

- [Laravel](https://github.com/laravel/framework/blob/5.2/src/Illuminate/Support/ClassLoader.php)
- [Symfony](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Validator/ConstraintViolation.php)
- [Composer](https://github.com/composer/composer/blob/master/src/Composer/Command/ArchiveCommand.php)

Crell: Drupal currently has no standard, but is leaning toward adopting a blank line standard: https://www.drupal.org/node/2248463

(Which references this issue.)

I'd favor having the blank line for consistency and ease of explanation.  Popularity contest alone should not be the deciding factor (although it is a factor to consider).

samdark: If we'll not mention opening PHP tag as a block neither consistency nor explanation would be hurt.

proposed/extended-coding-style-guide.md
```diff
@@ -120,42 +120,38 @@ The PHP types and keywords `array`, `int`, `true`, `object`, `float`, `false`, `
...
 
 ```php
 <?php
+
+/**
+ * @file Contains an example of coding styles.
+ */
+
 declare(strict_types=1);
```
glen-84: Don't you think that there should be a space on either side of the equals sign, as there is for assignment, default parameter values, etc.?

```php
declare(strict_types = 1);
```

Crell:
Is there a reason this hasn't been merged yet? I thought the Editor was leaning toward including the leading line break, and either way it's still a cleaner language than what's there now.  Let's get this merged and move on, please.

samdark:
@Crell I'll merge it when I'll get back from vacation.

michaelcullum:
If Korvin wants this then I can merge it now but it should have an okay from the editor before it gets merged. Korvin also has merge permissions himself.

samdark:
See https://github.com/php-fig/fig-standards/pull/730#discussion_r57304680. I don't think it should be merged as is.

glen-84:
Is this not something that should be part of the survey? (the survey that never seems to be ready)

I happen to agree with this (blank line after `<?php`), but it shouldn't be a decision made by a single person (or just a few) in my opinion.

PowerKiKi:
I am also waiting for the survey so that everyone's opinion can be taken
into consideration properly.

... although I'd rather see this PR merged asap.

PowerKiKi:
In case people here missed it, the survey is now available: https://goo.gl/forms/SonvjwqicPoBBwwl1

PR #731, opened 2016-01-05
Make the wording against underscore prefixes stronger

Crell:
As discussed on list. This is still being discussed I think but here's the PR for it if we go that route.

### DIFF
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -363,8 +363,9 @@ The `var` keyword MUST NOT be used to declare a property.
 
 There MUST NOT be more than one property declared per statement.
 
-Property names SHOULD NOT be prefixed with a single underscore to indicate
-protected or private visibility.
+Property names MUST NOT be prefixed with a single underscore to indicate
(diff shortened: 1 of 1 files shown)


samdark:
Could cause less adoption rate but :+1: from me.

samdark:
@michaelcullum, @KorvinSzanto ?

michaelcullum:
:+1:

KorvinSzanto:
:+1:

samdark:
Thanks!

dereuromark:
:-1: This is a nogo.
Those underscores can be useful to show the visbility on e.g. PRs and other (diff) tools where you do not have the direct visibility via IDE.

    return $this->_doSth();

etc.

Jean85:
I'm sorry but this part is already an established standard in PSR-2, see: https://www.php-fig.org/psr/psr-2/#42-properties

dereuromark:
Yeah, with SHOULD NOT, which means you can ignore it for good reasons.

Jean85:
My personal opinion is that that is a bad practice. Changing the visibility of a property shouldn't require a rename, for example. Hence, consistency is hard to maintain without that rule.

PR #733, opened 2016-01-05
Add space padding to both sides of a return declaration colon

Crell:
As discussed on list.

### DIFF
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -479,8 +479,8 @@ $app->get('/hello/{name}', function ($name) use ($app) {
 });
 ```
 
-When you have a return type declaration present there MUST be one space after
-the colon with followed by the type declaration. The colon and declaration MUST be
+When you have a return type declaration present there MUST be one space preceeding and following
+the colon followed by the type declaration. The colon and declaration MUST be
 on the same line as the argument list closing parentheses with no spaces between
 the two characters. The declaration keyword (e.g. string) MUST be lowercase.
 
@@ -492,7 +492,7 @@ namespace Vendor\Package;
 
 class ReturnTypeVariations
 {
-    public function functionName($arg1, $arg2): string
+    public function functionName($arg1, $arg2) : string
     {
         return 'foo';
     }


samdark:
That's subject for survey/voting. It's a single space after for many implementations so I don't think it's good to change it: http://www.typescriptlang.org/Handbook#interfaces

michaelcullum:
:+1:

samdark:
@michaelcullum how about conter-example of TypeScript?

proposed/extended-coding-style-guide.md
```diff
@@ -479,8 +479,8 @@ $app->get('/hello/{name}', function ($name) use ($app) {
 });
 ```
 
-When you have a return type declaration present there MUST be one space after
-the colon with followed by the type declaration. The colon and declaration MUST be
+When you have a return type declaration present there MUST be one space preceeding and following
+the colon followed by the type declaration. The colon and declaration MUST be
```
donquixote: typo? Should be "preceding".

samdark: Yes.

glen-84:
My vote: -1

^ Did you see what I did there?

Jokes aside – spaces on either side dissociates the type from the function declaration.

PowerKiKi:
+1 should be merged

michaelcullum:
Can we stop 'voting' please? This is waiting for the proper survey, and any +1s/-1s now will just be discarded.

samdark:
Here's the survey for voicing your opinion. It would matter: http://goo.gl/forms/LyXppn4i5e

KorvinSzanto:
@samdark I think it's probably not a good idea to distribute individual surveys like that. I've set aside some time next weekend to give that survey a pass, lets just focus on making the one monolith survey as complete as we can and release that.

samdark:
What if more topics would arise?

PowerKiKi:
I agree with @KorvinSzanto, a single survey as complete as possible would be for the best IMHO

michaelcullum:
:+1: Korvin

KorvinSzanto:
@samdark I'm definitely willing to discuss it as a group if you want, we can schedule a time in IRC to meet up or something and we can talk about whether we want to do individual surveys or a single huge one.

I figure if more issues come up after the survey, we can simply do another survey.

samdark:
OK.

hrvoj3e:
A single BIG survey would take too much time for people who track PSR-12 regularly 
_(Assuming it would have all questions I already answered before)._
It would make sense for people which are new to PSR-12 to express their opinion in bulk without needing to search for links to surveys in comments.

Could we have a thread or wiki page with links to all surveys and/or one big survey?

samdark:
No, there's no big survey yet. It's in the making.

PowerKiKi:
In case people here missed it, the survey is now available: https://goo.gl/forms/SonvjwqicPoBBwwl1

pierniq:
from typographic point of view, in most non-programming languages (except french), there should be [no space before colon](https://en.wikipedia.org/wiki/Colon_(punctuation)#Spacing).

donquixote:
@pierniq: I would argue that this colon is semantically different from a
non-programming language colon.

E.g. in a switch statement, the colon after an option is semantically
similar to normal language colon.
But in this case I think it is just an arbitrary symbol, and looking at it
from a normal-language perspective can be misleading.

This is not really meant as an argument for or against, I just want to
avoid that we do things for the wrong reasons.

pierniq:
from PHP 7.1 nullable types was introduced.
it makes me sure, that "that" colon means the same as "natural language colon".
```php
function foo(): ?int {}
```
it means: "return value type could be:
- int,
- null."

PowerKiKi:
I fail to see how the introduction of nullable types changes anything to the semantic. But more importantly than that, this issue was covered in the survey. And the survey result shall give this debate a final conclusion, [hopefully this week](https://groups.google.com/d/msg/php-fig/fGDuNFKy390/qBT_0XQpCQAJ). Meanwhile I don't think there is anything else constructive to add here.

michaelcullum:
The survey was in favour of the current implementation.

PR #734, opened 2016-01-11
Fix incorrect syntax for markdown link in PSR-6 meta document

dregad:


samdark:
What's incorrect exactly?

dregad:
See the patch... Markdown link syntax is `[text](url)` (URL in parenthesis), not `[text][url]` (square brackets)

samdark:
Ah, right :)

michaelcullum:
:+1:

@Crell

PR #740, opened 2016-01-23
Tabs are better than spaces

nemanjan00:
1) You can define width of tab in your editor
2) Files are smaller
3) You can not half indent


michaelcullum:
PSR-2 is final and this is not going to be discussed again. A decision was made, it won't be unmade.

PR #744, opened 2016-02-16
[PSR-12] Move note and example on multi-line arg

teohhanhui:
The note and example on "a single argument being split across multiple lines" were in the wrong section

(4.5 Method and function Arguments) -> (4.7 Method and Function Calls)

### DIFF
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -445,9 +445,7 @@ class ClassName
 
 Argument lists MAY be split across multiple lines, where each subsequent line
 is indented once. When doing so, the first item in the list MUST be on the
-next line, and there MUST be only one argument per line. A single argument being
-split across multiple lines (As might be the case with an anonymous function or
-array) does not constitute splitting the argument list itself.
+next line, and there MUST be only one argument per line.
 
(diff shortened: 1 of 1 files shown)


michaelcullum:
:+1: This draws attention to the fact in the spec we incorrectly use `arguments` in all instances when we should be talking about parameters [when discussing the method/function signature] and arguments [when discussing calls]. 

There might be an argument for having this stipulated for parameters, although I'm struggling to think of when an parameter might be multiple lines. Perhaps when the default parameter value is an array?

@KorvinSzanto @samdark Thoughts?

samdark:
For parameters complex array defaults are indeed possible.

Overall, parameters vs arguments usage is confusing. Parameters are what's in method signature while arguments is what's passed to methods when calling them, right?

KorvinSzanto:
I think we can omit judgement on how to handle really complex multiline parameter defaults.

KorvinSzanto:
:+1:

samdark:
@KorvinSzanto so merging as is?

KorvinSzanto:
I'd say so, yeah

samdark:
@teohhanhui thanks!

PR #746, opened 2016-03-10
PSR-7: Describe the format of $_FILES correctly for nested element names

akrabat:
When the input element's name is a nested array, then format of $_FILES is more complicated as the nested key names appear underneath the five _FILES core array element names.

This PR updates to match reality rather than the wishful thinking that it might have been reasonably logical…

### DIFF
--- a/accepted/PSR-7-http-message.md
+++ b/accepted/PSR-7-http-message.md
@@ -433,15 +433,31 @@ In the case of an input using array notation for the name:
 `$_FILES` ends up looking like this:
 
 ```php
-array(
-    'my-form' => array(
-        'details' => array(
-            'avatar' => array(
-                'tmp_name' => 'phpUxcOty',
-                'name' => 'my-avatar.png',
-                'size' => 90996,
-                'type' => 'image/png',
-                'error' => 0,
+array (
+    'my-form' => array (
+        'name' => array (
+            'details' => array (
+                'avatar' => 'my-avatar.png',
+            ),
+        ),
+        'type' => array (
+            'details' => array (
+                'avatar' => 'image/png',
+            ),
+        ),
+        'tmp_name' => array (
+            'details' => array (
+                'avatar' => 'phpmFLrzD',
+            ),
+        ),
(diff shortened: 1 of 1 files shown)


michaelcullum:
@weierophinney Thoughts?

weierophinney:
@michaelcullum Essentially, we messed up on this, though the idea is correct. What @akrabat proposes with the patch is fixing the spec to *what was intended*; we were trying to illustrate the weird behavior of PHP in this scenario, and how implementations should munge it to `UploadedFileInterface` instances. The problem is that we got the details of PHP's behavior incorrect for purposes of the narrative, despite the intention.

Diactoros implements it correctly in terms of intention; however, Slim originally implemented it *to the specification*, which meant that their implementation was broken until @akrabat discovered the discrepancy.

As such, we need to do *something*, to ensure other implementors do not stumble into the same issues. My understanding is that we cannot actually change anything substantively in the spec except typos without an errata. The question is whether or not this could be considered a typo, or deserves an errata. Thoughts?

samdark:
I'd make a revision 2 (2016-03-xx). If it requires changing of bylaws then we need that change. There are too many cases when we really need to fix things.

weierophinney:
@samdark This is not a revision; it's not changing intent of the specification in any way. It's a correction of a mistake in the text. That's why I was suggesting it *may* fall under the realm of "typo", but I'm more than happy to raise it as an errata if necessary.

michaelcullum:
@weierophinney I'm thinking this falls more under Errata as it previously could have been interpreted in two ways (intended [Zend] way and described [Slim] way)

akrabat:
It's a factually incorrect description of PHP behaviour, rather than any statement about the interfaces and how to use them.

As long as the published document on http://www.php-fig.org/psr/psr-7/ doesn't contain this incorrect text, I don't mind the process you use to sort it out :)

Jean85:
This was pending for far too long :disappointed: 

Closing in favor of #1110, which is just a rebase.

PR #747, opened 2016-03-12
[PSR-1] Extend the constant names standard to all constants, not only class constants

helysousa:
Constants can be declared outside classes also, thus, in order to harmonize their names, I am proposing to remove the restriction of PSR-1 to class constants.


samdark:
It changes actual rules so can't be considered errata thus we aren't able to merge it.

PR #750, opened 2016-04-07
Suggestion of standardizing interfaces for sending email

flug:


samdark:
Please announce it at https://groups.google.com/forum/#!forum/php-fig

michaelcullum:
As previously mentioned by @samdark, we discuss new proposals for PSRs on the mailing list first normally. If you'd like to continue with this, feel free suggest it on there where we can discuss it going to a potential entrance vote.

https://groups.google.com/forum/#!forum/php-fig

PR #756, opened 2016-05-13
[PSR-7] Errata: each URI modification should result in a valid URI

Tobion:
My proposal for #619

Also replaced "hostname" wording with "host" in the few left-over places as hostname is not correct.
Hostname is a reg-name (usually fully-qualified-domain). But the host can also be an IP address.

### DIFF
--- a/accepted/PSR-7-http-message-meta.md
+++ b/accepted/PSR-7-http-message-meta.md
@@ -647,3 +647,21 @@ used to populate the headers of an HTTP message.
 * Anton Serdyuk
 * Phil Sturgeon
 * Chris Wilkinson
+
+## 7 Errata
+
+### 7.1 Invalid `UriInterface` after modification of a component (2016/05/13)
+
+It is expected that an instance of `UriInterface` can be represented as a valid
+URI reference via the `UriInterface::__toString` method. This means that the
+modifying methods of the URI components, e.g. `withHost($host)`, have to ensure
+this is still the case. If a modification would result in an invalid URI, the
+methods MUST throw an `\InvalidArgumentException`. Invalid URIs can not only be
+caused by invalid characters in a URI component but also because the given
+URI components cannot be recomposed to a valid URI reference which may even
+depend on the current URI scheme. Two examples are:
+
+- `$uri->withHost('')->withPort(8080)` should fail because there cannot be a
+  port without a host.
+- `$uri->withPath('rootless')->withHost('example.com')` should fail because
+  a URI with an authority cannot have a relative path.
--- a/accepted/PSR-7-http-message-meta.md
+++ b/accepted/PSR-7-http-message-meta.md
@@ -661,7 +661,7 @@ caused by invalid characters in a URI component but also because the given
 URI components cannot be recomposed to a valid URI reference which may even
 depend on the current URI scheme. Two examples are:
 
-- `$uri->withHost('')->withPort(8080)` should fail because there cannot be a
-  port without a host.
+- `$uri->withHost('example.org')->withPath('//two-slashes')->withHost('')` should
+  fail because the path of a URI without an authority must not start with two slashes.
 - `$uri->withPath('rootless')->withHost('example.com')` should fail because
   a URI with an authority cannot have a relative path.
--- a/accepted/PSR-7-http-message.md
+++ b/accepted/PSR-7-http-message.md
@@ -231,7 +231,7 @@ of the request line. The request target can be one of the following forms:
 Aside from these request-targets, there is often an 'effective URL' which is
 separate from the request target. The effective URL is not transmitted within
 an HTTP message, but it is used to determine the protocol (http/https), port
-and host for making the request.
+and hostname for making the request.
 
 The effective URL is represented by `UriInterface`. `UriInterface` models HTTP
 and HTTPS URIs as specified in RFC 3986 (the primary use case). The interface
@@ -266,7 +266,7 @@ OPTIONS * HTTP/1.1
 ```
 
 But the HTTP client will be able to use the effective URL (from `getUri()`),
-to determine the protocol, host and TCP port.
+to determine the protocol, hostname and TCP port.
 
 An HTTP client MUST ignore the values of `Uri::getPath()` and `Uri::getQuery()`,
 and instead use the value returned by `getRequestTarget()`, which defaults
@@ -777,7 +777,7 @@ interface MessageInterface
      *
      * @param StreamInterface $body Body.
      * @return self
-     * @throws \InvalidArgumentException when the body is not valid.
+     * @throws \InvalidArgumentException When the body is not valid.
      */
     public function withBody(StreamInterface $body);
 }
@@ -1232,7 +1232,7 @@ interface ResponseInterface extends MessageInterface
      *     provided status code; if none is provided, implementations MAY
      *     use the defaults as suggested in the HTTP specification.
      * @return self
-     * @throws \InvalidArgumentException for invalid status code arguments.
+     * @throws \InvalidArgumentException For invalid status code arguments.
      */
     public function withStatus($code, $reasonPhrase = '');
 
@@ -1638,7 +1638,7 @@ interface UriInterface
      *
      * An empty host value is equivalent to removing the host.
      *
-     * @param string $host The host to use with the new instance.
+     * @param string $host The hostname to use with the new instance.
      * @return self A new instance with the specified host.
      * @throws \InvalidArgumentException for transformations that would result
      *     in a state that cannot be represented as a valid URI reference.
--- a/accepted/PSR-7-http-message.md
+++ b/accepted/PSR-7-http-message.md
@@ -231,7 +231,7 @@ of the request line. The request target can be one of the following forms:
 Aside from these request-targets, there is often an 'effective URL' which is
 separate from the request target. The effective URL is not transmitted within
 an HTTP message, but it is used to determine the protocol (http/https), port
-and hostname for making the request.
+and host for making the request.
 
 The effective URL is represented by `UriInterface`. `UriInterface` models HTTP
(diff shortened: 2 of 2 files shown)


accepted/PSR-7-http-message.md
```diff
@@ -1232,7 +1232,7 @@ interface ResponseInterface extends MessageInterface
      *     provided status code; if none is provided, implementations MAY
      *     use the defaults as suggested in the HTTP specification.
      * @return self
-     * @throws \InvalidArgumentException For invalid status code arguments.
+     * @throws \InvalidArgumentException for invalid status code arguments.
```
Tobion: consistency

accepted/PSR-7-http-message.md
```diff
@@ -1606,8 +1606,9 @@ interface UriInterface
      *
      * @param string $scheme The scheme to use with the new instance.
      * @return self A new instance with the specified scheme.
-     * @throws \InvalidArgumentException for invalid schemes.
```
Tobion: This is covered by the more generic wording I added. So no need for this anymore.

accepted/PSR-7-http-message.md
```diff
@@ -1697,7 +1704,6 @@ interface UriInterface
      *
      * @param string $query The query string to use with the new instance.
      * @return self A new instance with the specified query string.
-     * @throws \InvalidArgumentException for invalid query strings.
```
Tobion: there is no invalid query string possible and no invalid URI possible by modifying the query.

accepted/PSR-7-http-message.md
mnapoli: These changes be in a separate pull request IMO.

Tobion:
IMO this is good to go. Can we start a vote @weierophinney @nyamsprod ?

nyamsprod:
apart from my concern of a possible BC break with the path workaround removed I have no objection if @weierophinney does not share my concern then I'm okay with the errata as it stands

Tobion:
@michaelcullum can you move this forward with a vote as a secretary? @weierophinney seems to be busy.

michaelcullum:
@Tobion Sure. Would you mind writing up a summary paragraph for the rationale?

Tobion:
Thanks @michaelcullum. Here is the rationale:

The point of this errata is to clarify that the modifying methods of the URI components, e.g. `withHost($host)`,
should not only validate the individual URI component itself but the state of the URI. What is allowed for one URI
component often depends on the value of the other URI components (and those rules can even depend on the scheme that is used).

So what this errata does, is generalize the wording used to documented the exceptions. With this in place it also becomes clear
that the workaround we had in place for `__toString` (see diff) is not needed at all. This workaround was only there because exceptions cannot
be thrown in `__toString`. But since it makes much more sense to already throw exceptions in the `with*` methods when you try to
modify a URI in an invalid way, we can ensure `__toString` will always return a valid URI reference.

There is no BC break because it has already been documented before that those exceptions can be thrown. It only clarifies the
circumstances under which exceptions can be thrown. At the end it's still left to the implementation how much validation is implemented as
RFC 3986 is quite complex. Different implementations have different level of validation. Validation creates overhead and as mentioned above
can also depend on the URI scheme which will even trigger rules from other RFCs.

weierophinney:
Yes, quite busy. I've reviewed the changes, and my conversations with @nyamsprod, and these make sense. I do think they'll lead to some WTF moments for users who want to iteratively create a URI, but those can be countered via documentation in implementing projects.

:+1:

michaelcullum:
@Tobion Can you please remove the spec changes from this PR

Tobion:
So when/how do we fix the spec then?

michaelcullum:
With the Errata alone.

Tobion:
If we do not remove the the irrelevant phpdoc from `__toString` then the spec is just confusingly wrong and not in sync with the errata section. So only adding the errata alone makes no sense.

pmjones:
@michaelcullum You locked a Github issue yesterday for "discussion" in its comments. I presume you'll be doing the same here? If not, why not?

michaelcullum:
@pmjones I was discussing what should be in this PR and what should be separate which is fine, but yes, @Tobion if you want to discuss it further about how it's necessary for the Errata then it will need to go on the mailing list.

Tobion:
@michaelcullum Can we make the changes to the spec without a vote as soon as the errata passed? Then I'm fine to remove it for now here.

judgej:
Just as a general note, if every change to a URL should result in a valid state for that URL, then there will be states that cannot exist, that you cannot get to without triggering an exception. So if moving from state A to state B, where two or more related (by some rule) properties need to be set, is there a possibility that you cannot get to that state without the need to pass through an invalid state on the way?

I'm asking this question not only theoretically, but also practically. Going via a zig-zag path between states to avoid the invalid states is something that may have to be done. But how often could that be a problem?

Tobion:
> So if moving from state A to state B, where two or more related (by some rule) properties need to be set, is there a possibility that you cannot get to that state without the need to pass through an invalid state on the way?

You can always reach any valid state from any given valid state. That is not a problem theoretically. And practically the only thing that might surprise people is that you have to set the `host` before setting the `port`:
-  `(new Uri(''))->withHost('example.org')->withPort(8080)` is valid but 
- `(new Uri(''))->withPort(8080)->withHost('example.org')` should fail because at this point `(new Uri(''))->withPort(8080)` the URI would not make sense. A port without a host is not valid. And this removes ambiguity. If there was not exception triggered, then different implementations could have different behavior:
- one might ignore the port `(new Uri(''))->withPort(8080)->getPort() === null` which is awkward
- another might set the port but ingore it in `__toString` which would make the URI state inconsistent: `(new Uri(''))->withPort(8080)->getPort() === 8080` but `(string) (new Uri(''))->withPort(8080) === ''`

judgej:
Thanks @Tobion for those examples. It makes sense when constructing a URL if state `A` is an empty string, so you are building towards the final URL with a series of refinements. Moving from `A` to `B` will likely always involve changing or removing refinements, and you point out, there is little chance of needing to go though any of the [rare] invalid states.

accepted/PSR-7-http-message-meta.md
```diff
@@ -647,3 +647,21 @@ used to populate the headers of an HTTP message.
...
+URI reference via the `UriInterface::__toString` method. This means that the
+modifying methods of the URI components, e.g. `withHost($host)`, have to ensure
+this is still the case. If a modification would result in an invalid URI, the
+methods MUST throw an `\InvalidArgumentException`. Invalid URIs can not only be
+caused by invalid characters in a URI component but also because the given
+URI components cannot be recomposed to a valid URI reference which may even
+depend on the current URI scheme. Two examples are:
+
+- `$uri->withHost('example.org')->withPath('//two-slashes')->withHost('')` should
+  fail because the path of a URI without an authority must not start with two slashes.
```
weierophinney: No, this should fail because *there is no host*, and neither HTTP nor HTTPS schemes allow for an empty host.

Tobion: I don't see where you are going. This already says it should fail but for a different reason.

judgej: @weierophinney does that example even have a scheme? I'm not sure it does.

weierophinney: @judgej — PSR-7 specification indicates that implementations should default to an HTTP scheme, so, yes, it does.

@Tobion — mis-read, based on what I was reading on the list; you can disregard.

Tobion: > PSR-7 specification indicates that implementations should default to an HTTP scheme, so, yes, it does.

What? Where does it say so? This wouldn't make sense because then a network path cannot be represented like `//example.com/path`.

accepted/PSR-7-http-message.md
```diff
@@ -1637,7 +1640,8 @@ interface UriInterface
      *
      * @param string $host The hostname to use with the new instance.
      * @return self A new instance with the specified host.
-     * @throws \InvalidArgumentException for invalid hostnames.
+     * @throws \InvalidArgumentException for transformations that would result
```
nyamsprod: You can not change the wording of the docblock in an accepted PSR this is against the bylaws. So the text for the InvalidArgumentException must be reverted.

accepted/PSR-7-http-message.md
```diff
@@ -1727,13 +1736,7 @@ interface UriInterface
...
      * - If a scheme is present, it MUST be suffixed by ":".
      * - If an authority is present, it MUST be prefixed by "//".
-     * - The path can be concatenated without delimiters. But there are two
-     *   cases where the path has to be adjusted to make the URI reference
-     *   valid as PHP does not allow to throw an exception in __toString():
-     *     - If the path is rootless and an authority is present, the path MUST
-     *       be prefixed by "/".
-     *     - If the path is starting with more than one "/" and no authority is
-     *       present, the starting slashes MUST be reduced to one.
+     * - The path can be concatenated without delimiters.
```
nyamsprod: You can not change the wording of the docblock in an accepted PSR this is against the bylaws. this text needs to be reverted **BUT WE NEED TO EMPHASIZE** in the errata that this is useless with the apply errata.

Tobion:
So do we keep the PSR-7 spec broken and not reflect actual implementations?

iansltx:
@Tobion mind de-conflicting this?

Tobion:
Would it get merged?

Jean85:
You can try to bring it up to the mailing list if you want to push it forward.

Tobion:
I've already brought it up on the mailing list before and people agreed we should do it. But as long as people argue PSRs cannot evolve and get fixed in tiny ways, this is just a dead end. So we have to live with outdated PSRs. There is also the related discussion in https://groups.google.com/forum/#!topic/php-fig/OyC3plRYhqg/discussion
I'm closing this for now until there is a fig process that actually allows such changes.

PR #757, opened 2016-05-20
Convert ticks to tildes.

reinink:
This really is some of my better work. The new GitHub pages uses the rouge highlighter now, which doesn't support ticks for code blocks. We're getting build warnings on the website, and this is the first step to fixing this. Learn more here: https://github.com/php-fig/php-fig.github.com/issues/191

I've done a quick scan through, and I don't see any issues caused by this update. There will likely be some more tweaks yet, but I'll handle them as separate pull requests.

Thanks! 🚤


reinink:
Honest to goodness, I'm not sure how there could be conflicts, I JUST forked this.

PR #758, opened 2016-05-20
Convert ticks to tildes.

reinink:
Okay let's try this again. Evidently I don't know how to Git, and had an old copy of this repo around when I made these changes.

The new GitHub pages uses the rouge highlighter now, which doesn't support ticks for code blocks. We're getting build warnings on the website, and this is the first step to fixing this. Learn more here: https://github.com/php-fig/php-fig.github.com/issues/191

I've done a quick scan through, and I don't see any issues caused by this update. There will likely be some more tweaks yet, but I'll handle them as separate pull requests.

Thanks! :feelsgood:

### DIFF
--- a/accepted/PSR-0.md
+++ b/accepted/PSR-0.md
@@ -1,7 +1,7 @@
 Autoloading Standard
 ====================
 
-> **Deprecated** - As of 2014-10-21 PSR-0 has been marked as deprecated. [PSR-4] is now recommended 
+> **Deprecated** - As of 2014-10-21 PSR-0 has been marked as deprecated. [PSR-4] is now recommended
 as an alternative.
 
 [PSR-4]: http://www.php-fig.org/psr/psr-4/
@@ -51,7 +51,7 @@ Example Implementation
 Below is an example function to simply demonstrate how the above
 proposed standards are autoloaded.
 
-```php
+~~~php
 <?php
 
 function autoload($className)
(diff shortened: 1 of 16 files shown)


svpernova09:
Looks good from here 👍

samdark:
Is website OK with it?

svpernova09:
I'm unable to test this locally, I'm stuck in Ruby hell. I can't bundle install because I'm missing json 1.8.1, and I can't install that version because it isn't compatible with Ruby 2.3.x. Tried updating the gemfile to use json 1.8.3 and that fixed bundle install complaining however it isn't actually building the site.

reinink:
@samdark Yes, I have tested with the website, that is what kicked this off. Unless you're wondering if the website _team_ is okay with it, then, yes as well. ;)

reinink:
And for what it's worth, pulling this in won't actually affect the website immediately. That will happen when we update the submodule on the website.

michaelcullum:
Thanks for this @reinink! Your help on this kind of thing is always most welcome and appreciated.

Feel free to pull the website whenever is appropriate. :)

reinink:
You're quite welcome @michaelcullum! I may fix a few more things before I update the website. For instance, the new default Markdown parser (kramdown) has table support, which is nice, but if your table has no rows and only headers, the table doesn't render. And this will effect a few areas, ie:

image

I'm working on a solution. 🌞

PR #759, opened 2016-05-24
Create initial proposal for HTTP factories

shadowhand:
Based on discussions #755 and https://groups.google.com/forum/#!topic/php-fig/G5pgQfQ9fpA

### DIFF
--- a/proposed/http-factory/http-factory-meta.md
+++ b/proposed/http-factory/http-factory-meta.md
@@ -1,0 +1,108 @@
+PSR-N Meta Document
+===================
+
+1. Summary
+----------
+
+The purpose of this PSR is to provide factory interfaces that define methods to
+create PSR-7 objects.
+
+2. Why Bother?
+--------------
+
+The current specification for PSR-7 allows for most objects to be modified by
+creating immutable copies. However, there are two notable exceptions:
+
+* `StreamInterface` is a mutable object based on a resource that only allows
+  the resource to be written to when the resource is writable.
+* `UploadedFileInterface` is a read-only object based on a resource that offers
+  no modification capabilities.
+
+The former is a significant pain point for PSR-7 middleware, as it can leave
+the response in an incomplete state. If the stream attached to the response body
+is not seekable or not writable, there is no way to recover from an error
+condition in which the body has already been written too.
+
+This scenario can be avoided by providing a factory to create new streams. Due to
+the lack of formal standard for HTTP object factories, a developer must rely on
+a specific vendor implementation in order to create these objects. Creating a
+formal standard for factories will allow for developers to avoid dependency on
+specific implementations while having the ability to create new objects when
+necessary.
+
+3. Scope
+--------
+
+## 3.1 Goals
+
+* Provide a set of interfaces that define methods to create PSR-7 compatible objects.
+
+## 3.2 Non-Goals
+
+* Provide a specific implementation of PSR-7 factories.
+
+4. Approaches
+-------------
+
+### 4.1 Chosen Approach
+
+The factory method definition has been chosen based on whether or not the object
+can be modified after instantiation. For interfaces that cannot be modified, all
+of the object properties must be defined at the time of instantiation.
+
+In the case of `UriInterface` a complete URI may be passed for convenience.
+
+The method names used will not conflict. This allows for a single class to
+implement multiple interfaces when appropriate.
+
+#### 4.2 Existing Implementations
+
+All of the current implementations of PSR-7 have defined their own requirements.
+In most cases, these requirements are the same or very similar.
+
+#### 4.3 Potential Issues
+
+The most difficult part of defining the method signatures for the interfaces.
+As there is no clear declaration in PSR-7 as to what values are explicitly
+required, the properties that are read only must be inferred based on whether
+the interfaces have methods to copy-and-modify the object.
+
+5. People
+---------
+
+### 5.1 Editor(s)
+
+* Woody Gilk, <woody.gilk@gmail.com>
+
+### 5.2 Sponsors
+
+* Paul M Jones, <pmjones88@gmail.com> (Coordinator)
+* ?
+
+### 5.3 Contributors
+
+* Rasmus Schultz, <rasmus@mindplay.dk>
+
+6. Votes
+--------
+
+* **Entrance Vote:**  _(not yet taken)_
+* **Acceptance Vote:** _(not yet taken)_
+
+7. Relevant Links
+-----------------
+
+_**Note:** Order descending chronologically._
+
+* [PSR-7 Middleware Proposal](https://github.com/php-fig/fig-standards/pull/755)
+* [PHP-FIG mailing list discussion of middleware](https://groups.google.com/forum/#!topic/php-fig/vTtGxdIuBX8)
+* [ircmaxwell All About Middleware](http://blog.ircmaxell.com/2016/05/all-about-middleware.html)
+* [shadowhand All About PSR-7 Middleware](http://shadowhand.me/all-about-psr-7-middleware/)
+* [AndrewCarterUK PSR-7 Objects Are Not Immutable](http://andrewcarteruk.github.io/programming/2016/05/22/psr-7-is-not-immutable.html)
+* [shadowhand Dependency Inversion and PSR-7 Bodies](http://shadowhand.me/dependency-inversion-and-psr-7-bodies/)
+* [PHP-FIG mailing list thread about factories](https://groups.google.com/forum/#!topic/php-fig/G5pgQfQ9fpA)
+
+8. Errata
+---------
+
+...
--- a/proposed/http-factory/http-factory-meta.md
+++ b/proposed/http-factory/http-factory-meta.md
@@ -1,5 +1,5 @@
-PSR-N Meta Document
-===================
+PSR-17 Meta Document
+====================
 
 1. Summary
 ----------
--- a/proposed/http-factory/http-factory-meta.md
+++ b/proposed/http-factory/http-factory-meta.md
@@ -86,7 +86,7 @@ the interfaces have methods to copy-and-modify the object.
 6. Votes
 --------
 
-* **Entrance Vote:**  _(not yet taken)_
+* [Entrance Vote](https://groups.google.com/forum/#!topic/php-fig/6rZPZ8VglIM)
 * **Acceptance Vote:** _(not yet taken)_
 
 7. Relevant Links
--- a/proposed/http-factory/http-factory-meta.md
+++ b/proposed/http-factory/http-factory-meta.md
@@ -100,7 +100,8 @@ _**Note:** Order descending chronologically._
 * [shadowhand All About PSR-7 Middleware](http://shadowhand.me/all-about-psr-7-middleware/)
 * [AndrewCarterUK PSR-7 Objects Are Not Immutable](http://andrewcarteruk.github.io/programming/2016/05/22/psr-7-is-not-immutable.html)
 * [shadowhand Dependency Inversion and PSR-7 Bodies](http://shadowhand.me/dependency-inversion-and-psr-7-bodies/)
-* [PHP-FIG mailing list thread about factories](https://groups.google.com/forum/#!topic/php-fig/G5pgQfQ9fpA)
+* [PHP-FIG mailing list thread discussing factories](https://groups.google.com/d/msg/php-fig/G5pgQfQ9fpA/UWeM1gm1CwAJ)
+* [PHP-FIG mailing list thread feedback on proposal](https://groups.google.com/d/msg/php-fig/piRtB2Z-AZs/8UIwY1RtDgAJ)
 
 8. Errata
 ---------
--- a/proposed/http-factory/http-factory-meta.md
+++ b/proposed/http-factory/http-factory-meta.md
@@ -76,8 +76,8 @@ the interfaces have methods to copy-and-modify the object.
 
 ### 5.2 Sponsors
 
-* Paul M Jones, <pmjones88@gmail.com> (Coordinator)
-* ?
+* Roman Tsjupa, <draconyster@gmail.com> (Coordinator)
+* Paul M Jones, <pmjones88@gmail.com>
 
 ### 5.3 Contributors
 
--- a/proposed/http-factory/http-factory.md
+++ b/proposed/http-factory/http-factory.md
@@ -1,0 +1,183 @@
+HTTP Factories
+==============
+
+This document describes a common standard for factories that create [PSR-7][psr7]
+compliant HTTP objects.
+
+PSR-7 did not include a recommendation on how to create HTTP objects, which leads
+to difficulty when needing to create new HTTP objects within components that are
+not tied to a specific implementation of PSR-7.
+
+The interfaces described in this document describe methods by which PSR-7 objects
+can be instantiated.
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in [RFC 2119][rfc2119].
+
+[psr7]: http://www.php-fig.org/psr/psr-7/
+[rfc2119]: http://tools.ietf.org/html/rfc2119
+
+## 1. Specification
+
+An HTTP factory is a method by which a new HTTP object, as defined by PSR-7,
+is created. HTTP factories MUST implement these interfaces for each object type
+that is implemented.
+
+## 2. Interfaces
+
+The following interfaces MAY be implemented together within a single class or
+in separate classes.
+
+### 2.1 RequestFactoryInterface
+
+```php
+namespace Psr\Http\Message;
+
+interface RequestFactoryInterface
+{
+    /**
+     * Create a new request.
+     *
+     * @return RequestInterface
+     */
+    public function createRequest();
+}
+```
+
+### 2.2 ResponseFactoryInterface
+
+```php
+namespace Psr\Http\Message;
+
+interface ResponseFactoryInterface
+{
+    /**
+     * Create a new response.
+     *
+     * @return ResponseInterface
+     */
+    public function createResponse();
+}
+```
+
+### 2.3 ServerRequestFactoryInterface
+
+```php
+namespace Psr\Http\Message;
+
+interface ServerRequestFactoryInterface
+{
+    /**
+     * Create a new server request.
+     *
+     * @return ServerRequestInterface
+     */
+    public function createServerRequest();
+}
+```
+
+### 2.4 StreamFactoryInterface
+
+```php
+namespace Psr\Http\Message;
+
+interface StreamFactoryInterface
+{
+    /**
+     * Create a new stream.
+     *
+     * If a string is used to create the stream, a temporary resource will be
+     * created with the content of the string. The resource will be writable
+     * and seekable.
+     *
+     * If a resource is passed it must be readable.
+     *
+     * @param string|resource $body
+     *
+     * @return StreamInterface
+     *
+     * @throws \InvalidArgumentException
+     *  If a passed resource is not readable.
+     */
+    public function createStream($body = '');
+}
+```
+
+Implementations of this interface SHOULD use a temporary file when creating
+resources from strings. The RECOMMENDED method for doing so is:
+
+```php
+$resource = fopen('php://temp', 'r+');
+fwrite($resource, $body);
+```
+
+### 2.5 UploadedFileFactoryInterface
+
+```php
+namespace Psr\Http\Message;
+
+interface UploadedFileFactoryInterface
+{
+    /**
+     * Create a new uploaded file.
+     *
+     * If a string is used to create the file, a temporary resource will be
+     * created with the content of the string.
+     *
+     * If a size is not provided it will be determined by checking the size of
+     * the file.
+     *
+     * @see http://php.net/manual/features.file-upload.post-method.php
+     * @see http://php.net/manual/features.file-upload.errors.php
+     *
+     * @param string|resource $file
+     * @param integer $size in bytes
+     * @param integer $error PHP file upload error
+     * @param string $clientFilename
+     * @param string $clientMediaType
+     *
+     * @return UploadedFileInterface
+     *
+     * @throws \InvalidArgumentException
+     *  If the file resource is not readable.
+     */
+    public function createUploadedFile(
+        $file,
+        $size = null,
+        $error = \UPLOAD_ERR_OK,
+        $clientFilename = null,
+        $clientMediaType = null
+    );
+}
+```
+
+Implementations of this interface SHOULD use a temporary file when creating
+resources from strings. The RECOMMENDED method for doing so is:
+
+```php
+$resource = fopen('php://temp', 'r+');
+fwrite($resource, $body);
+```
+
+### 2.6 UriFactoryInterface
+
+```php
+namespace Psr\Http\Message;
+
+interface UriFactoryInterface
+{
+    /**
+     * Create a new URI.
+     *
+     * @param string $uri
+     *
+     * @return UriInterface
+     *
+     * @throws \InvalidArgumentException
+     *  If the given URI cannot be parsed.
+     */
+    public function createUri($uri = '');
+}
+```
+
--- a/proposed/http-factory/http-factory.md
+++ b/proposed/http-factory/http-factory.md
@@ -133,9 +133,6 @@ interface StreamFactoryInterface
      * @param resource $body
      *
      * @return StreamInterface
-     *
-     * @throws \InvalidArgumentException
-     *  If a passed resource is not readable.
      */
     public function createStreamFromResource($body);
 
@@ -173,8 +170,7 @@ interface UploadedFileFactoryInterface
     /**
      * Create a new uploaded file.
      *
-     * If a string is used to create the file, a temporary resource will be
-     * created with the content of the string.
+     * If a string is passed it is assumed to be a file path.
      *
      * If a size is not provided it will be determined by checking the size of
      * the file.
@@ -189,9 +185,6 @@ interface UploadedFileFactoryInterface
      * @param string $clientMediaType
      *
      * @return UploadedFileInterface
-     *
-     * @throws \InvalidArgumentException
-     *  If the file resource is not readable.
      */
     public function createUploadedFile(
         $file,
--- a/proposed/http-factory/http-factory.md
+++ b/proposed/http-factory/http-factory.md
@@ -22,7 +22,7 @@ interpreted as described in [RFC 2119][rfc2119].
 
 An HTTP factory is a method by which a new HTTP object, as defined by PSR-7,
 is created. HTTP factories MUST implement these interfaces for each object type
-that is implemented.
+that is provided by the package.
 
 ## 2. Interfaces
 
@@ -31,10 +31,15 @@ in separate classes.
 
 ### 2.1 RequestFactoryInterface
 
+Has the ability to create complete client requests, including the request URI
+and body stream.
+
 ```php
 namespace Psr\Http\Message;
 
-interface RequestFactoryInterface
+interface RequestFactoryInterface extends
+    UriFactoryInterface,
+    StreamFactoryInterface
 {
     /**
      * Create a new request.
@@ -47,10 +52,13 @@ interface RequestFactoryInterface
 
 ### 2.2 ResponseFactoryInterface
 
+Has the ability to create complete responses, including the body stream.
+
 ```php
 namespace Psr\Http\Message;
 
-interface ResponseFactoryInterface
+interface ResponseFactoryInterface extends
+    StreamFactoryInterface
 {
     /**
      * Create a new response.
@@ -63,10 +71,16 @@ interface ResponseFactoryInterface
 
 ### 2.3 ServerRequestFactoryInterface
 
+Has the ability to create complete server requests, including the request URI,
+uploaded files, and body stream.
+
 ```php
 namespace Psr\Http\Message;
 
-interface ServerRequestFactoryInterface
+interface ServerRequestFactoryInterface extends
+    UploadedFileFactoryInterface,
+    UriFactoryInterface,
+    StreamFactoryInterface
 {
     /**
      * Create a new server request.
@@ -78,6 +92,8 @@ interface ServerRequestFactoryInterface
 ```
 
 ### 2.4 StreamFactoryInterface
+
+Has the ability to create streams for requests and responses.
 
 ```php
 namespace Psr\Http\Message;
@@ -113,6 +129,8 @@ fwrite($resource, $body);
 ```
 
 ### 2.5 UploadedFileFactoryInterface
+
+Has the ability to create streams for uploaded files.
 
 ```php
 namespace Psr\Http\Message;
@@ -161,6 +179,8 @@ fwrite($resource, $body);
 ```
 
 ### 2.6 UriFactoryInterface
+
+Has the ability to creates URIs for client and server requests.
 
 ```php
 namespace Psr\Http\Message;
--- a/proposed/http-factory/http-factory.md
+++ b/proposed/http-factory/http-factory.md
@@ -44,9 +44,12 @@ interface RequestFactoryInterface extends
     /**
      * Create a new request.
      *
+     * @param string $method
+     * @param UriInterface|string $uri
+     *
      * @return RequestInterface
      */
-    public function createRequest();
+    public function createRequest($method, $uri);
 }
 ```
 
@@ -63,9 +66,11 @@ interface ResponseFactoryInterface extends
     /**
      * Create a new response.
      *
+     * @param integer $code HTTP status code
+     *
      * @return ResponseInterface
      */
-    public function createResponse();
+    public function createResponse($code = 200);
 }
 ```
 
@@ -85,9 +90,19 @@ interface ServerRequestFactoryInterface extends
     /**
      * Create a new server request.
      *
+     * @param string $method
+     * @param UriInterface|string $uri
+     *
      * @return ServerRequestInterface
      */
-    public function createServerRequest();
+    public function createServerRequest($method, $uri);
+
+    /**
+     * Create a new server request from PHP globals.
+     *
+     * @return ServerRequestInterface
+     */
+    public function createServerRequestFromGlobals();
 }
 ```
 
--- a/proposed/http-factory/http-factory.md
+++ b/proposed/http-factory/http-factory.md
@@ -108,22 +108,48 @@ namespace Psr\Http\Message;
 interface StreamFactoryInterface
 {
     /**
-     * Create a new stream.
+     * Create a new stream with no content.
      *
-     * If a string is used to create the stream, a temporary resource will be
-     * created with the content of the string. The resource will be writable
-     * and seekable.
+     * The stream will be writable and seekable.
      *
-     * If a resource is passed it must be readable.
+     * @return StreamInterface
+     */
+    public function createStream();
+
+    /**
+     * Create a new stream from a callback.
      *
-     * @param string|resource $body
+     * The stream will be read-only and not seekable.
+     *
+     * @param callable $callback
+     *
+     * @return StreamInterface
+     */
+    public function createStreamFromCallback(callable $callback);
+
+    /**
+     * Create a new stream from a resource.
+     *
+     * @param resource $body
      *
      * @return StreamInterface
      *
      * @throws \InvalidArgumentException
      *  If a passed resource is not readable.
      */
-    public function createStream($body = '');
+    public function createStreamFromResource($body);
+
+    /**
+     * Create a new stream from a string.
+     *
+     * A temporary resource will be created with the content of the string.
+     * The resource will be writable and seekable.
+     *
+     * @param string $body
+     *
+     * @return StreamInterface
+     */
+    public function createStreamFromString($body);
 }
 ```
 
--- a/proposed/http-factory/http-factory.md
+++ b/proposed/http-factory/http-factory.md
@@ -31,15 +31,12 @@ in separate classes.
 
 ### 2.1 RequestFactoryInterface
 
-Has the ability to create complete client requests, including the request URI
-and body stream.
+Has the ability to create client requests.
 
 ```php
 namespace Psr\Http\Message;
 
-interface RequestFactoryInterface extends
-    UriFactoryInterface,
-    StreamFactoryInterface
+interface RequestFactoryInterface
 {
     /**
      * Create a new request.
@@ -55,13 +52,12 @@ interface RequestFactoryInterface extends
 
 ### 2.2 ResponseFactoryInterface
 
-Has the ability to create complete responses, including the body stream.
+Has the ability to create responses.
 
 ```php
 namespace Psr\Http\Message;
 
-interface ResponseFactoryInterface extends
-    StreamFactoryInterface
+interface ResponseFactoryInterface
 {
     /**
      * Create a new response.
@@ -76,16 +72,12 @@ interface ResponseFactoryInterface extends
 
 ### 2.3 ServerRequestFactoryInterface
 
-Has the ability to create complete server requests, including the request URI,
-uploaded files, and body stream.
+Has the ability to create server requests.
 
 ```php
 namespace Psr\Http\Message;
 
-interface ServerRequestFactoryInterface extends
-    UploadedFileFactoryInterface,
-    UriFactoryInterface,
-    StreamFactoryInterface
+interface ServerRequestFactoryInterface
 {
     /**
      * Create a new server request.
@@ -215,4 +207,3 @@ interface UriFactoryInterface
     public function createUri($uri = '');
 }
 ```
-


pmjones:
Nice. If/when the time comes, I'm happy to act as a coordinator or sponsor. (I suggest using `new` instead of `create`, e.g. `newServerRequest()`, but it's not a deal-breaker.)

shadowhand:
@pmjones I thought you might say that. :) I also consider `make` as a prefix. Do you have any stats on what is currently used in the wild for factory method prefixing? I know Aura et al tend towards `new` but I am not sure what others do.

pmjones:
I have no stats. My preference for `new*` comes from the [`ReflectionClass::newInstance()`](http://php.net/manual/en/reflectionclass.newinstance.php)

cundd:
I don't think that my opinion matters, but a thought on the prefix topic: I would prefer `create` over `new`, simply because it's a verb.

dbu:
good to see movement on this again. the last attempt failed to get any traction, and people ended up creating such factories in the [php-http](http://php-http.readthedocs.io) organization: 

https://github.com/php-http/message-factory

quite some thought went into those interfaces, and there exist adapters for the commonly used http clients. i think it would be a good idea to compare this proposal with those interfaces and see if we can spot any issues with the interfaces proposed here in the standard.

proposed/http-factory/http-factory.md
```diff
@@ -0,0 +1,152 @@
...
+{
+    /**
+     * Create a new request.
+     *
+     * @param string|UriInterface $uri
+     * @param string $method
+     *
+     * @return RequestInterface
+     */
+    public function createRequest($uri = null, $method = null);
```
dbu: https://github.com/php-http/message-factory/blob/master/src/RequestFactory.php#L27-L32 allows to specify headers, body and the http version. not sure about the version, the rest can of course be added after the request is created, but it would be more convenient for the user.

proposed/http-factory/http-factory.md
```diff
@@ -0,0 +1,152 @@
...
+interface ResponseFactoryInterface
+{
+    /**
+     * Create a new response.
+     *
+     * @param string|StreamInterface $body
+     *
+     * @return ResponseInterface
+     */
+    public function createResponse($body = null);
```
dbu: https://github.com/php-http/message-factory/blob/master/src/ResponseFactory.php#L28-L33 has more fields: status, reason, headers, and the protocol version.

proposed/http-factory/http-factory.md
```diff
@@ -0,0 +1,152 @@
...
+
+```php
+namespace Psr\Http\Message;
+
+interface StreamFactoryInterface
+{
+    /**
+     * Create a new stream.
+     *
+     * @param string|resource $body
```
dbu: could body also be a StreamInterface already? can i create an empty stream by passing null?

proposed/http-factory/http-factory.md
```diff
@@ -0,0 +1,152 @@
...
+     * @param string $mode
```
dbu: there should be documentation what valid modes are and what they mean

proposed/http-factory/http-factory.md
```diff
@@ -0,0 +1,152 @@
...
+{
+    /**
+     * Create a new server request.
+     *
+     * @param string|UriInterface $uri
+     * @param string $method
+     *
+     * @return ServerRequestInterface
+     */
+    public function createServerRequest($uri = null, $method = null);
```
dbu: same as with RequestFactory, could have more arguments for convenience.

proposed/http-factory/http-factory.md
shadowhand: Correct, but all of those can be modified after the response has been created.

proposed/http-factory/http-factory.md
shadowhand: I think it probably makes more sense for this to be:

```php
createRequest($body = null)
```

Since everything else can be modified after creation.

proposed/http-factory/http-factory.md
shadowhand: What's the use case for creating a stream from a stream? Wouldn't it result in the same underlying resource being used, thereby not _really_ creating a new stream?

proposed/http-factory/http-factory.md
shadowhand: More arguments are not typically more convenient. Using the methods provides better self-documentation and is ultimately easier to work with.

shadowhand:
@dbu the documentation on those classes is nice. I'm not at all convinced that the method signatures are generally better, as too many arguments makes calling code confusing.

sagikazarmark:
> as too many arguments makes calling code confusing.

That I can agree with. But I think it's a bit weird that you create a request with a body and a uri and you have to manually set the method and probably some headers afterwards. I think this is rather confusing.

proposed/http-factory/http-factory.md
sagikazarmark: I think we did this for convenience:

https://github.com/php-http/message/blob/master/src/MessageFactory/DiactorosMessageFactory.php#L40

However I think that's a design flaw, because as @shadowhand point: it's not a new stream, BUT it does rewind the stream:

https://github.com/php-http/message/blob/master/src/StreamFactory/DiactorosStreamFactory.php#L35

This is probably not a good idea either, but we faced the problem that during stream creation, we have to actually rewind it.

proposed/http-factory/http-factory.md
sagikazarmark: Although those can be modified after creation, I think it is confusing. (I think we also had that minimalistic interface at the beginning for the same reasons, but we added those arguments for convenience and they work quite well).

I agree, that version might not be required, we added it to fully cover an HTTP message.

dbu: thinking about this some more, i actually quite like the idea of writing:

```php
$request = $requestFactory->createRequest('/foobar')
    ->withMethod('POST')
    ->withProtocolVersion('1.0')
    ->withHeader('accept', 'application/json')
    ->withBody($streamFactory->...)
;
```

we end up with something like a query builder. 
i guess a bunch of CPU cycles to instantiate a couple of objects we only use to create another object are negligible compared to a HTTP request execution roundtrip. i hope the performance question of this was investigated with PSR-7 already...

proposed/http-factory/http-factory.md
cundd: And/or constants for those modes?

proposed/http-factory/http-factory.md
sagikazarmark: Hm, I didn't think about that. That actually looks pretty cool. Although: `$streamFactory->...` That adds an extra dependency to your code.

dbu: yeah from a convenience point of view, RequestFactory could extend StreamFactory. but i guess that is really not elegant design wise.

sagikazarmark: I don't mean extension. I think right now we just use the stream factories within message factories. That is a dependency as well, but only within the message factory, which depends on the message implementation anyway, so from my point of view, it's better than adding dependency in your code.

proposed/http-factory/http-factory.md
```diff
@@ -0,0 +1,172 @@
...
+     * If a resource is passed it must be readable.
+     *
+     * @param string|resource $body
+     *
+     * @return StreamInterface
+     *
+     * @throws \InvalidArgumentException
+     *  If a passed resource is not readable.
+     */
+    public function createStream($body = '');
```
shadowhand: This method no longer contains a `$mode` parameter, which was unnecessary in the first place, since the `$body` could not represent a file.

proposed/http-factory/http-factory.md
```diff
@@ -0,0 +1,172 @@
...
+}
+```
+
+Implementations of this interface SHOULD use a temporary file when creating
+resources from strings. The RECOMMENDED method for doing so is:
+
+```php
+$resource = fopen('php://temp', 'r+');
+fwrite($resource, $body);
```
shadowhand: Added a recommendation for how to create resources from strings.

proposed/http-factory/http-factory.md
```diff
@@ -0,0 +1,189 @@
...
+     *
+     * @return UploadedFileInterface
+     *
+     * @throws \InvalidArgumentException
+     *  If the file resource is not readable.
+     */
+    public function createUploadedFile(
+        $file,
+        $size = null,
+        $error = \UPLOAD_ERR_OK,
```
shadowhand: Added defaults for `$size` and `$error` for uploaded files.

proposed/http-factory/http-factory-meta.md
```diff
@@ -0,0 +1,105 @@
...
+5. People
+---------
+
+### 5.1 Editor(s)
+
+* Woody Gilk, <woody.gilk@gmail.com>
+
+### 5.2 Sponsors
+
+* Paul M Jones, <pmjones88@gmail.com> (Coordinator)
```
shadowhand: @pmjones since neither of us learned out lesson with #755.

shadowhand:
@dbu @sagikazarmark I've updated a bunch of signatures and refined the arguments in an attempt to minimize them as much as possible and prevent any dependency between interfaces. This means that `createResponse` will no longer take a `$body`, as that would create a dependency on `StreamFactoryInterface`. This seems ideal to me, as it separates concerns as much as possible. 

(Server)Request still have `$uri` parameter though, which implies an implicit dependency on `UriFactoryInterface`. Perhaps it would make more sense for `createRequest($method = null)` to be the signature?

proposed/http-factory/http-factory-meta.md
pmjones: @shadowhand Not sure what you mean but pretty sure it'll be funny when I do. ;-)

dbu:
> (Server)Request still have |$uri| parameter though, which implies an
> implicit dependency on |UriFactoryInterface|. Perhaps it would make more
> sense for |createRequest($method = null)| to be the signature?

wouldn't the radical way be |createRequest()| if both method and uri can
be changed later?

shadowhand:
@dbu I just had the same thought and was about to say so. I'm going to remove all parameters that can be set after instantiation, because that's it is the most sane thing to do.

proposed/http-factory/http-factory.md
```diff
@@ -0,0 +1,183 @@
...
+namespace Psr\Http\Message;
+
+interface RequestFactoryInterface
+{
+    /**
+     * Create a new request.
+     *
+     * @return RequestInterface
+     */
+    public function createRequest();
```
shadowhand: Updated request/response signatures to have no parameters, since the objects can be modified-by-copy after creation.

proposed/http-factory/http-factory.md
```diff
@@ -0,0 +1,183 @@
...
+     * Create a new URI.
+     *
+     * @param string $uri
+     *
+     * @return UriInterface
+     *
+     * @throws \InvalidArgumentException
+     *  If the given URI cannot be parsed.
+     */
+    public function createUri($uri = null);
```
shadowhand: I think keeping the `$uri` parameter here makes sense, since there is no method in `UriInterface` to set a complete URI, only parts of it. This would be unpleasant:

```php
$uri = $this->uriFactory->createUri()
    ->withScheme('http')
    ->withHost('example.com')
    ->withPath('/');
```

Most implementations of `UriInterface` (Guzzle, Diactoros, etc) either allow the URI to be passed to the constructor, or provide some other factory to do so (League/URI). Slim is a [notable exception](https://github.com/slimphp/Slim/blob/88bc70973d6dde7620f1287b4d7681923e0e616a/Slim/Http/Uri.php#L112-L121) as it defines parameters for each part of the URI, rather than parsing it during construction.

JoeBengalen: Slim has a factory method for creating a Uri instance from string:
https://github.com/slimphp/Slim/blob/3.x/Slim/Http/Uri.php#L140

slavcodev:
I would prefer the builders instead of factories. The builders with similar API as PSR-7 interfaces.

```php
$uri = (new UriBuilder())
  ->withScheme('http')
  ->withHost('example.com')
  ->withPath('/')
  ->create()
;
$request = (new RequestBuilder())
  ->withMethod('GET')
  ->withUri($uri)
  ->withHeader([]);
  ->create()
;
```

That make more sense for me, unlike methods with only a part of objects parameters.

Implementation extending:

```php
$uri = (new MyUriBuilder())
  ->withPrototype($baseUri)
  ->create()
;
```

shadowhand:
@slavcodev I don't see how that offers anything over factories, since it would make the interfaces more complex and duplicate existing methods that already exist in PSR-7 interfaces.

slavcodev:
It won't duplicate. The builder is not immutable.

shadowhand:
@slavcodev it appears to be immutable, since it uses `with*` method prefixes. And you didn't answer my question: what benefit does this offer over the current factory proposal?

sagikazarmark:
I don't like the builder idea. And it is actually a duplicate, since we have this provided by the PSR-7 interfaces. It's immutable, but you can just use the exact same syntax to get the same result with a PSR-7 implementation, so it is actually duplication.

As for removing method arguments, I have two comments:

First of all it seems better to have either none or all the parameters configurable through the factory. But having some is actually confusing.

The other one:

> This means that createResponse will no longer take a $body, as that would create a dependency on StreamFactoryInterface. This seems ideal to me, as it separates concerns as much as possible.

Although it decouples the factories from each other, it is not really convenient this way. If you want to create a request, you need three factories now to be injected into your class. On the other hand, the message implementation already have those implementations (ideally), so why can't we depend on their factories in each other? (I mostly mean constructing a stream and a uri within a message factory). It doesn't seem so bad to me.

pmjones:
@sagikazarmark:

>  If you want to create a request, you need three factories now to be injected into your class

(/me ponders)

One could create a Factory class that implements the three interfaces.

sagikazarmark:
@pmjones That's still three dependency. You want to depend on the contract, not on an implementation and you need all the three. Creating a common interface would be really bad, so as depending on one implementation.

shadowhand:
> Although it decouples the factories from each other, it is not really convenient this way. If you want to create a request, you need three factories now to be injected into your class.

Only if you need to define a complete request. I really don't think this is so bad:

```php
public function __construct(
    RequestFactoryInterface $requestFactory,
    StreamFactoryInterface $streamFactory,
    UriFactoryInterface $uriFactory
) { ... }
```

shadowhand:
> First of all it seems better to have either none or all the parameters configurable through the factory. 

Some of the interfaces (UploadedFile, Stream) cannot be modified once created. All other interfaces have methods for modification. Hence, we have to provide arguments for these interfaces. For other interfaces there is no reason to choose some parameters over others. By including parameters that implicitly depend on other factories, it makes implementation of the factories unnecessarily complex and does not provide proper separation of concerns.

sagikazarmark:
Don't forget the HTTP Client and some kind of serializer in case of an API client. It can become bloated quite fast IMO.

sagikazarmark:
>  Some of the interfaces (UploadedFile, Stream) cannot be modified once created.

I should have said with the obvious exception of requirements, but I thought it was obvious. :smile:

shadowhand:
> I should have said with the obvious exception of requirements, but I thought it was obvious.

Which is to say, the current implementation is exactly what you are looking for: **all** required parameters are currently required, and **no** optional requirements are required.

sagikazarmark:
Exactly

shadowhand:
The main reason we cannot have interfaces that depend on other interfaces is it prevents partial PSR-7 implementations from providing factories. For instance, if I create an implementation of `RequestInterface` and want to provide a factory for it, I would either have to depend on a separate implementation of `UriInterface` or _also_ provide an implementation of `UriInterface`. Neither scenario is ideal to me.

sagikazarmark:
> The main reason we cannot have interfaces that depend on other interfaces is it prevents partial PSR-7 implementations from providing factories.

Fair enough. Although dependencies are between implementations or an implementation and a contract: for example a "complete" implementation of PSR-7 factories can depend on each other, but an implementation of a RequestFactory can also depend on the UriFactoryInterface. A "complete" implementation can also include stream construction WITHIN the message factory itself, but that's probably duplication. So in theory, there are no dependencies between the interfaces.

From a standard perspective: you are perfectly right. But it still remains inconvenient to use IMO.

slavcodev:
> it appears to be immutable, since it uses with* method prefixes

hmm, interesting. May be I'm missing something, why do you think that `with` prefix in method name, make the object immutable?

sagikazarmark:
It's commonly accepted AFAIK that `with*` assumes immutability. It would cause a huge confusion if not, since PSR-7 follows this pattern and it's immutable.

slavcodev:
> And you didn't answer my question: what benefit does this offer over the current factory proposal?

I think, I did.

- Factory which accepts only certain parameters to create an object, and setting the rest of the object by changing it, like 

```php
$request = $this->requestFactory->createRequest('GET')
    ->withUri($uri)
    ->withHeader(...);
```

make no sense, IMO.

- Builders are more extendable, e.q.

```
$response = (new ResponseBuilder())
    ->withJsonBody(...)
    ->create();
```

> It's commonly accepted AFAIK that with* assumes immutability.

Okay, sincerely first time I see this convention, well, feel free to use better names of method.

shadowhand:
@slavcodev look the previous discussion and the current proposed interfaces. Only objects that cannot be modified after creation include parameters. The argument about builders makes no sense in this context, because the PSR-7 interfaces (except stream and uploaded file) are essentially builders.

sagikazarmark:
I agree with @shadowhand. I don't see why builders can't be immutable. From a functional POV the only difference is that you don't have to call a create method and you don't have shared state. Both of them seem to be advantages for me.

slavcodev:
> look the previous discussion and the current proposed interfaces. Only objects that cannot be modified after creation include parameters.

Sweet. I've seen the older version, like `createRequest($uri = null, $method = null)`. I see, you've updated docs.

> The argument about builders makes no sense in this context, because the PSR-7 interfaces (except stream and uploaded file) are essentially builders.

Oh, this is the reason for my disagreement with the proposal. For me, PSR-7 describes interfaces of value objects, however builders are not. I understand the builders as services which takes given values and produce those data objects.

sagikazarmark:
> I understand the builders as services which takes given values and produce those data objects.

This is perfectly fine for PSR-7 objects as well: They take an argument (`->withMethod('GET')`) and return a data object (since they are immutable).

dracony:
I believe all these methods should also be annotated with proper `@throws`e,g the standard InvalidArgumentException in the definition. The PSR-7 already makes use of these, so this one probably should too.

dbu:
> The main reason we cannot have interfaces that depend on other
> interfaces is it prevents partial PSR-7 implementations from providing
> factories. For instance, if I create an implementation of
> |RequestInterface| and want to provide a factory for it, I would either
> have to depend on a separate implementation of |UriInterface| or /also/
> provide an implementation of |UriInterface|. Neither scenario is ideal
> to me.

the implementation of the RequestFactory can ask for an
UriFactoryInterface in its constructor. it would just provide the
composition in the interfaces instead of leaving that to the user who
either creates a composition class themselves or injects 3 arguments
instead of one.

unless i am mistaken, you can build neither a client nor a server
without a uri implementation.

> builders

to me that is a no-go. the point of the factories is that
implementations can depend on interfaces instead of concrete
implementations. the builders are concrete implementations so they would
need to know about all possible existing implementations plus need some
magic way to choose between them if more than one implementation is
available in a project.

> all these methods should also be annotated with proper |@throws|

+1
(though lets first sort out how exactly they look, to avoid wasted time.
the less arguments, the less exceptions)

dracony:
@shadowhand see my PR: https://github.com/shadowhand/fig-standards/pull/1

There must also be a general HttpFactoryInterface defined that extends all these interfaces. Otherwise classes will have to depend on all 6 Interface implementations being injected separately.

sagikazarmark:
> There must also be a general HttpFactoryInterface defined that extends all these interfaces.

This is a very bad idea. Not sure if you realized that I also brought up this issue, but having a common interface goes against every rules and is not a solution.

dracony:
Why is it a bad idea? I can bet you a horse and a small cat that we will never see an implementation that has these factories separate, because it doesn't make sense.

sagikazarmark:
Please read the PR discussion again to see why it's a bad idea. It has been mention quite a few times, although not always in this context.

dracony:
Already read it. I can't recall any design patterns book recommending creating a separate Factory per class. The idea of separating concerns between factories seems like fighting a windmill while frustrating users, who need to inject 6 classes to get things working.

Any middleware developer will hate such an interface from day 1. Not only do you need to accept 6 parameters, you also have to now store them in some private variables and write code that assigns those 6 private variables. Which, docblocks included is about 30 lines of totally redundant code.

sagikazarmark:
https://github.com/php-fig/fig-standards/pull/759#issuecomment-222079532

This comment summarizes the pros and cons of a proposed solution.

dracony:
Sooo? The RequestFactory that provides the composition could easily impement the combined HttpFactory Interface, by proxying method calls to proper factories. Where is the problem?

sagikazarmark:
In my opinion it is a bad idea, I think I explained why. I would like to hear other's opinion as well on this matter.

Also composition would be a perfect solution as in most cases you only need to create a message and not specifically a URI or a Stream. If you have to, you need to inject those as well.

Your solution would only work if a PSR-7 implementation is complete one, meaning provides implementation of all interfaces and factories.

dracony:
No, not really. Any library can still choose to implement only some of
these interfaces, and not implement the HttpRequestFactory. Where is the
problem in that? It's an addition to these interfaces, not a replacement.

shadowhand:
@dbu 
> the implementation of the RequestFactory can ask for an UriFactoryInterface in its constructor.

This could create an infinite recursion scenario if the same class implements multiple interfaces.

@dracony 
> There must also be a general HttpFactoryInterface defined that extends all these interfaces.

I don't believe there is a need for that. It would also make the situation less transparent due to loss of separation of concerns. Injecting 3 or even 6 copies of the same class (assuming one class implemented all the interfaces) is not really a problem. Considering that each HTTP object serves different needs, I don't see that as a problem.

sagikazarmark:
> This could create an infinite recursion scenario if the same class implements multiple interfaces.

Care to elaborate? I don't see the problem.

shadowhand:
@sagikazarmark ah, when I actually try to create code that reflects my thought process, it seems impossible. Strike that thought! 😀

dracony:
Injecting 6 copies of the same class looks clunky. Also middleware developers will have to track those 6 private variables. It is super inconvenient for the sake of very questionable architecture goal.

Also, I am yet to see a PSR-7 implementation that only provides factories for some interfaces. in 99.999% of the cases, users will use an implementation like zend-diactoros that provide the entire PSR-7 implementation. So this separation of concerns is catering to a very small minority (if anyone) with the cost of bad user experience for everyone, else. Why?

If you look at the actual packages and PSR-7 implementations, you'll see that nobody does it in this fragmented way. So maybe there is a reason to it? Sure, we can keep the fragmented interfaces, they're nice and all. But not providing a single aggregated factory interface is just being annoying to users.

dracony:
Imagine if you use Doctrine, you instead of injecting the EntityManager, had to inject the query builder and each repository separately. It might make sense in some cases, but the 99% case is the one that counts.

shadowhand:
@dracony 
> Injecting 6 copies of the same class looks clunky.

As I already stated, no middle (to my knowledge) would ever need to do this. Do you have an example of one that would need to do so?

> I am yet to see a PSR-7 implementation that only provides factories for some interfaces.

Did you miss where I mentioned [league/uri](https://github.com/thephpleague/uri/blob/master/src/Schemes/Http.php#L26) earlier in this thread?

> But not providing a single aggregated factory interface is just being annoying to users.

It's true, enforcing SOLID principles is often annoying but ultimately worth it when testing or needing to refactor some code.

Your arguments for a composite interface ignore the real world scenario that exists: HTTP objects are most often created in separate places at separate times. Any complete PSR-7 implementation will (most likely) implement all of the interfaces in a single class, __which is totally okay__. The implementation can then self-reference the combined implementation. The split interfaces are much better for clearly stating dependencies:

```php
class FooMiddleware
{
    public function __construct(
        ResponseFactoryInterface $responseFactory,
        StreamFactoryInterface $streamFactory
   ) { /* ... */ }
}
```

When I see this at the top of a file, I immediately know that the middleware will create responses and streams. I don't have to inspect the code more to check if it also creates URIs, because it didn't inject a `UriFactoryInterface`. This is high quality separation of concerns. If the middleware *only* took your `HttpFactoryInterface`, I would have to search the entire code to see what objects might be created.

sagikazarmark:
I don't think there is a limitation that says you can't implement the interfaces in one class. But having a common interface for them is out of scope of this proposal (and is an anti-pattern IMO).

Your doctrine example does not prove your point, but @dbu's and mine: introduce dependency between factories if necessary.

sagikazarmark:
> When I see this at the top of a file, I immediately know that the middleware will create responses and streams.

I still think that you don't need to know about the stream in this case: the body creation could be part of the response.

shadowhand:
> I still think that you don't need to know about the stream in this case

If the body of a response is replaced, it will need a new stream. See https://github.com/zendframework/zend-expressive/issues/347.

sagikazarmark:
You are right, but in that case, you don't need the response factory, as you don't create a response.

michaelcullum:
Can we try and keep discussions on the mailing list please to ensure they are centralised and keep PRs to discussing typos/language.

Thanks

proposed/http-factory/http-factory.md
nyamsprod: the uri default value should be an empty string not `null` as `null` is not allowed in UriInterface

shadowhand: @nyamsprod updated to use a string as the default.

proposed/http-factory/http-factory.md
```diff
@@ -0,0 +1,183 @@
...
+
+```php
+namespace Psr\Http\Message;
+
+interface UriFactoryInterface
+{
+    /**
+     * Create a new URI.
+     *
+     * @param string $uri
```
shadowhand: Reiterating an earlier comment here that was lost:

I think keeping the `$uri` parameter here makes sense, since there is no method in `UriInterface` to set a complete URI, only parts of it. This would be unpleasant:

```php
$uri = $this->uriFactory->createUri()
    ->withScheme('http')
    ->withHost('example.com')
    ->withPath('/');
```

Most implementations of `UriInterface` (Guzzle, Diactoros, Slim, etc) either allow the URI to be passed to the constructor, or provide some other factory to do so (League/URI).

Regardless, implementations of the factory interface will be able to parse the URI however they wish and hydrate the resulting object.

shadowhand:
@drumshaman would you consider being a sponsor of this PSR?

shadowhand:
@opengeek would you consider sponsoring this PSR?

opengeek:
@shadowhand I'll catch up on this proposal early this week and let you know

pmjones:
@shadowhand It occurs to me that there might only be three factories needed, each with multiple methods: one for Response (which hase methods for new Response and Stream), one for Request (which has methods for new Request, Uri, and Stream), and one for ServerRequest (which has methods for new ServerRequest, Uri, Stream, and UploadedFile).  Just a thought for consideration.

pmjones:
Further, as far as middleware is concerned, could it be that the only factories it will need are the Response-related ones? The Request is generated outside of the middleware as an input, so the middleware should not need to create it.

shadowhand:
@pmjones that doesn't solve the problem of packages that only provide a limited number of implementations, such as `league/uri`. However, maybe it would be appropriate in the form:

```php
interface ResponseFactoryInterface extends StreamFactoryInterface { ... }
```

```php
interface RequestFactoryInterface extends
    StreamFactoryInterface,
    UriFactoryInterface
{ ... }
```

```php
interface ServerRequestFactory extends
    StreamFactoryInterface,
    UriFactoryInterface,
    UploadedFileFactoryInterface
{ ... }
```

This is somewhat similar to what @dracony was proposing but keeps the separate interfaces for Stream/Uri/UploadedFile so that packages can implement the non-aggregate interfaces as appropriate.

shadowhand:
Middleware would need response and stream factory, yes.

shadowhand:
@dracony @pmjones I've updated the proposal with 7d6261e7f9b022ddf98ab47d4064061576a446cd, which modifies the request and response interfaces to extend lower level interfaces. This ensures that a single injection of e.g. `ResponseFactoryInterface` contains the necessary methods to create both the response itself and the output stream.

pmjones:
@shadowhand I like it!

sagikazarmark:
I don't. I think it kills the separation of concerns. I think it's the same as the "partial implementation" argument. Messages are forced to also implement streams and uris?

shadowhand:
@sagikazarmark I think @dracony made a good case for ensuring that injections don't get crazy. Implementation is fairly straight forward:

```php
class RequestFactory implements RequestFactoryInterface
{
    use UriFactoryTrait;
    use StreamFactoryTrait;

    public function createRequest()
    {
        // ...
    }
}
```

The only downside is that traits will have to be used, since PHP doesn't have multiple inheritance. In an ideal world, we could do:

```php
class RequestFactory
    extends UriFactory, StreamFactory
    implements RequestFactoryInterface
{ ... }
```

shadowhand:
@sagikazarmark it might help if we had an example of a partial message implementation that clearly shows interface extension would cause issues.

sagikazarmark:
> @sagikazarmark it might help if we had an example of a partial message implementation that clearly shows interface extension would cause issues.

True, but even in that case, you can still require an implementation in the constructor

dbu:
i like the interface extension. implementations of the uri interface and
the like need not change, and the RequestFactory can use traits or require
the base factories in constructor and proxy the methods

dracony:
@shadowhand Good middleground, I like it. Most middlewares will probably need to either build request or response, so they will need just one of these. Actually the PSR could provide a compositor implementation like:

```php
class CompositeRequestFactory implements RequestFactoryInterface
{
    public function __construct(UriFactory $uriFactory, StreamFactory $streamFactory) 
    {
          $this->uriFactory = $uriFactory;
          $this->streamFactory= $streamFactory;
    }

    public function createUri($uri = '')
    {
        $this->uriFactory->createUri($uri);
    }

    //...etc
}
```


Also can we please use NULL instead of '' ? It just feels weird

michaelcullum:
As I've previously said

> Can we try and keep discussions on the mailing list please to ensure they are centralised and keep PRs to discussing typos/language.

If discussion continues I'll have to lock this PR which I'd rather avoid if possible.

Thanks

shadowhand:
@michaelcullum apologies, but it seems to me all the discussion has been directly relevant to the content of this PSR. Is this not the appropriate place to discuss the proposal content?

opengeek:
It's certainly easier to discuss code here than it is to try to on Google Groups. Why in the world would you  need to lock a PR because constructive discussions are going on?

pmjones:
Locking and banning appear to be the order of the day.

michaelcullum:
It's been well established that these kinds of discussions should take place on the mailing list, not in pull requests, as here they are significantly more likely to get lost.

dracony:
@michaelcullum But then, we wouldn't be able to actually discuss the ideas, because of self-throttling rules. @shadowhand is really trying to address everyones issues and ideas, and with self throttling the process will just take longer. Having this discussion here has already saved people over 90 emails! 

I understand that mailing list is important, but I can't see why it would be ok for us to talk about the PSR on the IRC but not here.

michaelcullum:
hmmm. You make your argument well and in good points. Once the PSR is formalised we'll have a chat with Woody/shadowhand as Editor(s) about these this PSR and the other, it's just we don't do that until the entrance vote is through normally, and this hasn't even had an entrance vote yet. Ideally you guys will come up with some place where you'll use as your main communication medium for this kind of thing. Whether that be github issues on a fork or your own mailing list, perhaps in addition to gitter or the like.

My suggestion would be, get this through an entrance vote, then work on the spec content. You're able to have, but not expected (in fact it's discouraged), to have written, debated and discussed a PSR before the entrance vote.

Also, self-throttling shouldn't actually result in points not being made, simply that you reply less often with small points, but more often with more thought out replies addressing multiple points.

I'll discuss this with the other secretaries as you make fair points, but I've also had project representatives ping me to ask if we can ask you to stop as lots of people are missing out on this discussion and I can see both sides of the argument. This isn't a problem after PSRs have gone to entrance votes as afterwards there are other mediums you can use like a dedicated mailing list, but it's awkward trying to come up with a process before the entrance vote.  However I would say that along the lines of the IRC discussion topic on the ML, this is fine, so long as you're surmising discussions to the mailing list.

dracony:
@shadowhand @pmjones Can we get it to an entrance vote then? 90+ messages in the thread are a clear indicator that people actually care about this one. And if they don't , and it doesn't get through the entrance vote, we're kind of just wasting time ))))

I can also sponsor it, just say the word =)

dbu:
ups, i was not aware that there had been no entrance vote. it looks to
me like this proposal is very ready for the entrance vote as its already
quite clear.

lets wait with further discussion until the vote happened.

michaelcullum:
In the meantime, to keep the discussion momentum going, you're welcome to discuss here so long as you surmise to the mailing list the different discussion points. Although I'd recommend maybe hopping on IRC at some point to has out points there, it can be quite useful to have such sessions.

I hope you'll agree that that way we've got the best of both worlds

proposed/http-factory/http-factory-meta.md
```diff
@@ -0,0 +1,108 @@
...
+5. People
+---------
+
+### 5.1 Editor(s)
+
+* Woody Gilk, <woody.gilk@gmail.com>
+
+### 5.2 Sponsors
+
+* Roman Tsjupa, <draconyster@gmail.com> (Coordinator)
```
shadowhand: @dracony you're not the coordinator, since @pmjones is coordinating #755.

shadowhand:
@dracony this is now ready for an entrance vote.

dracony:
Well it has to go through the review phase first: https://groups.google.com/forum/#!topic/php-fig/PrOxm7Lq-zE

AndrewCarterUK:
No it doesn't @dracony  - https://github.com/php-fig/fig-standards/blob/master/bylaws/004-psr-workflow.md#21-pre-draft

shadowhand:
@dracony @AndrewCarterUK is correct: Entrance -> Review -> Acceptance.

dracony:
Arghhh, wow, I'm amazing at reading today =\

dracony:
...aand the vote has started

Crell:
Here's the key problem with a GH thread: I'm subscribed to the entire repo.  According to Google, I am subscribed to this thread right now.  I've not gotten a single notice about it, once, so only discovered its existence by accident.

The discoverability of GitHub threads is well-known to be godawful. :frowning:

sagikazarmark:
Personally I think the people working on a proposal/draft should have the freedom to decide how they communicate. Especially because this self-throttling rule makes the mailing list useless when it comes to actual communication, not novel writing.

The most that can be required to at least post a link to the discussion on the list and ask the people working on it to post every bigger change to the list as well. Any other rule would only make everyone's life harder.

shadowhand:
@Crell this sounds like configuration error. If you are Watching the repo, you get notifications for everything unless you unsubscribe from specific issue, or if you [turned off notification emails](https://github.com/settings/notifications).

shadowhand:
As per [the voting results](https://groups.google.com/d/msg/php-fig/pCsF-dSXUqo/XRs8Zky4EgAJ) this proposal is now accepted and ready to be merged.

Ping @michaelcullum @squinones

shadowhand:
@michaelcullum @squinones ping again for a merge. 

cc @dracony @pmjones

proposed/http-factory/http-factory-meta.md
```diff
@@ -0,0 +1,109 @@
...
+* [PSR-7 Middleware Proposal](https://github.com/php-fig/fig-standards/pull/755)
+* [PHP-FIG mailing list discussion of middleware](https://groups.google.com/forum/#!topic/php-fig/vTtGxdIuBX8)
+* [ircmaxwell All About Middleware](http://blog.ircmaxell.com/2016/05/all-about-middleware.html)
+* [shadowhand All About PSR-7 Middleware](http://shadowhand.me/all-about-psr-7-middleware/)
+* [AndrewCarterUK PSR-7 Objects Are Not Immutable](http://andrewcarteruk.github.io/programming/2016/05/22/psr-7-is-not-immutable.html)
+* [shadowhand Dependency Inversion and PSR-7 Bodies](http://shadowhand.me/dependency-inversion-and-psr-7-bodies/)
+* [PHP-FIG mailing list thread discussing factories](https://groups.google.com/d/msg/php-fig/G5pgQfQ9fpA/UWeM1gm1CwAJ)
+* [PHP-FIG mailing list thread feedback on proposal](https://groups.google.com/d/msg/php-fig/piRtB2Z-AZs/8UIwY1RtDgAJ)
+
+8. Errata
```
michaelcullum: This can be removed as there is no Errata just yet

proposed/http-factory/http-factory-meta.md
```diff
@@ -0,0 +1,109 @@
...
+
+In the case of `UriInterface` a complete URI may be passed for convenience.
+
+The method names used will not conflict. This allows for a single class to
+implement multiple interfaces when appropriate.
+
+#### 4.2 Existing Implementations
+
+All of the current implementations of PSR-7 have defined their own requirements.
+In most cases, these requirements are the same or very similar.
```
michaelcullum: Not now, but during the draft phase, this should probably be expanded upon with further details on existing implementations.

michaelcullum:
The couple of comments I made should be addressed ASAP, but I'm happy for them to be in their own PR.

mindplay-dk:
Glossing over this a while later, I still feel like bundling multiple factory methods into the same interface is wrong.

`ServerRequestFactoryInterface` provides two methods, one for the creation of an empty server request, and another for the creation of a server request from globals. There is no way those two methods are both a dependency of the same consumer?

I still believe you're combining these methods in the interface because you don't want to have to inject too many dependencies - but there is no (or at least no common) use-case in which the same consumer needs to create both an empty and a populated server request; the number of dependencies you need to inject for this purpose is still (in almost every common use-case) one, regardless.

Same situation with `StreamFactoryInterface` - I'm going to need one kind of stream factory to create one stream instance; it's extremely unlikely I'll need more than one of those methods at the same time, so this interface isn't covering any (common) real-world use-case.

sagikazarmark:
I agree with Rasmus 100%. Why not just let the implementation decide how to handle stream creation: inject in the constructor of a message factory or implement a simple stream factory as well in it with protected/private visibility.

As for the consuming code: the above examples are perfect why the SRP should not be broken here.

harikt:
Hey @mindplay-dk ,

The `StreamFactoryInterface` should need multiple methods for sure. It depends on what you want to return. You can see some of the examples shown in https://github.com/phly/psr7examples

mindplay-dk:
@harikt where? I don't see an example in there.

I think you misunderstand - I'm not suggesting we have only one means of creating a stream, that wouldn't make any sense. I'm suggesting each type of stream have it's own dedicated stream factory interface instead, e.g.:

```php
interface MemoryStreamFactoryInterface
{
    /**
     * Create a new memory-based stream from a string.
     *
     * A temporary resource will be created with the content of the string.
     * The resource will be writable and seekable.
     *
     * @param string $body
     *
     * @return StreamInterface
     */
    public function createMemoryStream($body = '');
}

interface CallbackStreamFactoryInterface
{
    /**
     * Create a new stream from a callback.
     *
     * The stream will be read-only and not seekable.
     *
     * @param callable $callback
     *
     * @return StreamInterface
     */
    public function createStreamFromCallback(callable $callback);
}

interface ResourceStreamFactoryInterface
{
    /**
     * Create a new stream from a resource.
     *
     * @param resource $body
     *
     * @return StreamInterface
     */
    public function createStreamFromResource($body);
}
```

Note that I folded the generic "stream" and "string stream" factories into one, since they both create a writable/seekable (temp) memory stream - I added a default argument of empty string, which makes the implementation and usage of those two factory methods identical.

mindplay-dk:
@shadowhand What about iterator/generator stream factories? Aren't these common enough that we should have interfaces for them?

Crell:
This seems to me like a question that should be answered experimentally.  Vis, let's create a repository with the current draft interfaces, expose it to Packagist, and have various people build with them and see where the pain is.  Then adjust as necessary.

I'm happy to participate in said experimentation if someone can push the package for us to work from.

nyamsprod:
@Crell I believe @shadowhand already made this package based on Diactoros https://github.com/shadowhand/http-factory-diactoros but not sure if it is on packagist yet

harikt:
@nyamsprod no. It is not in packagist. But you can always play with it without having it on packagist also as in https://github.com/shadowhand/http-factory-diactoros/blob/b49e40474564a9b77fc9a84e729462e3b00c034f/composer.json#L2-L9 . Probably what @Crell was suggesting may be let people play with it before the vote begins. Something like alpha for psr ?

Crell:
Not just before vote begins, really.  We want to move to having the "trial implementation" phase be required for FIG 3, but it's a good idea to do now anyway.  It helps flesh out the warts.  (It really helped on PSR-7, and the lack of it on PSR-6 hurt it.)

pmjones:
Sounds like a splendid candidate for a *-interop project; no FIG 3.0 needed there.

harikt:
@mindplay-dk from https://en.wikipedia.org/wiki/Single_responsibility_principle 

> A class should have only one reason to change.

Does a factory change anything for it always returns `StreamInterface` ? According to my knowledge, no. So does it breaks SRP?

I may be wrong, forgive for my ignorance.

Regarding https://github.com/phly/psr7examples I was just pointing out example.

Thank you

nyamsprod:
@harikt @Crell I've already been playing with it and implementing the stream factory is a bit overkill if I'm only interested in a the `StreamFactory::createStream` method yet I'm **force** to implement the other methods that I don't really need

mindplay-dk:
On Wed, Aug 10, 2016 at 5:53 PM, Hari K T <notifications@github.com> wrote:

> @mindplay-dk <https://github.com/mindplay-dk> from
> https://en.wikipedia.org/wiki/Single_responsibility_principle
>
> A class should have only one reason to change.
>
> Does a factory change anything for it always returns StreamInterface ?
> According to my knowledge, no. So does it breaks SRP?
>
> Yes, it does.

Stream is just an interface - what we're describing is the ability to
create different distinct types of Streams; each factory and each Stream
implementation are potentially completely different classes with nothing in
common beyond implementing the same interface.

Also, each type of Stream will satisfy a different use-case - the same
middleware component is with all likelyhood never going to need to create
two different types of Streams, or even two different Streams of the same
type for that matter.

Also, as Ignace pointed out below...:

On Wed, Aug 10, 2016 at 5:53 PM, ignace nyamagana butera <
notifications@github.com> wrote:

> @harikt <https://github.com/harikt> @Crell <https://github.com/Crell>
> I've already been playing with it and implementing the stream factory is a
> bit overkill if I'm only interested in a the StreamFactory::createStream
> method yet I'm *force* to implement the other methods that I don't really
> need
>
This is a clear example of how bundling essentially unrelated factories for
different types, into a single interfaces, forces you to depend on factory
features you don't really depend on at all. Mixing factories clearly blurs
responsibilities and creates dependency clutter - good OOP principles have
been teaching us this for many years, I don't think experimental validation
is required to verify that.

(not saying we shouldn't experiment with the factory interfaces, of course
- verifying the concept in practice is still crucial for many other
reasons!)

nyamsprod:
Hi all. 

As I mentioned yesterday, I've been playing around with the current HTTP Factory proposal. The goal was to see how I could use it to create a simple PSR-7 HTTP client using any PSR-7 implementation. You can view [the result of this experiment here](https://github.com/nyamsprod/psr7-http-client).

I know @shadowhand used Diactoros to implement the factories. Because I'm more familiar with Slim I tried to implement them using Slim. I'm not a Slim expert but creating them was fairly easy. You can view [my implementation here](https://github.com/nyamsprod/http-factory-slim). It is merely a port of @shadowhand works but on Slim. 

Now that we have at least 2 implementations here's my findings:

- The `UploadedFileFactoryInterface::createUploadedFile` expects its first argument to be a string OR a resource. Well in Slim you can not instantiate and UpdloadedFile object with a resouce as I understand it, @akrabat may correct me,  so this factory is not easily adaptable. Maybe this can be easily fix in Slim, or we may have to modify the `UploadedFileFactoryInterface::createUploadedFile` argument signature.

- The `StreamInterface` as it stand can not be easily implemented with Slim. The reason is that there's no `CallbackStream` in Slim. Since this particular stream implementation is not required by PSR-7 I assume that many other implementations won't have it. This means that the HTTP Factory is forcing the addition of something that is not required by PSR-7 and that maybe not so many developers use.

- Another issue is that I believe that having `StreamInterface::createStreamFromResource` and `StreamInterface:createStreamFromString` is kind of redundant as the latter can be achieve using the former.

So IMHO the `StreamInterface` needs to be re-think or improved on the basis that unlike the other interface. A `StreamInferface` implementing object could be made in many different ways whereas this is not true for the UriFactoryInterface for instance. The question being does the StreamFactoryInterface needs to cover all possible Stream objects that could be implemented or not ?
 
I understand that HTTP Factories were first brought up to improve the Middleware proposal but as you can see with my experiment they can be used elsewhere. Of course feedback and improvement of my two repos are welcomed.

harikt:
> The StreamInterface as it stand can not be easily implemented with Slim. The reason is that there's no CallbackStream in Slim.

That doesn't means Slim cannot implement it.

> Since this particular stream implementation is not required by PSR-7 I assume that many other implementations won't have it. This means that the HTTP Factory is forcing the addition of something that is not required by PSR-7 and that maybe not so many developers use.

I believe that is why this http-factory will improve PSR-7 . It is something like PSR-2 which is on top of PSR-1. It looks good to me to have all different implementations for this. As I mentioned earlier I am looking at https://github.com/phly/psr7examples/ for mwop have very much looked into PSR-7 .

nyamsprod:
@harikt the problem is not that Slim don't have CallbackStream, the problem is trying to create a factory which will incliude each possible StreamInterface implementation that exists. Look at [guzzle streams implementations](https://github.com/guzzle/psr7/tree/master/src). How many of them should be taken into account by the StreamFactoryInterface ? just like @mindplay-dk said at some point it is better to split the factory or found a common way to address this issue instead of creating a huge interface that no-one will try to implement.

my 2cent.

weierophinney:
@nyamsprod —

> The StreamInterface as it stand can not be easily implemented with Slim. The reason is that there's no CallbackStream in Slim. Since this particular stream implementation is not required by PSR-7 I assume that many other implementations won't have it. This means that the HTTP Factory is forcing the addition of something that is not required by PSR-7 and that maybe not so many developers use.

You're generalizing without evidence.

Callback streams were one of the first custom streams we added to Diactoros... because it was requested frequently immediately. Iterator/Generator streams were not far behind.

Yes, they are not required by PSR-7, but PSR-7 *also does not say a StreamInterface implementation MUST be backed by a stream resource*.

weierophinney:
What I'm seeing in this thread are two competing paradigms for how the factories might be used.

- @mindplay-dk and others are arguing for greater granularity. My assumption is that this use case is for *injecting factories into middleware as dependencies.*
- @shadowhand, @crell, and others are arguing from the perspective of including the factories *as part of the FrameInterface of PSR-15*.

In the first paradigm, the middleware would compose whatever factories it requires:

```php
class SomeMiddleware implements MiddlewareInterface
{
    private $streamFactory;
    private $responseFactory;
    
    public function __construct(CallableStreamFactoryInterface $streamFactory, ResponseFactoryInterface $responseFactory)
    {
        $this->streamFactory = $streamFactory;
        $this->responseFactory = $responseFactory;
    }
    
    public function process(ServerRequestInterface $request, FrameInterface $frame)
    {
        return $this->responseFactory->build(
            $this->streamFactory->build(Closure::fromCallable([$this, 'someMethod'])
	);
    }
}
```

The second use case looks like the following:

```php
class SomeMiddleware implements MiddlewareInterface
{
    public function process(ServerRequestInterface $request, FrameInterface $frame)
    {
        return $frame->createResponse(
            $frame->createCallableStream(Closure::fromCallable([$this, 'someMethod'])
	);
    }
}
```

The arguments for both are reasonable:

- Granular interfaces:
  - allow implementing libraries the ability to choose which factories they want to implement
  - allow granular composition within middleware
- A "merged" interface for use with `FrameInterface`:
  - provides an easier use case for middleware authors
  
The primary argument against merging interfaces I've seen is with the stream factory, as not every library implementing PSR-7 will necessarily provide all possible stream implementations. This also makes sense.

Here's my thought: make the granular interfaces, but then have one or more "kitchen sink" interface that inherits from all of them or a subset. This will allow the second use case, while keeping the first.

weierophinney:
One other note: I do not think that *implementation* must be an all-or-nothing approach. Guzzle, for instance, originally chose not to implement PSR-7's `ServerRequestInterface` or `UploadedFileInterface`, as they were not relevant to its primary use case as an HTTP *client*. I think implementations of this PSR can also be selective in what they implement, so long as they document what they do and do not implement, so users can know what is and isn't possible when consuming the library.

dbu:
> Here's my thought: make the granular interfaces, but then have one or
> more "kitchen sink" interface that inherits from all of them or a
> subset. This will allow the second use case, while keeping the first.

would it be possible that the PSR code itself also contains an
implementation of the kitchen sink? having a constructor with each
granular factory interface and then just forwarding calls to the
respective factory sounds so trivial that it could go into the PSR
itself. i would still keep the kitchensink interface, so a factory
implementing all streams can implement the interface directly.

Crell:
@weierophinney: Clarification: I'm not arguing for either approach at the moment, as I've not tried implementing it personally either way.  That said, it would never have occurred to me to put the factories on the frame interface.  I always assumed it would be injected into each object's constructor as needed.  Remember, there are plenty of objects that may want to spawn PSR-7 objects that are NOT middlewares, and those should be supported.

@dbu: Such boilerplate is probably quite easy to write, yes. Although I would again favor moving that to a -util repository rather than in the spec itself.

Question: Do we know that the number of possible factories is fixed and finite, or is it conceptually infinite?  That would help determine whether an "all in one" factory is even reasonable.  (Vis, if "all" is a small finite value vs. an unbounded value.)

dbu:
-util repo sounds good to me too. of course this can lead to discussions
all over how much to include in util, but for obvious cases like this
proxy it would feel stupid to see multiple implementations of the same
thing.

nyamsprod:
@Crell 

> Question: Do we know that the number of possible factories is fixed and finite, or is it conceptually infinite? That would help determine whether an "all in one" factory is even reasonable. (Vis, if "all" is a small finite value vs. an unbounded value.)

Well technically the number of factory is conceptually infinite with the StreamInterface

mindplay-dk:
@weierophinney 

> A "merged" interface for use with FrameInterface: provides an easier use case for middleware authors

I agree this makes things *easier*, but at the same time, I think it makes things *less simple* than they could be.

I think that this kind of concept belongs in a framework, not in a standard - the standard exists to make things interoperable and generic; the frame concept is opinionated, not about interoperability, and not generic in the sense that it provides a lot of freedom to come up with your own patterns - a simple set of factories meet those requirements with, yes, somewhat less convenience, but still does satisfy the requirements, and *does not* dictate architecture.

@Crell

> it would never have occurred to me to put the factories on the frame interface. I always assumed it would be injected into each object's constructor as needed. Remember, there are plenty of objects that may want to spawn PSR-7 objects that are NOT middlewares, and those should be supported.

Yes, yes and yes :-)

> Do we know that the number of possible factories is fixed and finite, or is it conceptually infinite? 

Excellent point!

If we decide on four kinds of stream factory methods for four types of streams today, we will almost certainly need to make a breaking change to that interface in the future, when somebody comes up with a useful new type of stream, or a useful new way to construct an existing type of stream.

Same is true for the two ways you might currently create a request object - there's no guarantee we won't come up with a third, like (say) constructing a request object from raw HTTP data.

If we segregate factories into separate interfaces, arguably this makes it a lot simpler and less destructive to add to the standard in the future. In my opinion, that argument is practically unbeatable.

shadowhand:
@mindplay-dk I think I'm leaning towards the narrowly scoped interfaces. Any project that chooses to implement the interfaces would be able to implement multiple interfaces within the same class for any types that can be constructed.

@weierophinney It looks like your `FrameInterface` example is implying that `Frame` is also a HTTP factory, which is definitely __not__ something I have advocated for and I do not think it would be a good idea. Major violation of SRP happening there.

I have a number of fixes that are reflected in my local copy of [http-factory](https://github.com/shadowhand/http-factory) that are not reflected in the current meta doc that came out of trying to implement `http-factory-diactoros`. I will work thought them today and update the document, `http-factory

mindplay-dk:
@shadowhand awesome sauce :-)

shadowhand:
I think `UploadedFileFactoryInterface` accepting both `string|resource` is a bug. It should only accept a string, because that is how [PHP expresses an uploaded file][1] in `$_FILES['userfile']['tmp_name']`.

[1]: http://php.net/manual/en/features.file-upload.post-method.php

shadowhand:
Similarly, I don't think the `$size` parameter is necessarily or desirable, since the file size can be determined at runtime by using `filesize($tmp_file)`.

mindplay-dk:
@shadowhand I agree that support for `string` path is likely enough, at least initially. I'm not certain that there is *no* use-case for creating an uploaded file  from a `resource`? But, either way, if we were to support both, they should be two different factory interfaces - both result in an uploaded file model, but the factory implementations would be completely distinct; and, there's no saying every framework would (or could) support both factory methods.

shadowhand:
Perhaps @weierophinney can fill us in on the use case for creating an uploaded file from a resource, since Diactoros supports it.

weierophinney:
> Perhaps @weierophinney can fill us in on the use case for creating an uploaded file from a resource, since Diactoros supports it.

I can indeed! The use case is for non-SAPI servers — in particular, async servers such as ReactPHP, Icicle, etc. In those cases, those servers tend to create `php://memory` or `php://temp` resources for the uploaded files; `UploadedFileInterface` was designed in large part with that in mind, as we wanted to ensure the possibility of PSR-7 being usable on those platforms.

PR #762, opened 2016-05-27
Apparently this needed to be written down

philsturgeon:
It seems fairly obvious, and the FIG Secretaries considered it to be fairly obvious, but [Paul Jones was freaking out again](https://groups.google.com/forum/#!topic/php-fig/gMnU02lU1D0).

The FIG does not need to get into the habit of making bylaws for every damn decision the secretaries make, but this should calm the situation down a bit.

### DIFF
--- a/bylaws/003-membership.md
+++ b/bylaws/003-membership.md
@@ -70,7 +70,7 @@ become a Member Project.
 Voting Representatives
 -----------------------
 
-The votes of all Member Projects are cast by Voting Representives who have been
+The votes of all Member Projects are cast by Voting Representatives who have been
 authorised to do so by the Member Project. Each Voting Representative is chosen
 solely by a Member Project and their selection is not subject to the approval
 of PHP-FIG. A Voting Representative may be replaced by the Member Project which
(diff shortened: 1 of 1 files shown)


kayladnls:
:+1:

rask:
Okay, I am a userland developer, but this needs to be asked:

Why is someone in the FIG board (and voting) while their votes are being nulled? Would this mean that they should not be on the board at all? Why not expell the said party from voting in the first place?

pmjones:
Paul Jones is a fan of votes being counted unless they are expressly to be discounted, and not leaving it to whim. Call the defense of that principled position a "freak-out" if you wish.

For my part, I have no objection to this being explicitly specified.

tedivm:
@philsturgeon- I yelled at Paul a week ago for his calling you out unnecessarily, and now I'll do the same to you. This thread could have been opened in a lot more professional manner. Perhaps you guys should both just vow never to directly refer to the other person, as it would solve a lot of issues.

I also don't think this solved the problem that came up on the list. The issue isn't about whether members should be allowed to vote on their own expulsion. The question is whether secretaries can remove voting members rights when there is nothing in the bylaws that allows this. This change you are proposing solves part of the problem, but there should be another line that makes it explicit that all members can vote on all issues unless the bylawys specifically say they can't. Otherwise this issue will just crop up again when the Secretaries decide to restrict people from voting.

pmjones:
> Perhaps you guys should both just vow never to directly refer to the other person, as it would solve a lot of issues.

Fine with me; if Phil can keep his mouth shut about me, even from behind a Twitter block, I can certain let his stay beneath my notice as well.

michaelcullum:
I think we can all just be a bit nicer to each other. Discussion of this can continue on the mailing list.

PR #752, opened 2016-04-22
FIG 3.0

michaelcullum:
Bylaws in their split up form. You can view the new bylaws in a single document for ease of reading [here](https://docs.google.com/document/d/1ZlJiOB-Ab_c-8-6BHpQhu0ZqE2Rr1kP7sWoWBbM7v_Q/edit?usp=sharing).

Once merged (and the links exist) we will add cross links between the bylaws for easier navigation. For example with links to the relevant votes from the PSR workflow.

Please discuss these changes [here](https://groups.google.com/forum/#!topic/php-fig/h3wrQePdzfc)

PR #764, opened 2016-05-27
Clarify the role of secretary as "aide" not "judge"

pmjones:


Jean85:
Shouldn't this be closed or rebased/reworded since #752 got merged?

PR #766, opened 2016-05-29
EventManager: proposal of different set of contracts

unkind:
#### Issues of existing EventManager proposal

In this PR I'd like to pay your attention to some issues of existing proposal:
- Naming. Existing proposal doesn't differentiate events and hook messages. The difference is that events are immutable messages and they don't affect the caller. Potentially, it may introduce some issues in future implementations. For example, events can be queued, because caller doesn't care about response; unlike events, hook messages must be dispatched at runtime. Queued hook message doesn't make sense, so it may lead to bugs.
- `EventManagerInterface` doesn't respect interface segregation principle. In most cases, you need just single method to dispatch/trigger event and you don't care about listeners configuration. Some implementations like `QueuedEventManager` would suffer from requirement of `attach`, `detach` and `clearListeners` methods.
- Vague and ambiguous signature of the `trigger()` method.
- `EventInterface::setName()` seems like a harmful method. Pass event/hook name through constructor, there is no reason to change it later.
- `EventManagerInterface::attach()` and `detach` return bool. It's very unclear what's supposed to do when it returns false. Try to attach again? :)
-  The `EventInterface` doesn't help much in framework-agnosticism. This statement requires further explanation. See below.
#### What's the problem with `EventInterface`?

If I understand correctly, the idea of this PSR is to make callers and listeners framework-agnostic.

In order to dispatch an event, caller has to create instance of an event object. Usually, it would be (sub)class of some existing event-library such as Symfony EventDispatcher's `Symfony\Component\EventDispatcher\Event`. If event is framework-aware, then abstraction is leaky. In this context `EventInterface` doesn't help.

On other hand, the interface helps to make listeners framework-agnostic if they type hint against  `EventInterface`. However, it hurts readability and makes impossible to autocomplete event args:

``` php
public function onProjectWasRenamed(EventInterface $event);
// vs 
public function onProjectWasRenamed(ProjectWasRenamed $event);
```

But if listeners depend on concrete event like `ProjectWasRenamed`, they would become framework-aware as well, because of base class `Symfony\Component\EventDispatcher\Event`.

It makes sense to avoid an interface if you don't want to end up with `EventFactoryInterface` (looking at you, PSR-7). 
#### Suggestions
1. Replace `EventInterface` with `Event` and `Hook` classes, so developer is able to extend them or use directly:
   
   ``` php
   use Psr\EventManager\Event;
   
   final class ProjectWasRenamed extends Event
   {
       // ...
   }
   
   $this->eventDispatcher->dispatch(new ProjectWasRenamed(...));
   
   // or
   
   $this->eventDispatcher->dispatch(new Event('project_was_renamed'));
   ```
   
   Also it allows to distinguish between  `Event` and `Hook`  very clearly:
   
   ``` php
   final class PreResponseHook extends Hook
   {
       // ...
   }
   
   final class ResponseWasSent extends Event
   {
       // ...
   }
   
   // You are able to modify response in handlers
   $hook = $this->eventDispatcher->hook(new PreResponseHook($response));
   $response = $hook->getResponse();
   $this->send($response);
   
   // You are stating a fact, handlers can't do anything with response anymore
   $this->eventDispatcher->dispatch(new ResponseWasSent($response));
   ```
2. Split the `EventManagerInterface` into `EventDispatcherInterface` and `EventSubscriberInterface`. It makes easier to decorate dispatcher and doesn't force to implement those subscriber methods in general case. `EventSubscriberInterface` is OK for runtime subscriptions, however, in most cases listeners can be subscribed more efficiently at framework level with lazy loading of listeners from DI-container, etc.
3. Make the Event immutable.
4. Remove `target` parameter. It looks like it was inspired by ZF2 EventManager, however, it doesn't look like a perfect idea. It increases coupling and it has lack of explanation why developer needs this.
5. Change return value of `attach` and `detach` to `void`.

There are things I'm not sure about:
- placing `hook()` method;
- name of PSR after splitting the `EventManagerInterface`.

#### How should event-libraries implement this PSR with different base event class?

It's possible to change existing event-libraries so they would able to dispatch an object of `Psr\EventManager\Event` (and `Hook`) by wrapping it into base event class of a library, e.g.

``` php
use Psr\EventManager\Event;
use Psr\EventManager\EventDispatcherInterface;
use Symfony\Component\EventDispatcher\EventDispatcherInterface as SymfonyDispatcher;
use Symfony\Component\EventDispatcher\PsrEvent;

final class SymfonyEventDispatcher implements EventDispatcherInterface
{
    private $dispatcher;

    public function __construct(SymfonyDispatcher $dispatcher)
    {
        $this->dispatcher = $dispatcher;
    }

    public function dispatch(Event $event)
    {
        $this->dispatcher->dispatch($event->getName(), new PsrEvent($event));
    }

    // hook()
}
```

Reverse converting `PsrEvent` → `ProjectWasRenamed extends \Psr\EventManager\Event` is possible by extension point like symfony/symfony#18873.

---

I'd like to know your thoughts. /cc @bretterer @manchuck

### DIFF
--- a/proposed/event-manager.md
+++ b/proposed/event-manager.md
@@ -25,6 +25,8 @@ doesn't have permission.
 
 ## Terms
 
+**TODO**
+
 *   **Event** - An action that about to take place (or has taken place).  The
 event name MUST only contain the characters `A-Z`, `a-z`, `0-9`, `_`, and '.'.
 It is RECOMMENDED that words in event names be separated using '.'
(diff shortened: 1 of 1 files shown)


proposed/event-manager.md
```diff
@@ -36,132 +36,201 @@ priority.  Listeners MUST BE called based on priority.
...
+### Event
 
-### EventInterface
+The Event defines base contract needed to dispatch an event. Each event MUST contain a event name in order trigger the listeners. Each event MUST be immutable. Event class OPTIONALLY may be inherited by concrete event class. OPTIONALLY the event can have additional parameters for use within the event.
 
-The EventInterface defines the methods needed to dispatch an event.  Each event
-MUST contain a event name in order trigger the listeners. Each event MAY have a
-target which is an object that is the context the event is being triggered for.
-OPTIONALLY the event can have additional parameters for use within the event.
+~~~php
```
samdark: `s should be used.

unkind: I'm not sure: #758

PR #767, opened 2016-06-01
Added PHP opening tags

AndrewCarterUK:
Not having these ruins syntax highlighting on the website (it's also more accurate).

http://www.php-fig.org/psr/psr-6/
https://github.com/php-fig/php-fig.github.com/issues/207


michaelcullum:
Thanks

PR #771, opened 2016-06-15
Update interfaces with a lambda style approach

shadowhand:
Based on feedback from the community, a lambda style approach offers
better runtime checking and provides a more resilient foundation for
middleware containers and components.

Also clarifies the goals in the meta document accordingly.

http://blog.ircmaxell.com/2016/05/all-about-middleware.html
https://groups.google.com/d/msg/php-fig/sYecuqdYIkY/Gh5ATLn_EAAJ

PR #773, opened 2016-06-15
Rename handle() to process()

shadowhand:
Using the same method name for the stack and middleware is logical and
may reduce conflicts with existing middleware implementations.

Refs #771


shadowhand:
@weierophinney this should resolve the conflict, as referenced in https://groups.google.com/d/msg/php-fig/sYecuqdYIkY/OZ1GIfAOEwAJ

weierophinney:
:+1: from me! Thanks, @shadowhand !

shadowhand:
@pmjones this is ready to be merged.

PR #774, opened 2016-06-15
Add chosen approach explanation to meta document

shadowhand:
Provide a historical record for why lambda was chosen over double pass.

Refs #771


shadowhand:
@weierophinney do you think I need to add anything else to this regarding the reasoning for lambda?

proposed/http-middleware/middleware-meta.md
```diff
@@ -171,6 +170,33 @@ that are based around StackPHP.
...
+
+The double-pass approach has some significant issues regarding implementation.
+
+The most severe is that passing an empty response has no guarantees that the
+response is in a usable state. This is further exacerbated by the fact that a
+middleware may modify the response before passing it for further dispatching.
+
+Further compounding the problem is that there is no way to ensure that the
+response body has not been written to, which can lead to incomplete output or
+error responses being sent with cache headers attached.
```
weierophinney: Another issue, as detailed in a [blog post by Rob Allen](https://akrabat.com/filtering-the-psr-7-body-in-middleware/), is that it's difficult to get the response *body* into a clean state if it has been previously written to; rewinding and writing back to it can lead to issues if the new content is of shorter length than the previous. As such, developers currently must depend on a concrete `StreamInterface`, populate it, and pass it to generate a new response.

proposed/http-middleware/middleware-meta.md
```diff
@@ -171,6 +170,33 @@ that are based around StackPHP.
...
+
+Some have argued that passing the response helps ensure dependency inversion.
+While this it is true that it helps avoid depending on a specific implementation
```
weierophinney: `s/this//` (should read, "While it is true…")

proposed/http-middleware/middleware-meta.md
```diff
@@ -171,6 +170,33 @@ that are based around StackPHP.
...
+of HTTP messages, the problem can also be resolved by injecting factories into
+the middleware to create HTTP message objects. With the creation of HTTP Factories
```
weierophinney: > … to create HTTP message objects.

I'd add:

> … objects, or by injecting empty message instances.

proposed/http-middleware/middleware-meta.md
```diff
@@ -171,6 +170,33 @@ that are based around StackPHP.
...
+in PSR-17, a standard approach to handling dependency inversion is possible.
```
weierophinney: Eventually, we'll want to link to PSR-17.

proposed/http-middleware/middleware-meta.md
```diff
@@ -171,6 +170,33 @@ that are based around StackPHP.
...
+
+Using lambda approach ensures a more consistent and predictable developer
```
weierophinney: `s/Using lambda/Using the lambda/`

proposed/http-middleware/middleware-meta.md
```diff
@@ -171,6 +170,33 @@ that are based around StackPHP.
...
+experience, which helps prevent incorrect or less optimal implementation.
+The lambda approach is also somewhat easier to explain and has wider overall
+adoption within the ecosystem due to existing implementations.
+
+Using interface declarations instead of `callable` eliminates some flexibility
+in favor of better runtime safety and debugging.
```
weierophinney: I'd add some narrative here about the fact that it also allows existing implementations to adapt in a way that keeps their own backwards compatibility, while providing forwards compatibility with this proposal.

weierophinney:
@shadowhand Provided some notes inline. :smile:

proposed/http-middleware/middleware-meta.md
michaelcullum: This can now be done

shadowhand:
@weierophinney updated based on your commends.

proposed/http-middleware/middleware-meta.md
```diff
@@ -6,14 +6,18 @@ PSR-15 Meta Document
 
 The purpose of this PSR is to provide an interface that defines the formal
 method signature for HTTP Middleware that is compatible with HTTP Messages,
-as defined in PSR-7.
+as defined in [PSR-7][psr-7].
+
+[psr7]: http://www.php-fig.org/psr/psr-7/
```
weierophinney: `s/psr7/psr-7/` (otherwise I think the link reference breaks)

shadowhand: Fixed.

proposed/http-middleware/middleware-meta.md
```diff
@@ -112,17 +118,16 @@ and is defined as:
...
 
-* The middleware is defined as a [callable][php-callable] using a [closure][php-closure]
-  or a class with an `__invoke` method.
+* The middleware is defined with a specific interface with a method that takes
+  the request for processing.
 * The middleware is passed 2 arguments during invocation:
   1. A `RequestInterface` implementation.
-  3. A `callable` that receives the request to dispatch next middleware.
-
+  3. A `FrameInterface` that receives the request to dispatch next middleware.
```
weierophinney: `s/3/2/` (as there are only two arguments now)

shadowhand: Fixed.

proposed/http-middleware/middleware-meta.md
```diff
@@ -171,6 +176,50 @@ that are based around StackPHP.
...
+handling dependency inversion is possible.
+
+[psr17]: https://github.com/php-fig/fig-standards/blob/master/proposed/http-factory/http-factory-meta.md
+
+Using the lambda approach ensures a more consistent and predictable developer
+experience, which helps prevent incorrect or less optimal implementation.
+The lambda approach is also somewhat easier to explain and has wider overall
+adoption within the ecosystem due to existing implementations.
+
+Using interface declarations instead of `callable` eliminates some flexibility
```
weierophinney: One thing we could add in here is that implementing libraries can provide the flexibility of `callable` using a decorator class:

```php
class CallableMiddleware implements ServerMiddlewareInterface
{
    private $callable;

    public function __construct(callable $callable)
    {
        $this->callable = $callable;
    }

    public function process(ServerRequestInterface $request, FrameInterface $frame)
    {
        $middleware = $this->callable;
        return $middleware($request, $frame);
    }
}
```

This approach allows for the type-safety of using strict interfaces, while providing the usability of defining middleware via callables.

shadowhand: I'm always hesitant to add too much code to meta documents, but this probably makes sense.

weierophinney: Considering that almost every middleware microframework currently uses `callable` middleware, this seems like a good target for the meta document, as I predict the question of "How can I use PSR-15 with callable middleware?" will come up frequently. The fact that the code sample can be so succinct appeals to me. :smile:

proposed/http-middleware/middleware-meta.md
```diff
@@ -171,6 +176,50 @@ that are based around StackPHP.
...
+in favor of better runtime safety and debugging. Using stricter interfaces also
+makes it easier for existing implementations to provide wrappers or adapters
+to provide forward compatibility with the proposed interfaces.
+
+#### 4.5 Middleware Stack and Dispatching
+
+This proposal also includes a definition of a middleware stack container that
+is an entry point for middleware dispatching. This interface is entirely optional
+and is intended to illustrate that client middleware can be used in a server
+context but server middleware cannot be used in a client context.
```
weierophinney: I've replied on the mailing list for this aspect.

proposed/http-middleware/middleware-meta.md
shadowhand: @weierophinney I took at stab at writing a method that wrapped `callable` middleware, but it requires interfaces from PSR-17 and really doesn't work properly. Can we carry on without addressing this for now and look at it again when PSR-17 is finished? (I anticipate that PSR-17 will be released first, as it is far less contentious and is a fundamental building block for proper PSR-15 implementation.)

shadowhand:
Closing this. Will resubmit on a clean branch.

PR #775, opened 2016-06-15
Link template handling

Crell:
As discussed on list.

Open question: Should we add utility methods to `LinkCollectionInterface` to get just templated or untemplated values?  Or does that result in too much complexity alongside `getLinksByRel()`? Or should that and `getLinks()` get a parameter to filter by templated or not?


Crell:
Discussion on the list suggests no, we shouldn't add those. Filtering it out of scope.

PR #777, opened 2016-06-19
Method body indentation rule added + updated formatting

hanvari:
- In section "4.3. Methods", the rule for method body indentation was missing which is added.
- Also the formatting of the rules listing in this section is updated as bullet-points, as in section "5. Control Structures"


samdark:
PSR-2 can't be modified. Next coding standard which is in the works is PSR-12 which you can send pull request against.

PR #778, opened 2016-06-20
Issue with return type

phpinfointh:
Current spec says that there MUST not be spaces before colon.
But this raises readability issue with multi-line parameters.

If we can have one space before colon, return type declarations can be aligned beautifully like parameters.

``` php
<?php
namespace Vendor\Package;

class ClassName
{
    public function someMethod() : string
    {
        // method body
    }

    public function aVeryLongMethodName(
        ClassTypeHint $arg1,
        &$arg2,
        array $arg3 = []
    ) : array
    {
        // method body
    }
}
```

### DIFF
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -52,7 +52,7 @@ use const Vendor\Package\{ConstantA, ConstantB, ConstantC};
 
 class Foo extends Bar implements FooInterface
 {
-    public function sampleFunction(int $a, int $b = null): array
+    public function sampleFunction(int $a, int $b = null) : array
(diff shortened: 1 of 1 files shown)


samdark:
Personally I don't agree. Need a survey on the matter.

rob006:
> If we can have one space before colon, return type declarations can be aligned beautifully like parameters.

This actually makes this less readable - it is easier to miss return type when it is aligned in the same way as parameters. Do not assume that "more beautiful" == "more readable" - in many cases it is quite the opposite.

michaelcullum:
This has gone through the survey which agrees with the current implementation

PR #780, opened 2016-06-21
Name Michael Hess Editor of PSR-9/10

Crell:
Michael has agreed to take over PSR-9 and PSR-10.  Also, the Coordinator was listed wrong on PSR-10.  (Korvin and I are coordinating one spec each at present, per the index.)


michaelcullum:
Shall we clarify Lukas as `(Former Editor)`?

Crell:
I don't think we've done that in the past, have we?  I've largely just moved former Editor/Sponsor/Coordinators to Contributors and left it at that.

Crell:
For reference, Michael Dowling is listed as a Contributor on PSR-7 with no additional marker, despite having been the original editor: https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-7-http-message-meta.md

Crell:
Note to self: update the index file, too. :smile:

PR #782, opened 2016-06-26
Add createUploadedFiles()

webdevxp:
Should `UploadedFileFactoryInterface` have a method to create array tree of `UploadedFileInterface` from source array (for example `$_FILES`)?


samdark:
What for?

PR #783, opened 2016-06-29
Add PSR-12 default class property values

kj187:


krzysiekpiasecki:
IMO initializing properties at declaration does not improve readability. There is only one proper place to initialize the new instance - ```__construct``` method of course.

Also we have parameters types yet so again the ```__construct``` method is the entry point to the class.

It's readable for me:

```php
public function __construct(int $x, float $r) {
    $this->x = x;
    $this->r = $r;
    $this->pi = 3.14;
    $this->foo = 'bar';
   ...
}

The details should be in a doc blocks inside a class.


From your PR int or float?
```
public $fooCount = 0;```
```

samdark:
I don't think it makes any sense. This sentence doesn't enforce anything and it could be done even with current PSR-1 and PSR-2.

PR #784, opened 2016-07-07
fixed a typo in the autoloader example description

royallthefourth:


Crell:
This seems correct, and trivial enough of a fix.  Leaving for the Editor/Secretaries but :+1:

pmjones:
Good catch, and thanks. If there's no objection I'm happy to merge.

michaelcullum:
Thanks @royallthefourth

PR #787, opened 2016-07-26
Add Errata regarding DateTimeInterface handling for pre-5.5 environments

Crell:
Based on implementation discussion in https://github.com/google/google-auth-library-php/pull/126

### DIFF
--- a/accepted/PSR-6-cache-meta.md
+++ b/accepted/PSR-6-cache-meta.md
@@ -291,23 +291,23 @@ specification.
 Despite that, any value that is not an object of  `\DateTimeInterface` (either
 `\DateTime` or `\DateTimeImmutable`, or possibly others added to the language in
 the future) MUST be treated as an invalid syntax error.  Implementers are
-encouraged to actively reject values that do not implement that interface. The
-recommended way of doing so is by using an assertion, as assertions are the
-closest equivalent to type checks.  Throwing `\InvalidArgumentException` is
-another widely used alternative, however, that would be a change in the
-specification (as Calling Libraries would then need to know to catch that
-exception as well).
+encouraged to actively reject values that do not implement that interface.
+Simulating a failed type check unfortunately varies between PHP versions. In PHP 5,
+it is a trigger_error().  In PHP 7, it is a thrown `\TypeError`. Implementing
+Libraries SHOULD try to mimic the behavior of the appropriate PHP version.
 
-Example of a PHP 5.3-compatible assertion:
+The following sample code is recommended in order to enforce the type check on
+the expiresAt() method:
 
 ```php
-assert('$expiration instanceof \DateTime || $expiration instanceof \DateTimeImmutable');
+if (! ($expiration instanceof \DateTime || $expiration instanceof \DateTimeImmutable)) {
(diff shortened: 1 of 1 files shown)


accepted/PSR-6-cache-meta.md
```diff
@@ -276,25 +276,59 @@ default control information is all that's possible), and because it's trivial
...
+specification.
+
+Despite that, any value that is not an object of  `\DateTimeInterface` (either
+`\DateTime` or `\DateTimeImmutable`, or possibly others added to the language in
+the future) MUST be treated as an invalid syntax error.  Implementers are
+encouraged to actively reject values that do not implement that interface. The
+recommended way of doing so is by using an assertion, as assertions are the
+closest equivalent to type checks.  Throwing `\InvalidArgumentException` is
+another widely used alternative, however, that would be a change in the
+specification (as Calling Libraries would then need to know to catch that
```
Tobion: `assert` can throw an exception as well when using http://php.net/manual/en/info.configuration.php#ini.assert.exception
So in that regard using `assert` would be a change of the spec as well.

danhunsaker: In that case, the developer is explicitly converting warnings to exceptions, so the spec isn't violated, because it's the dev throwing the exceptions, not the PSR-6 implementation itself.

accepted/PSR-6-cache-meta.md
```diff
@@ -300,7 +300,7 @@ The following sample code is recommended in order to enforce the type check on
 the expiresAt() method:
 
 ```php
-if (! ($expiration instanceof \DateTime || $expiration instanceof \DateTimeImmutable)) {
+if (! (is_null($expiration) || $expiration instanceof \DateTime || $expiration instanceof \DateTimeInterface)) {
```
mbeccati: `null === $expiration` would be preferrable. Also the `!` shouldn't be followed by a space.

Crell: Why on the former? The two are identical, no?

And is the second a PSR-2 thing?

michaelcullum: It is not a PSR-2 thing no. PSR-2 doesn't touch on operators and PSR-12 excludes ruling on the unary not operator. https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#6-operators

mbeccati: Albeit micro-optimization, `null === $expiration` is much faster as it doesn't require a function call. About unary operators, I believe @michaelcullum is right, but we've used a different style here: https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-6-cache-meta.md#41-chosen-approach

michaelcullum: So long as it is consistent within a spec (Like British and American English), it doesn't matter.

Crell: You're going to use my own sample code against me??? :stuck_out_tongue: 

Updated both.

Seldaek:
:+1: FWIW, ping @php-fig/secretaries

michaelcullum:
@Seldaek Yes? This was voted down in August (didn't meet quorum either) and @Crell wanted to bring it back up (after revisions) once the FIG 3 structure was in place.

Seldaek:
Alright I just thought this made sense and was just stalled/blocked, never mind me :p

Crell:
Closing in favor of #915

PR #788, opened 2016-08-01
Add evolvable interfaces

Crell:
As discussed on list, add optional interfaces for "evolvable" (with-able) objects, a la PSR-7. These are optional by design.

### DIFF
--- a/proposed/links.md
+++ b/proposed/links.md
@@ -101,7 +101,7 @@ that does not support URI Templates MUST ignore any templated Links it encounter
 
 ## 1.5 Evolvable collections
 
-In some cases, a Link Collection may need the ability to hbave additional links
+In some cases, a Link Collection may need the ability to have additional links
 added to it. In others, a link collection is necessarily read-only, with links
 derived at runtime from some other data source. For that reason, modifiable collections
 are a secondary interface that may optionally be implemented.
--- a/proposed/links.md
+++ b/proposed/links.md
@@ -108,14 +108,17 @@ are a secondary interface that may optionally be implemented.
 
 Additionally, some Link Collection objects, such as PSR-7 Response objects, are
 by design immutable.  That means methods to add links to them in-place would be
-incompatible.
+incompatible. Therefore, the EvolvableLinkCollectionInterface's single method
+requires that a new object be returned, identical to the original but with
+an additional link object included.
 
 ## 1.6 Evolvable link objects
 
 Link objects are in most cases value objects. As such, allowing them to evolve
 in the same fashion as PSR-7 value objects is a useful option. For that reason,
-an additional EvolvableLinkInterface is included that provides immutable modifiers
-using the same pattern as PSR-7.
+an additional EvolvableLinkInterface is included that provides methods to
+produce new object instances with a single change.  The same model is used by PSR-7
+and, thanks to PHP's copy-on-write behavior, is still CPU and memory efficient.
 
 ## 2. Package
 
--- a/proposed/links.md
+++ b/proposed/links.md
@@ -120,6 +120,10 @@ an additional EvolvableLinkInterface is included that provides methods to
 produce new object instances with a single change.  The same model is used by PSR-7
 and, thanks to PHP's copy-on-write behavior, is still CPU and memory efficient.
 
+There is no evolvable method for templated, however, as the templated value of a
+link is based exclusively on the href value.  It MUST NOT be set independently, but
+derived from whether or not the href value is an RFC 6570 link template.
+
 ## 2. Package
 
 The interfaces and classes described are provided as part of the
@@ -214,15 +218,6 @@ interface EvolvableLinkInterface extends LinkInterface
      * @return static
      */
     public function withHref($href);
-
-    /**
-     * Returns an instance with a specified templated value set.
-     *
-     * @param bool $templated
-     *   True if the link object should be templated, False otherwise.
-     * @return static
-     */
-    public function withTemplated($templated);
 
     /**
      * Returns an instance with the specified relationship included.
--- a/proposed/links.md
+++ b/proposed/links.md
@@ -99,6 +99,24 @@ tool.  Some hypermedia formats support templated links while others do not, and 
 have a special way to denote that a link is a template.  A Serializer for a format 
 that does not support URI Templates MUST ignore any templated Links it encounters.
 
+## 1.5 Evolvable collections
+
+In some cases, a Link Collection may need the ability to hbave additional links
+added to it. In others, a link collection is necessarily read-only, with links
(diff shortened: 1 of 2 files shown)


proposed/links.md
```diff
@@ -99,6 +99,24 @@ tool.  Some hypermedia formats support templated links while others do not, and
 have a special way to denote that a link is a template.  A Serializer for a format 
 that does not support URI Templates MUST ignore any templated Links it encounters.
 
+## 1.5 Evolvable collections
+
+In some cases, a Link Collection may need the ability to hbave additional links
```
weierophinney: `s/hbave/have/`

proposed/links.md
```diff
@@ -99,6 +99,24 @@ tool.  Some hypermedia formats support templated links while others do not, and
...
+Additionally, some Link Collection objects, such as PSR-7 Response objects, are
+by design immutable.  That means methods to add links to them in-place would be
+incompatible.
+
+## 1.6 Evolvable link objects
+
+Link objects are in most cases value objects. As such, allowing them to evolve
+in the same fashion as PSR-7 value objects is a useful option. For that reason,
+an additional EvolvableLinkInterface is included that provides immutable modifiers
+using the same pattern as PSR-7.
```
weierophinney: Describe what "the same pattern as PSR-7" means, so folks do not need to reference that document as well.

proposed/links.md
```diff
@@ -162,14 +181,92 @@ interface LinkInterface
...
+
+namespace Psr\Http\Link;
+
+/**
+ * An evolvable link value object.
+ */
+interface EvolvableLinkInterface extends LinkInterface
+{
+    /**
+     * Returns an instance with the specified href.
```
weierophinney: `s/an instance/a new instance/

proposed/links.md
```diff
@@ -162,14 +181,92 @@ interface LinkInterface
...
+     *     - A relative URI, as defined by RFC 5988. The base of the relative link
+     *       is assumed to be known based on context by the client.
+     *     - A URI template as defined by RFC 6570.
+     *     - An object implementing __toString() that produces one of the above
+     *       values.
+     *
+     * An implementing library SHOULD evaluate a passed object to a string
+     * immediately rather than waiting for it to be returned later.
+     *
+     * @return self
```
weierophinney: `s/self/static/`. `self` implies the same instance is returned, while `static` implies a new instance. (This is something clarified after PSR-7 was ratified.)

proposed/links.md
weierophinney: and same comment throughout these classes.

proposed/links.md
weierophinney: and throughout all new `with()` methods.

proposed/links.md
```diff
@@ -194,3 +291,27 @@ interface LinkCollectionInterface
...
+ * An evolvable link collection value object.
+ */
+interface EvolvableLinkCollectionInterface extends LinkCollectionInterface
+{
+    /**
+     * Returns an instance with the specified link included.
+     *
+     * @param LinkInterface $link
+     *   A link object that should be included in this collection.
+     *
```
weierophinney: Remove the extra empty line. Also, IIRC, lines wrapped from `@param/@return` notations are supposed to be four spaces, no?

proposed/links.md
Crell: The wording here is the same as in PSR-7. I copied that deliberately to handle the "same instance no-op" optimization that happened to PSR-7 late in the game.

proposed/links.md
Crell: I don't believe there is a standard on that. PSR-6 also uses 2 indents.

proposed/links.md
weierophinney: Per discussion on IRC, leave as-is; we used this verbiage specifically in PSR-7 due to late feedback that indicated that "a new instance" would be too narrowly interpreted as not allowing cloning.

proposed/links.md
```diff
@@ -162,14 +184,92 @@ interface LinkInterface
...
+     * @param bool $templated
+     *   True if the link object should be templated, False otherwise.
+     * @return static
+     */
+    public function withTemplated($templated);
+
+    /**
+     * Returns an instance with the specified relationship included.
+     *
+     * If the specified rel is already present, this methid MUST return
```
jeroenseegers: s/methid/method/

proposed/links.md
```diff
@@ -162,14 +184,92 @@ interface LinkInterface
...
+     * normally without errors, but without adding the rel a second time.
+     *
+     * @param string $rel
+     *   The relationship value to add.
+     * @return static
+     */
+    public function withRel($rel);
```
jeroenseegers: Why not use the full `withRelationship($relationship)` here? It seems like a style mismatch in relation (hehe) to the other method names and arguments which are spelled out in full.

Same goes for the below `withoutRel($rel)` method.

Crell: The main interface has getRel(), so it makes sense to be consistent with that.  If we were to change it, we'd need to change both.

We went with "rel" primarily because that's what implementations tend to use, if anything.  Both HTML and HTTP Link headers use "rel", not "relationship".  The wisdom of that decision is debatable, but it is what it is, and since that property maps directly to those representations it makes sense to use the same name throughout.

proposed/links.md
Crell: Oops.  Fixed.

proposed/links.md
```diff
@@ -162,14 +184,92 @@ interface LinkInterface
...
+    /**
+     * Returns an instance with the specified attribute added.
+     *
+     * @param string $attribute
+     *   The attribute to include.
+     * @param string $value
+     *   The value of the attribute to set.
+     * @return static
+     */
+    public function withAttribute($attribute, $value);
```
weierophinney: Shouldn't we also have a `withoutAttribute($attribute)` method? (Considering we have a `withoutRel()` method...)

proposed/links.md
```diff
@@ -194,3 +294,26 @@ interface LinkCollectionInterface
...
+interface EvolvableLinkCollectionInterface extends LinkCollectionInterface
+{
+    /**
+     * Returns an instance with the specified link included.
+     *
+     * @param LinkInterface $link
+     *   A link object that should be included in this collection.
+     * @return static
+     */
+    public function withLink(LinkInterface $link);
```
weierophinney: Should we provide a `withoutLink()` method?

michaelcullum:
👍 to providing `without...` methods for consistency.

proposed/links.md
Crell: Ugh. I could have sworn I added one.  Will add.

PR #789, opened 2016-08-02
Rename PSR-13 interfaces to not be http specific

michaelcullum:


Crell:
Please wait on this until #788 is in, since it adds 2 more interfaces that would need to be updated.

michaelcullum:
Ah sure.

michaelcullum:
Rebased on top, see https://github.com/Crell/fig-standards/compare/link-evolvable...michaelcullum:psr-13-domain for diff.

PR #793, opened 2016-08-03
Sync with php-fig/http-message#50

weierophinney:
This patch updates PSR-7 to always use `@return static` instead of `@return self` within method annotations, as done in php-fig/http-message#50. Per the discussion on that issue, `static` is interpreted by phpDoc and most IDEs in such a way that it always resolves to the class against which it was invoked, and not the defining class, unlike `self`, which always resolves to the class in which it was originally defined. Additionally, `static` has the *connotation* that a new instance, not the same instance, will be returned — which is the intention as outlined already in each method providing such a return value.

### DIFF
--- a/accepted/PSR-7-http-message.md
+++ b/accepted/PSR-7-http-message.md
@@ -631,7 +631,7 @@ interface MessageInterface
      * new protocol version.
      *
      * @param string $version HTTP protocol version
-     * @return self
+     * @return static
      */
     public function withProtocolVersion($version);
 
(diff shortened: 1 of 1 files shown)


michaelcullum:
This now has conflicts, sorry @weierophinney

weierophinney:
@michaelcullum I've rebased and re-pushed now.

michaelcullum:
Thanks. I'll merge this and the other tomorrow pending no objections.

PR #803, opened 2016-08-15
[PSR-15] Remove StackInterface

mnapoli:
Related discussion: https://groups.google.com/forum/#!topic/php-fig/GTm9ho6rFts


michaelcullum:
@php-fig/psr-15

shadowhand:
👎 At this time, I do not agree with this change. See http-interop/http-middleware#3 for an alternative.

I would however support removal of the `with*()` methods to allow for a more correct separation of concerns. Additional documentation would need to be made regarding the correct type hinting that is required for properly mixing client/server middleware.

PR #806, opened 2016-08-15
Method versus function naming consistency change

elliottb:
The existing example of a class in section 1.1 includes the method `sampleFunction`, but this should be named `sampleMethod` to be consistent with the class and method language used throughout the rest of this document and [PSR-1](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-1-basic-coding-standard.md#43-methods).


michaelcullum:
👍 

@php-fig/psr-1-2

michaelcullum:
Pinging @php-fig/psr-1-2 (@pmjones) again?

pmjones:
Thanks for this!

PR #814, opened 2016-09-09
Remove PMJ as PSR-8 sponsor

michaelcullum:
Remove @pmjones as PSR-8 sponsor after he resigned


michaelcullum:
@php-fig/secretaries

michaelcullum:
👍

PR #816, opened 2016-09-21
[PSR-5] Corrected ToC anchor links

JeffMatson:
Several of the Table of Contents anchor links were incorrect. Just a quick adjustment.


michaelcullum:
This PSR is without Editor but this isn't a content change so 👍

michaelcullum:
👍

PR #817, opened 2016-09-23
fix incorrect return type-hints

mindplay-dk:
Return type is `void`, not `null` - these are incorrectly hinted and static analysis strictly should fail. (and does, at least in Php Storm.)

### DIFF
--- a/accepted/PSR-3-logger-interface.md
+++ b/accepted/PSR-3-logger-interface.md
@@ -169,7 +169,7 @@ interface LoggerInterface
      *
      * @param string $message
      * @param array $context
-     * @return null
+     * @return void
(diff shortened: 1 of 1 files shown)


michaelcullum:
As per php-fig/log#49 this should go to the ML first

mindplay-dk:
My PR against the interface package itself was accepted, so this is just out of date now, and probably should be merged?

gsdevme:
There is an outstanding reason why this isn't merged?

Jean85:
Since it's just to get in sync with the interface yes, this can be merged as a typo! :+1: 
Thanks for bumping this up!

PR #823, opened 2016-10-17
Remove scalar lowercase rules from unrelated sections

cryode:
The rules regarding capitalization of scalar type hints and reserved keywords are specifically mentioned in [Section 2.5](https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#25-keywords-and-types). Repeating this standard in other sections related to other style guides seems redundant. If there is an argument for doing so that I have missed in the discussion group, I'd love to hear about it.

Also, specifically the text "The declaration keyword (e.g. string) MUST be lowercase." located in [Section 4.5](https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#45-method-and-function-arguments) does not specify a scalar type or keyword, and could be interpreted as the specified style for _any_ return type hint.

### DIFF
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -117,6 +117,8 @@ PHP [keywords][] MUST be in lower case.
 The PHP types and keywords `array`, `int`, `true`, `object`, `float`, `false`, `mixed`,
 `bool`, `null`, `numeric`, `string`, `void` and `resource` MUST be in lower case.
 
+Any new types and keywords added to future PHP versions MUST be in lower case.
+
 Short form of type keywords MUST be used i.e. `bool` instead of `boolean`,
 `int` instead of `integer` etc.
 
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -456,8 +456,6 @@ MUST be one space after each comma.
 Method and function arguments with default values MUST go at the end of the argument
 list.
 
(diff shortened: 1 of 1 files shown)


samdark:
Makes sense.

cryode:
@michaelcullum That change could be considered a separate PR, since the proposed changes in this PR are not related to Section 2.5. Is there a preference or precedence for how SRP changes should be? I have no problem adding it here, if that's the preferred way.

michaelcullum:
It's most certainly related to this PR IMO. Currently, the document text states that any return type hint or parameter type hint must be lowercase which is a catch-all, including for non-mentioned/future types; you're removing those statements so the only statement left is the one where they are all explicitly mentioned, if you wish to remove those statements, you should add a catch-all to 2.5.

michaelcullum:
@cryode Are you willing to make those changes?

cryode:
@michaelcullum Yes! Thank you for the reminder, this completely slipped my memory. I'll do my best to get to that today.

KorvinSzanto:
@cryode I agree with @michaelcullum, if you can get those changes in there I think we'll be good to merge.

cryode:
I'm disappointed in myself for not taking care of such an easy change a long time ago. I appreciate the reminder - please let me know if any language adjustments are recommended.

KorvinSzanto:
Looks like we still need to get a thumbs up from @michaelcullum

dragonmantank:
+1 on this from me

PR #825, opened 2016-10-18
Less ambiguous final line ending

judgej:
The original text was ambiguous in stating that a file must end with "a single line, containing only a single newline (LF) character". Issues with that are:
- A line will only ever contain one LF character, since the LF is what defines the end of the line. So there is an implication the term "single newline" is hinting at something else the author meant.
- The intent of this is to ensure:
  - The file does not end with a string of blank lines. In fact, it does not end with ANY blank lines. That was the intent behind the original PSR-2 when it stated all files "MUST end with a single blank line" - it actually meant a single LF.
  - Every line has a LF line ending. The LF line ending and everything before it (up to the previous LF or the start of the file) is what defines a line.
- This standard should be looking at lines from the file POV, and not how an editor may present it. Some editors may show a blank line on the end, and some may not. It all depends on how they choose to position the cursor when going beyond the final LF.

~~I think perhaps the statement "All PHP files MUST use the Unix LF (linefeed) line ending" is now surplus, and can be removed? The "Every line MUST have a single LF (linefeed) line ending" statement kind of says the same thing, but looks at it from a _line_ POV and not a _file_ POV. For example, a file could still use LF line endings, and yet never put a LF on the last line (older DOS files were like this) while stateing that every line MUST end in a LF makes sure the file both uses LFs _and_ will always end in a LF.~~ Okay, merged these.

### DIFF
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -84,7 +84,7 @@ each word is capitalised including the very first letter.
 
 All PHP files MUST use the Unix LF (linefeed) line ending, applying to every line in the file.
 
-All PHP files MUST end with a non-blank line, terminated with a single LF like all other lines in the file.
+All PHP files MUST end with a non-blank line, terminated with a single LF.
 
 The closing `?>` tag MUST be omitted from files containing only PHP.
 
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -82,9 +82,9 @@ each word is capitalised including the very first letter.
 
 ### 2.2 Files
 
-All PHP files MUST use the Unix LF (linefeed) line ending.
+All PHP files MUST use the Unix LF (linefeed) line ending, applying to every line in the file.
 
-All PHP files MUST end with a single line, containing only a single newline (LF) character.
+All PHP files MUST end with a non-blank line.
 
 The closing `?>` tag MUST be omitted from files containing only PHP.
 
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -84,7 +84,7 @@ each word is capitalised including the very first letter.
 
 All PHP files MUST use the Unix LF (linefeed) line ending, applying to every line in the file.
 
-All PHP files MUST end with a non-blank line.
+All PHP files MUST end with a non-blank line, terminated with a single LF like all other lines in the file.
 
 The closing `?>` tag MUST be omitted from files containing only PHP.
 
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -82,7 +82,7 @@ each word is capitalised including the very first letter.
 
 ### 2.2 Files
 
-All PHP files MUST use the Unix LF (linefeed) line ending, applying to every line in the file.
+All PHP files MUST use the Unix LF (linefeed) line ending only.
 
 All PHP files MUST end with a non-blank line, terminated with a single LF.
 


judgej:
My assumption here is that the intent of PSR-2 to NEVER end a PHP file with a blank line still applies here, i.e. that this requirement has not changed, even if it was ambiguous in PSR-2.

proposed/extended-coding-style-guide.md
```diff
@@ -82,9 +82,9 @@ each word is capitalised including the very first letter.
 
 ### 2.2 Files
 
-All PHP files MUST use the Unix LF (linefeed) line ending.
+All PHP files MUST use the Unix LF (linefeed) line ending, applying to every line in the file.
```
samdark: This is OK.

proposed/extended-coding-style-guide.md
```diff
@@ -82,9 +82,9 @@ each word is capitalised including the very first letter.
...
 
-All PHP files MUST end with a single line, containing only a single newline (LF) character.
+All PHP files MUST end with a non-blank line.
```
samdark: This is not OK since "non-blank" isn't "empty but containing only a single newline (LF) character". Non-blank matches just a line with some code in it.

samdark: Oh, right. Then how about `ending with a single newline (LF) character`

judgej:
> This is not OK since "non-blank" isn't "empty but containing only a single newline (LF) character". Non-blank matches just a line with some code in it.

Yes, that is exactly right. The last line of the file must be non-blank, just as you define it.

A file with *nothing* in it strikes me as a perfectly value PHP file, and that does not end with a `LF` since it has no lines. It might be better to say that the last character cannot be anything other than a `LF`. But then are we forcing the file to be an ASCII or UTF-8 encoded? I suppose it is important to note these are characters and not bytes.

It's really about defining at the right level of abstraction. If the spec states that a file must end in a single `LF` (i.e. not more than one continuous `LF`s) then the file is being described in the context of a *stream*, i.e. totally ignoring what line structure the file may have, we are describing what the last character must be, and what the character before that must NOT be.

The way I look at a test file (and am describing it here), is abstracted into *lines*. Each line contains zero or more characters and is terminated by a `LF`. That's it. Define a line and a file is then constructed from (arguably) zero or more lines. If there is no `LF` at the end of the file, then by definition, the last *line* is not valid. If there are two `LF`s at the end of the file, then *by definition* the last line is blank, so we say the last line cannot be blank.

Looking at the editor image from @michaelcullum the *last line* is `}<LF>`. That is line 35. What the editor shows as line 36 is imaginary - it is not a line in the file - it does not exist. The editor is just giving you somewhere to put a cursor for when you *do* want to create a line 36.

judgej:
Similar with other editors, e.g. Editplus on Windows:

clipboard06

Line 492 is the last line when defining a line as ending with a `LF` (if we defined lines as being *separated* by `LF`s, we would be in an entirely different place).

Just a note on how PHP interprets line endings outside of the `<?php ?>` tags: PHP will just steam all content through, with line endings exactly as they are in the file, *with the exception* of the last `LF` on the last line if the file ends with `?>LF`. It makes a special exception for that last `LF` on that last line. I'm sure early versions of PHP did not do that, and you had to be very careful to make sure the closing `?>` was not `LF` terminated, back in the olden days when still used a final closing `?>`. Anyway, that's a diversion, but hopefully helps with some context.

michaelcullum:
@judgej I know and agree, I'm just saying we should have an explicit note to prevent confusions/misunderstandings such as the one @samdark made.

> We should add an explicit note that the last line in the file should be a non-blank line ending in a newline character (which would, therefore, make it appear as though the last line in the file is a blank line in an editor).

judgej:
IMO noting what a line may look like in an editor is something for the meta document, since it gives some subjective background to what people may see in their tools. It is subjective because different people will interpret what they see in their tools in different ways.

An explicit note to state that the last line must not be blank, and must be terminated with a [single] `LF` like any other line in the file (i.e. it it no more special or exceptional than any other line in the file).

The PSR does need to be accessible to people using it - which I guess is where this is coming from - but it is also arguably a set of standards that feed into compliance tools. When PSR-2 first came out and stated that the files needed to end with a single blank line, I did just that - added a single blank line to the ends of my PHP scripts. The PSR-2 rules in travis then told me that was wrong. This is what I am trying to avoid (we know what it *says* but we also know what it does not quite *mean* that), so I strongly believe we need to avoid the mention of a blank line at the end of the file; there is no blank line at the end of the file. There may be a numbered line in your editor, but that is still no blank line in the file by any stretch if we are defining a line as zero or more characters terminated by a `LF`.

So how about:

> All PHP files MUST end with a non-blank line, terminated with a single `LF` like any other line in the file.

Then we explain how that may appear in some tools in the meta document.

Here is a screenshot from `vi` in a putty terminal:

clipboard07

There are two lines of five characters here, both lines ending in `LF`. The file is 12 bytes long. I did not add that final `LF` on that last line - `vi` did that, because that is *the Unix way*. Notice it is clear where the file ends - it does not try to add that fake extra "blank" line on the end. That is a GUI invention, giving you somewhere to click the mouse and start typing. If I show lines numbers, it shows me line 1 and line 2. There is no line 3. If I edit the same file in a GUI editor, it may show me a fake line 3. The whole point of this is that the PSR standard should be talking about the standard that applies to the file, and not the tools. because tools vary a lot.

judgej:
Yikes. Why do they make that "comment and close" button so easy to click?

judgej:
How are we with this? I left it with the suggestion that we use this statement to clarify how the file should end:

> All PHP files MUST end with a non-blank line, terminated with a single LF like any other line in the file.

How it looks in your favourite editor really depends on your favourite editor, and belongs IMO in a meta document. This PSR should be about the standard, end end result, and not a guide on how to use your tools.

By requiring a file to end in a non-blank line, it means empty files (either "" or "\n") are not allowed. I'm assuming that is not going to be a problem.

samdark:
@judgej that is much better. Can you edit the pull request to state exactly that?

proposed/extended-coding-style-guide.md
```diff
@@ -82,9 +82,9 @@ each word is capitalised including the very first letter.
 
 ### 2.2 Files
 
-All PHP files MUST use the Unix LF (linefeed) line ending.
+All PHP files MUST use the Unix LF (linefeed) line ending, applying to every line in the file.
 
-All PHP files MUST end with a single line, containing only a single newline (LF) character.
+All PHP files MUST end with a non-blank line, terminated with a single LF like all other lines in the file.
```
schnittstabil: I would suggest to remove `like all other lines in the file.`, for two reasons:
1. `like` always provide room for interpretation (the word _like_ is about similarity).
2. `like all other lines in the file` forces the reader to reread every single rule about `lines in the file`.

judgej: I tend to agree. My original approach was based on:

* Defining a <line> (zero or more non-<newline> characters followed/terminated by a <newline>)
* Stating a file consists of zero or more <line>s
* Defining a blank <line>
* Stating that a blank line cannot go on the end of the file.

We can show what that looks like in various editors in the metadoc.

proposed/extended-coding-style-guide.md
schnittstabil: @judgej What do you think about this version:

```diff
-All PHP files MUST use the Unix LF (linefeed) line ending.
+All PHP files MUST use the Unix LF (linefeed) line ending only.
```

proposed/extended-coding-style-guide.md
samdark: Agree. Let's remove it.

proposed/extended-coding-style-guide.md
samdark: Good suggestion.

samdark:
@KorvinSzanto, @rdeutz is it good to merge?

michaelcullum:
👍

PR #826, opened 2016-10-18
Compound namespaces with a depth of *three* or more MUST not be used.

judgej:
It says:

> Compound namespaces with a depth of two or more MUST not be used.

Then gives some examples of compound namespaces with a depth of two, stating that they are valid. If they really are valid, then the depth that cannot be used is *three* or more. This is a one-indexed reality and not zero-indexed ;-)

### DIFF
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -178,7 +178,7 @@ class FooBar
 
 ~~~
 
-Compound namespaces with a depth of two or more MUST not be used. Therefore the
+Compound namespaces with a depth of three or more MUST not be used. Therefore the
 following is the maximum compounding depth allowed:
 ~~~php
 <?php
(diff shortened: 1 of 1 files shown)


judgej:
Hey, I like the way github now looks after the fork when creating typo-PRs through the GUI. It used to force you to create the fork in your own account, which was a pain if you wanted multiple forks, and then it inevitably was left hanging around for far too long. I'm rambling...

samdark:
You're correct about the fact that text contradicts examples.

proposed/extended-coding-style-guide.md
```diff
@@ -178,7 +178,7 @@ class FooBar
 
 ~~~
 
-Compound namespaces with a depth of two or more MUST not be used. Therefore the
+Compound namespaces with a depth of three or more MUST not be used. Therefore the
```
michaelcullum: Can we rephrase this to `with a depth of more than two`

judgej:
@michaelcullum yes, that is a much clearer way of putting it.

michaelcullum:
👍

PR #828, opened 2016-10-26
Update container.md

cebe:
fixed formatting of the composer packages and version.

### DIFF
--- a/proposed/container.md
+++ b/proposed/container.md
@@ -97,9 +97,9 @@ into the same container (or another container) instead of the delegate container
 The interfaces and classes described as well as relevant exception are provided as part of the
 [psr/container](https://packagist.org/packages/psr/container) package. (still to-be-created)
 
-Packages providing a psr container implementation should declare that they provide psr/container-implementation 1.0.0
+Packages providing a psr container implementation should declare that they provide `psr/container-implementation` `1.0.0`.
 
-Projects requiring an implementation should require psr/container-implementation 1.0.0
+Projects requiring an implementation should require `psr/container-implementation` `1.0.0`.
 
 2. Interfaces
 -------------


proposed/container.md
```diff
@@ -97,9 +97,9 @@ into the same container (or another container) instead of the delegate container
 The interfaces and classes described as well as relevant exception are provided as part of the
 [psr/container](https://packagist.org/packages/psr/container) package. (still to-be-created)
```
cebe: `(still to-be-created)` should be removed too as the package already exists, right?

samdark: Yes.

proposed/container.md
```diff
@@ -97,9 +97,9 @@ into the same container (or another container) instead of the delegate container
...
 
-Packages providing a psr container implementation should declare that they provide psr/container-implementation 1.0.0
+Packages providing a psr container implementation should declare that they provide `psr/container-implementation` `1.0.0`.
```
weierophinney: I'm not sure I agree with this formatting change; it feels unnecessary. Neither package names nor versions need to be formatted as code in any style guide I've ever used. Can you explain your justification for the change, please?

cebe: Well, at least the sentence should have a `.` at the end, that was the original reason for a change. After adding a `.` it would be `1.0.0.` and I thought it would be better to separate the version string from the last `.` as they do not belong together.
And as the package name and version is something that will be put into a composer.json file, I'd consider it as formatted code.

michaelcullum:
@php-fig/psr-11 @weierophinney ^

weierophinney:
👍

michaelcullum:
Can someone merge/fix commit messages on merge please? :)

samdark:
@michaelcullum GitHub can.

michaelcullum:
@samdark I know, I was saying to whoever merged but Matthew just asked me to do the merge in private.

samdark:
Ah, OK :)

PR #829, opened 2016-11-01
Add default value parameter to get()

digitalnature:
`->get()` should have a fallback argument for convenience, to avoid additional isset() calls / null checks when using this method.


michaelcullum:
@php-fig/psr-16

Seldaek:
Closing as I included this PR's commit in #838

PR #832, opened 2016-11-05
Specify that ContainerInterface::has() takes a string

mnapoli:
See https://groups.google.com/d/msg/php-fig/L8rDUwRFsOU/R2f9-TKYCAAJ

> Small point: should the has() section in 1.1 say that the entry identifier MUST be a string, similar to how it does so for get()?

### DIFF
--- a/proposed/container.md
+++ b/proposed/container.md
@@ -32,7 +32,7 @@ Users of dependency injections containers (DIC) are referred to as `user`.
   While `ContainerInterface` only defines one mandatory parameter in `get()`, implementations
   MAY accept additional optional parameters.
 
-- `has` takes one unique parameter: an entry identifier. It MUST return `true`
+- `has` takes one unique parameter: an entry identifier.  It MUST be a string. It MUST return `true`
   if an entry identifier is known to the container and `false` if it is not.
   `has($id)` returning true does not mean that `get($id)` will not throw an exception.
   It does however mean that `get($id)` will not throw a `NotFoundExceptionInterface`.


michaelcullum:
@php-fig/psr-11

proposed/container.md
```diff
@@ -32,7 +32,7 @@ Users of dependency injections containers (DIC) are referred to as `user`.
   While `ContainerInterface` only defines one mandatory parameter in `get()`, implementations
   MAY accept additional optional parameters.
 
-- `has` takes one unique parameter: an entry identifier. It MUST return `true`
+- `has` takes one unique parameter: an entry identifier.  It MUST be a string. It MUST return `true`
```
schnittstabil: Hmm, is a `has` implementation still PSR compatible, if it additionally takes identifiers of different types, e.g. `string` and `integer`?

mnapoli: I've pushed an update. @schnittstabil `has()` accepts only a string

proposed/container.md
```diff
@@ -32,7 +32,7 @@ Users of dependency injections containers (DIC) are referred to as `user`.
   While `ContainerInterface` only defines one mandatory parameter in `get()`, implementations
   MAY accept additional optional parameters.
 
-- `has` takes one unique parameter: an entry identifier. It MUST return `true`
+- `has` takes one unique parameter: an entry identifier, that MUST be a string. `has` MUST return `true`
```
weierophinney: `s/that/which/`: "an entry identifier, *which* MUST be a string." Otherwise, LGTM.

weierophinney: As @kynx pointed out, Should also likely change to:

```markdown
`has` MUST return a boolean value.
```

weierophinney: Perfect - @mnapoli ^^^^^

mnapoli:
"has must return true" is just the beginning of the next sentence (which continues on the next line).

Please read the Markdown diff to get a better idea. I've applied the suggested changes, I've also synced the description of get.

https://github.com/php-fig/fig-standards/pull/832/files?short_path=30123dd#diff-30123dd583e2151fb660404c99ea8fba

mnapoli:
@weierophinney since you requested changes on this PR I think we'll be waiting on your approval (the PR has been updated since).

PR #835, opened 2016-11-10
Clarify closing PHP tag is optional

JoeMurray:


samdark:
PSR-1 is already accepted so it can't be modified. Next coding style guide is PSR-12 and it's [already handled there](https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md):

```
2.2 Files
...
The closing `?>` tag MUST be omitted from files containing only PHP.
```

JoeMurray:
I noticed PSR-2 also handles this, but didn't realize (duh!) these standards are not living documents.

PR #838, opened 2016-11-13
SimpleCache adjustments

Seldaek:
- Added default return value to `get()`
- Removed TTL_ constants
- Clarified wording in a couple places

With this I think we are good for a final review.. /cc @dragoonis any comments?

### DIFF
--- a/proposed/simplecache.md
+++ b/proposed/simplecache.md
@@ -182,16 +182,17 @@ interface CacheInterface
 
     /**
      * Identify if an item is in the cache.
-     * NOTE: It is recommended that exists() is only to be used for cache warming type purposes
+     *
+     * NOTE: It is recommended that has() is only to be used for cache warming type purposes
      * and not to be used within your live applications operations for get/set, as this method
-     * is subject to a race condition where your exists() will return true and immediately after,
+     * is subject to a race condition where your has() will return true and immediately after,
      * another script can remove it making the state of your app out of date.
      *
      * @param string $key The cache item key
      *
      * @return bool
      */
-    public function exists($key);
+    public function has($key);
 }
 ```
 
--- a/proposed/simplecache.md
+++ b/proposed/simplecache.md
@@ -114,8 +114,8 @@ interface CacheInterface
     /**
      * Fetch a value from the cache.
      *
-     * @param string $key The unique key of this item in the cache
-     * @param mixed $default Default value to return if the key does not exist
+     * @param string $key     The unique key of this item in the cache
+     * @param mixed  $default Default value to return if the key does not exist
      *
      * @return mixed The value of the item from the cache, or $default in case of cache miss
      */
@@ -124,10 +124,11 @@ interface CacheInterface
     /**
      * Persist data in the cache, uniquely referenced by a key with an optional expiration TTL time.
      *
-     * @param string $key The key of the item to store
-     * @param mixed $value The value of the item to store
-     * @param null|int|DateInterval $ttl Optional. The TTL value of this item. If no value is sent and the driver supports TTL
-     *                                       then the library may set a default value for it or let the driver take care of that.
+     * @param string                $key   The key of the item to store
+     * @param mixed                 $value The value of the item to store
+     * @param null|int|DateInterval $ttl   Optional. The TTL value of this item. If no value is sent and
+     *                                     the driver supports TTL then the library may set a default value
+     *                                     for it or let the driver take care of that.
      *
      * @return bool True on success and false on failure
      */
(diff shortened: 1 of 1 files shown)


Seldaek:
BTW https://github.com/php-fig/simplecache/tree/master/src is up to date with this PR and I also added the PSR-6 bridge class back in there.

unknownbliss:
It was decided that the PSR-6 bridge should go with cach-utils and PSR-16 itself would just be it's own interface designed so a bridge is possible no?

Seldaek:
So you would place that in the fig/cache-util package? I think it adds complexity for nothing really as one would have to require psr/simple-cache + fig/cache-util + a psr-6 implementation to start working with it. It's one simple class that doesn't have to change ever.. and it coming with psr/simple-cache means if you get that and a psr-6 implementation you can start using simple-cache interfaces easily. I don't have very strong feelings about this though.. Just strikes me as less useful in the other package than having it in with the interfaces.

michaelcullum:
fig/cache-utils will require both psr-16 and psr-6 so all you would have to require is fig/cache-utils theoretically.

Adding this bridge to psr/simple-cache means that PSR-16 would require PSR-6 https://github.com/php-fig/simplecache/blob/master/src/Psr6Bridge.php#L5.

The psr/simple-cache repo should really just contain the interfaces/stuff defined in the spec. Utils can essentially have whatever we like (and is appropriate) in it and can have bug fixes applied if ever necessary.

Also, as a general note, should we build a bridge for the other direction (interface with it in a PSR-6 type way to interact with a PSR-16 library).

Crell:
I concur with @michaelcullum.  The spec itself should be interface-only, or interfaces-and-constants maybe.  Even partial implementations do not belong in the specification itself.  That's what the -util repos are for.

Someone implementing PSR-6 may or may not be using cache-utils for it.  That's fine.  Someone implementing PSR-16 may or may not be building it on PSR-6.  That's also fine.  If they are, and want to use the provided bridge, then there's decent odds they'll want the helper code that's in the rest of the cache-utils library anyway.  (And if not, it's just a few dead files on disk, no biggie.)

proposed/simplecache.md
```diff
@@ -111,26 +111,24 @@ namespace Psr\SimpleCache;
 
 interface CacheInterface
 {
-    const TTL_MINUTE = 60;
-    const TTL_HOUR = 3600;
-    const TTL_DAY = 86400;
```
Crell: Why remove these? They actually seem like they'd be useful.

Seldaek: Because they are kinda useless and someone suggested dropping them.. Any dev worth their salt probably knows 60/3600/86400 by heart.

I don't see the rationale of keeping this but dropping the Bridge for example. Note that the bridge is not in the spec, it's just in the package, I get that cache-util was made a separate package as it contains a bunch of code, but in this case it's *one* class with a few lines of code.

Seldaek:
Ah well, I am not going to fight for the bridge, I don't really care, and I suspect implementors will do the work of implementing PSR-16 themselves anyway if it's got any success. But I don't see the point in having it bundled in cache-util which is all PSR-6 specific, and suddenly it would require PSR-16 too which means you kinda get both specs if you use either with cache-util.. Doesn't make sense to me.

michaelcullum:
@Seldaek cache-util will include PSR-16 stuff too later on. @dragoonis tells me he has a bunch of stuff for it.

Crell:
Composer supports a "suggests" key, doesn't it. :smile:

michaelcullum:
I've been meaning to add the utils to suggests for psr-6 and psr-16 (I had this thought randomly lying in bed a few nights ago).

Seldaek:
Yeah well as I said, not gonna fight this, I mostly care about the spec, if @dragoonis is cool with it too then I'll remove the bridge again.. (and TMI :p)

dragoonis:
Hello Folks,

a) I've renamed exists() to has() and I’ve deleted PSR-6 bridge from https://github.com/php-fig/simplecache

b) There will be no mixing of PSR-16 utils and existing PSR-6 utils repo. They will be separate packages and individually have their own `suggest` packagist entry. There will be clear isolation of standards and as such a separate `fig/simplecache-utils` will be created and maintained separately.

c) @Seldaek will be updating this PR with the 2 commits I pushed to php-fig/simplecache just now.

I will be updating the mailing list shortly thereafter about next steps.

Seldaek:
OK all up to date here..

dragoonis:
I have approved this PR. I need someone else in the organisation to press approve so that I may merge it.

michaelcullum:
@dragoonis If you reply with `:+1:` then it approves the PR and makes it mergeable. :)

Seldaek:
👍 then.. why is this not integrated with the GH review process :(

michaelcullum:
Short version: We've had it longer than github has had approvals.

Long version: They keep assuring me they're working on integrating that, we would disable it but most people know how it works now so it's not a huge issue and it gives us some nice functionality on different files requiring approval from different groups. It first was added after people started merging stuff to other PSRs against Editor's wishes and merging changes that shouldn't be merged to approved PSRs and people were getting annoyed with other people because of it.

dragoonis:
Good to know, I will :+1: things in the future that I will merge. Thanks.

PR #842, opened 2016-11-16
Update simplecache.md

SpacePossum:
- Scoping down the `value` as Resources and things like lambda's won't fit in a cache ;) 
- Update some return hints to `false` since `bool` is to wide for no reason IMHO (as `true` is never returned).
- Added some notes about edge cases

Not sure about the semantics about `delete` not returning a `bool` to state if success or not while `setters` do.

Lets wrap this up and ship it! : D)

### DIFF
--- a/proposed/simplecache.md
+++ b/proposed/simplecache.md
@@ -127,7 +127,7 @@ interface CacheInterface
      * Persist data in the cache, uniquely referenced by a key with an optional expiration TTL time.
      *
      * @param string                $key   The key of the item to store
-     * @param mixed                 $value The value of the item to store
+     * @param mixed                 $value The value of the item to store, must be serializable
      * @param null|int|DateInterval $ttl   Optional. The TTL value of this item. If no value is sent and
      *                                     the driver supports TTL then the library may set a default value
      *                                     for it or let the driver take care of that.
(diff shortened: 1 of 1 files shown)


michaelcullum:
@php-fig/psr-16

proposed/simplecache.md
```diff
@@ -212,24 +212,26 @@ interface CounterInterface
     /**
      * Increment a value atomically in the cache by the given step value and return the new value
      *
-     * If the key does not exist, it is initialized to the value of $step
+     * If the key does not exist, it is initialized to the value of $step.
+     * If the value is increased above PHP_INT_MAX the return value is undefined.
```
samdark: Isn't it better to throw exception in this case?

proposed/simplecache.md
```diff
@@ -212,24 +212,26 @@ interface CounterInterface
...
+     * @return int|false The new value on success and false on failure
      */
     public function increment($key, $step = 1);
 
     /**
      * Decrement a value atomically in the cache by the given step value and return the new value
      *
-     * If the key does not exist, it is initialized to the value of -$step
+     * If the key does not exist, it is initialized to the value of -$step.
+     * If the value is decreased below PHP_INT_MIN the return value is undefined.
```
samdark: Isn't it better to throw exception in this case?

proposed/simplecache.md
```diff
@@ -127,7 +127,7 @@ interface CacheInterface
      * Persist data in the cache, uniquely referenced by a key with an optional expiration TTL time.
      *
      * @param string                $key   The key of the item to store
-     * @param mixed                 $value The value of the item to store
+     * @param mixed                 $value The value of the item to store, must be serializable
```
samdark: Yes, that's a good addition.

proposed/simplecache.md
SpacePossum: The problem will be to detect the cases since not all cache implementations deal with this the same way. Checking it on each call will create overhead without much gain. Therefor I'm in fav. that these edge cases should be handled at application level if needed.

Jean85: Also, a cache miss shouldn't break your application, as remarked on the ML.

Seldaek: No undefined means the behavior is not defined and maybe you get PHP_INT_MAX if you get an underflow, or you get 0 or false because it got borked, or god knows what the cache backend does :)

Seldaek:
Thanks, overall good cleanups I think :) :+1:

Seldaek:
@michaelcullum can we get this merged when you get a second? I don't know why the :+1: isn't unlocking the merge button this time around :/ Maybe just a delay thing.. will see tomorrow otherwise.

schnittstabil:
@Seldaek: I'm not a PullAprove expert, but the config says:
```
approve_regex: '^:\+1:'
```

Thus I think your comment – or at least a comment line – must start with :+1:, not end with it…

dragoonis:
:+1:

dragoonis:
@seldaek approved. We all happy here ?

dragoonis:
Jordi if you reply with just a thumb it will let you or me, as authors, merge it.

SpacePossum:
Thanks @Seldaek & @dragoonis for all the work on this, hope to see it out there soon! :)
Thanks @ the reviewers and maintainers/members for reviewing my PR.

PR #843, opened 2016-11-17
Updated PSR-11 info in numerical index

weierophinney:
Missed this when I added to to the Review section. This updates the coordinator and sponsor, as well as the status.


michaelcullum:
:+1:

PR #844, opened 2016-11-19
Add MisconfiguredServiceExceptionInterface

Crell:
Clearly distinguish between "missing service", "broken service", and "other" error conditions.

### DIFF
--- a/proposed/container.md
+++ b/proposed/container.md
@@ -45,11 +45,10 @@ Exceptions directly thrown by the container SHOULD implement the
 A call to the `get` method with a non-existing id MUST throw a
 [`Psr\Container\Exception\NotFoundExceptionInterface`](#not-found-exception).
 
-A call to `get` can trigger additional calls to `get` (to fetch the dependencies).
-If one of those dependencies is missing, the `NotFoundExceptionInterface` triggered by the
-inner `get` call SHOULD NOT bubble out. Instead, it should be wrapped in an exception 
-implementing the `ContainerExceptionInterface` that does not implement the 
-`NotFoundExceptionInterface`.
+A cal to the `get` method for a a service that is defined but cannot be instantiated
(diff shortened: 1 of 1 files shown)


unkind:
`MisconfiguredServiceExceptionInterface` is particular case of "something is broken" exception. It makes me wonder: what makes this exception so special that user needs to know about it? I think that from the caller perspective, it doesn't really matter if container is broken or service is misconfigured. The only thing that matters is that the container doesn't provide required stuff and it's not user's fault.

`MisconfiguredServiceExceptionInterface` is classic unchecked exception, so if user doesn't want to bubble out "something is broken" exception, he has to catch `\Exception` or even `\Throwable`.

The real-life example of catching `MisconfiguredServiceExceptionInterface` would be great addition to this PR.

michaelcullum:
@php-fig/psr-11 FYI: This [whether or not to split it into two exceptions] was discussed in depth at the in-person meeting at php[world] and those present were unanimous in agreeing this to be the best path. Notes will be coming out shortly.

cc/ @GeeH

Crell:
"Misconfigured" was the best generic name I could up with, and seems accurate; the service is defined, but "something about it is wrong".  I'm open to other names if there's a consensus on some other term.

mnapoli:
@michaelcullum As said on the ML I'm 👎 unless there's a use case, maybe those notes will help regarding that?

weierophinney:
Per a [comprehensive post to the list from @moufmouf](https://groups.google.com/forum/#!msg/php-fig/2pnhudRUpQg/ewoGoNtFCgAJ), while there may have been consensus at php[world], there have been no concrete demonstrations nor examples of what purpose this new interface serves for consumers; additionally, there's some ambiguity as to whether or not adding this interface might dictate implementation details to container authors (the working group around the specification feel there would be performance and maintenance ramifications).

As such, and on discussion with the working group, I am closing this request.

PR #845, opened 2016-11-21
[PSR-15] HTTP Middleware: Rename DelegateInterface to RequestHandlerInterface

schnittstabil:
@akrabat stated at [the ML](https://groups.google.com/d/topic/php-fig/V12AAcT_SxE/discussion) about the `FrameInterface`:
>  It means nothing to me

And after a long discussion about the pros and cons of having an interface in contrast to a callable, we ended up by the idea of `DelegateInterface` (please forgive me, if I oversimplify things).

The idea, as far as I can see, behind this:
* the middleware container sends a *delegate* of him to a middleware, thus
* the middleware can interact with the *delegate* and doesn't have to know the concrete container

This sounds a bit better, but the interaction is essentially:
```
$response = $delegate->delegate($request);
```

Thus, we end up with the strange oddity: the first `delegate` means something quiet different than the second `delegate`.

---

Therefore, I suggest a rename of `DelegateInterface` to `RequestHandlerInterface`:

```php
interface RequestHandlerInterface
{
    /**
     * Process a request and return the response.
     *
     * @param RequestInterface $request
     *
     * @return ResponseInterface
     */
    public function process(RequestInterface $request);
}
```

The pros, I see:
1. The concept of request handlers is easy to understand and well known.
2. It is a more general concept – the middleware doesn't have to know that it is dealing with a container (and it shouldn't know as @shadowhand mentioned).
3. It reflects the idea, that a middleware comes between a request and response (mentioned @alamin3000).
4. Containers or more precisely dispatchers do exactly the same thing, and MAY implement `RequestHandlerInterface` too.

The cons, I see:
* To be honest: none

**One more thing: This PR is not about `process` vs `__invoke`. Please, don't try to start a new debate about that here.**

### DIFF
--- a/proposed/http-middleware/middleware-meta.md
+++ b/proposed/http-middleware/middleware-meta.md
@@ -32,7 +32,7 @@ provides a number of benefits:
 ## 3.1 Goals
 
 * Create a middleware interface that uses HTTP Messages.
-* Provide a suggested interface for middleware stack containers.
+* Provide a suggested interface for middleware containers.
 * Ensure that middleware will not be tied to a specific implementation of HTTP Messages.
 * Implement a middleware signature that is based on best practices.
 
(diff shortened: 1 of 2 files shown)


shadowhand:
This should be filed against the http-interop/http-middleware repo. These documents are out of date, as per #34.

schnittstabil:
Closing in favor of https://github.com/http-interop/http-middleware/issues/35

PR #846, opened 2016-11-26
Proposal coming from Symfony trial implementation

nicolas-grekas:
I tried to implement the current state of PSR-16 on Symfony Cache to see what it could look like:
https://github.com/symfony/symfony/pull/20636

This has raised a number of issues/questions, namely:

On CacheInterface:
- doesn't say what happens when $key is invalid => the same exception as PSR-6?
- the fact that `getMultiple` returns *all* keys, even cache misses, makes it impossible to (efficiently) implement a PSR-16 to PSR-6 bridge, because it makes it impossible to detect cache misses. When given an array, `apcu_fetch` for example has this other behavior of returning only the "hit", so it doesn't suffer from this. Could this be worth considering?
- accepting `Traversable` for `*Multiple` methods is not consistent with PSR-6 which only takes arrays as arguments
- returning `array` for `getMultiple` is not consistent with PSR-6 `getItems` which returns `array|Traversable`
- some methods return `void` when they could return `bool` => this is both inconsistent with some other methods, and with PSR-6

On CounterInterface:
- the draft doesn't say what happens when $key is invalid => the same exception as PSR-6?
- nor does it say what happens when $step is not an integer => return false? throw something?
- what should happen when $key already exists in the storage but is not "incrementable"? (Redis INCR fails, I didn't check apcu) => return false? throw? erase and store $step? other?
- atomicity misses a normative MUST or SHOULD.

About exceptions:
- if the PSR is going to document when exceptions should be thrown, then it should either define new exception classes or reuse those from PSR-6
- reusing exceptions defined in PSR-6 would look the most sensible to me
- yet, they are currently not in the same namespace. But: couldn't all these new interfaces move in the Psr\Cache namespace (thus releasing them as psr/cache 1.1 on packagist?)

This PR is a proposal to fix all aforementioned notes; namespace renaming excluded for now.

### DIFF
--- a/proposed/simplecache.md
+++ b/proposed/simplecache.md
@@ -31,7 +31,6 @@ standardized streamlined interface for common cases.  It is independent of
 PSR-6 but has been designed to make compatibility with PSR-6 as straightforward
 as possible.
 
-
 ### 1.2 Definitions
 
 Definitions for Calling Library, Implementing Library, TTL, Expiration and Key
@@ -52,7 +51,7 @@ below with whole-second granularity.
 when that item is stored and it is considered stale. The TTL is normally defined
 by an integer representing time in seconds, or a DateInterval object.
 
-*    **Expiration** - The actual time when an item is set to go stale. This it
+*    **Expiration** - The actual time when an item is set to go stale. This is
 calculated by adding the TTL to the time when an object is stored.
 
     An item with a 300 second TTL stored at 1:30:00 will have an expiration of
@@ -82,9 +81,8 @@ supported by implementing libraries: `{}()/\@:`
 *    **Cache** - An object that implements the `Psr\SimpleCache\CacheInterface` interface.
 
 *    **Cache Misses** - A cache miss will return null and therefore detecting
-if one stored null is not possible. This is the main deviation from PSR-6's
+if one stored `null` is not possible. This is the main deviation from PSR-6's
 assumptions.
-
 
 ### 1.3 Cache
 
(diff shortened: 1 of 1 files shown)


dragoonis:
Give me a few days to review this @nicolas-grekas

michaelcullum:
@php-fig/psr-16

Seldaek:
I put all my feedback on the ML.. Thanks for the PR but let's please 
discuss it on ML :)

dragoonis:
@nicolas-grekas I have added InvalidArgumentException class to PSR-16. Can you re-update this PR to reference that class instead of the global php exception class

proposed/simplecache.md
```diff
@@ -155,32 +167,44 @@ interface CacheInterface
...
      *
-     * @return void
+     * @return bool True if the item was successfully removed. False if there was an error.
+     *
+     * @throws InvalidArgumentException
+     *   If any of the keys in $keys are not a legal value a \Psr\Cache\InvalidArgumentException
+     *   MUST be thrown.
      */
-    public function deleteMultiple($keys);
+    public function deleteMultiple(array $keys);
```
dragoonis: remove this array typehint - we're keeping traversable|array

proposed/simplecache.md
```diff
@@ -155,32 +167,44 @@ interface CacheInterface
...
      *
-     * @return array An array of key => value pairs. Cache keys that do not exist or are stale will have a value of null.
+     * @return array|Traversable A list of key => value pairs. Cache keys that do not exist or are stale must not be returned.
+     *
+     * @throws InvalidArgumentException
+     *   If any of the keys in $keys are not a legal value a \Psr\Cache\InvalidArgumentException
+     *   MUST be thrown.
      */
-    public function getMultiple($keys);
+    public function getMultiple(array $keys);
```
dragoonis: remove array, we're sticking with Traversable|array

nicolas-grekas:
@dragoonis PR updated, I reverted the type hints on `*Multiple` methods, although I find it shame that this will force us to write boilerplate code to validate the input while a simple `array` type hint would've done the job perfectly to me.

Side note: PSR-6 has ExceptionInterface. Should PSR-16 have it also?

proposed/simplecache.md
Seldaek: Really? I thought you only wanted it for returns.. I didn't see anyone advocating for Traversable inputs.

Seldaek:
@nicolas-grekas we just added the CacheException yesterday IIRC?

nicolas-grekas:
Indeed, I missed it, my bad!

proposed/simplecache.md
```diff
@@ -114,85 +112,117 @@ namespace Psr\SimpleCache;
...
+     * @param string $key     The unique key of this item in the cache.
+     * @param mixed  $default Default value to return if the key does not exist.
      *
-     * @param string $key     The unique key of this item in the cache
-     * @param mixed  $default Default value to return if the key does not exist
+     * @return mixed The value of the item from the cache, or $default in case of cache miss.
      *
-     * @return mixed The value of the item from the cache, or $default in case of cache miss
+     * @throws InvalidArgumentException
+     *   If the $key string is not a legal value a \Psr\SimpleCache\InvalidArgumentException
```
dragoonis: Can you tweak this to `@throws \Psr\SimpleCache\InvalidArgumentException` and then `If the $key string is not a legal value` ?

proposed/simplecache.md
```diff
@@ -114,85 +112,117 @@ namespace Psr\SimpleCache;
...
+     * @param string                $key   The key of the item to store.
+     * @param mixed                 $value The value of the item to store, must be serializable.
      * @param null|int|DateInterval $ttl   Optional. The TTL value of this item. If no value is sent and
      *                                     the driver supports TTL then the library may set a default value
      *                                     for it or let the driver take care of that.
      *
-     * @return bool True on success and false on failure
+     * @return bool True on success and false on failure.
+     *
+     * @throws InvalidArgumentException
```
dragoonis: same as above. Put the FQCN in the @throws

proposed/simplecache.md
```diff
@@ -114,85 +112,117 @@ namespace Psr\SimpleCache;
...
-     * @param array|Traversable     $items An array of key => value pairs for a multiple-set operation.
+     * @param array|\Traversable    $items A list of key => value pairs for a multiple-set operation.
      * @param null|int|DateInterval $ttl   Optional. The TTL value of this item. If no value is sent and
      *                                     the driver supports TTL then the library may set a default value
      *                                     for it or let the driver take care of that.
      *
-     * @return bool True on success and false on failure
+     * @return bool True on success and false on failure.
+     *
+     * @throws InvalidArgumentException
```
dragoonis: @throws FQCN ..

Then: `If $keys is neither an array nor a Traversable, or if any of the $keys are not a legal value.`

dragoonis:
@nicolas-grekas solid work so far! Just some minor @throws updates and we're ready to merge!

proposed/simplecache.md
nicolas-grekas: updated. I kept the normative MUST which is important to me to force an imperative interpretation (that's why RFC 2119 is for)

proposed/simplecache.md
```diff
@@ -114,85 +112,110 @@ namespace Psr\SimpleCache;
...
+     * @param null|int|DateInterval $ttl    Optional. The TTL value of this item. If no value is sent and
+     *                                      the driver supports TTL then the library may set a default value
+     *                                      for it or let the driver take care of that.
      *
-     * @return bool True on success and false on failure
+     * @return bool True on success and false on failure.
+     *
+     * @throws \Psr\SimpleCache\InvalidArgumentException
+     *   MUST be thrown if $keys is neither an array nor a Traversable,
+     *   or if any of the $keys are not a legal value.
```
Seldaek: Typo on those two lines, should be $values and not $keys

Seldaek:
Apart from that last typo, looks all good to me!

proposed/simplecache.md
nicolas-grekas: fixed

Seldaek:
:+1: Thanks

matthiasmullie:
Can we get an 0.3.0 release with these changes (along with InvalidArgumentException & CounterInterface removal)?

michaelcullum:
Sure. I can do a tag when I get to my laptop shortly.

Seldaek:
FYI @matthiasmullie sent a PR and the 0.3 of the interfaces package is now out, if anyone else is interested :)

michaelcullum:
Sorry about that folks, I totally missed this off my todo list.

matthiasmullie:
Don't worry about it - I should've sent that PR in the first place, instead of expecting someone to change their schedule ;) Thanks!

nicolas-grekas:
Symfony implementation updated, all good to me.
See https://github.com/symfony/symfony/pull/20694

PR #848, opened 2016-11-27
[PSR-16] Adding in SimpleCache CacheException and InvalidArgumentException

dragoonis:


Seldaek:
:+1:

schnittstabil:
According to [PSR-7](https://github.com/php-fig/fig-standards/blob/b4e1163f9f80d8b35c287b73769abfd69f7eafd6/bylaws/007-psr-naming-conventions.md):
> Interfaces MUST be suffixed by `Interface`: e.g. `Psr\Foo\BarInterface`.

Is that outdated?

proposed/simplecache.md
```diff
@@ -197,3 +197,33 @@ interface CacheInterface
...
 ```
+
+### 2.2 CacheException
+```
+namespace Psr\SimpleCache;
+
+/**
+ * Interface used for all types of exceptions thrown by the implementing library.
+ */
+interface CacheException
```
schnittstabil: `interface CacheExceptionInterface`!?

proposed/simplecache.md
```diff
@@ -197,3 +197,33 @@ interface CacheInterface
...
+
+namespace Psr\SimpleCache;
+
+/**
+ * Exception interface for invalid cache arguments.
+ *
+ * When an invalid argument is passed it must throw an exception which implements
+ * this interface
+ */
+interface InvalidArgumentException extends CacheException
```
schnittstabil: `interface InvalidArgumentExceptionInterface extends CacheExceptionInterface`!?

Seldaek:
I don't think this applies to *exception* interfaces? @crell @michaelcullum any clue?

michaelcullum:
Unfortunately it does. This is the main reason why it was raised and voted on recently.

Seldaek:
Wat? How does this make any sense? It even goes against PSR-6.. And you're using this in a catch block, adding Interface suffix there has no use whatsoever?!

michaelcullum:
Indeed.

Re: PSR-6 just after it got voted in and everyone was like 'well it has passed now so too late'

Then the container folks raised it on the mailing list due to https://github.com/php-fig/fig-standards/blob/master/proposed/container.md#22-psrcontainerexceptioncontainerexceptioninterface and then they put it to a vote and people voted to continue having it.

Seldaek:
Please tell me I didn't inadvertently vote on that :(

This has to be the most regrettable nitpick decision the FIG made, but I 
guess that's that..

Seldaek:
I guess we could declare them as exception classes extending from \Exception instead of an interface, but that's probably crippling to some use cases.. Damn, time for bed I guess.

michaelcullum:
Well, you wrote the initial bylaw: https://github.com/php-fig/fig-standards/commit/b834116b94b1ed34fd86c4d6597ddbda8db9858b

And you voted -1
image

Seldaek:
Yeah because it makes sense to have the suffix on interfaces.. And I voted on the tweak -1 because it still made sense to have it on interfaces. The problem is I never foresaw we'd have silly package-level Exception classes, nor did I remember seeing in the discussion that this was about exceptions, I only remember people complaining about having to type-hint against CachePoolInterface for example, which I still think would have been a silly change. Adding clarification about excluding exceptions though would make sense IMO because doing a ` catch (SomethingInterface $e)` is confusing as hell.

schnittstabil:
LOL, I don't like them either. Currently we only have proposed ones, these are:
* `Psr\Container\Exception\ContainerExceptionInterface`
* `Psr\Container\Exception\NotFoundExceptionInterface`

Seldaek:
Then I'd stick to previous art (PSR-6 has no Interface suffix on exceptions, PSR-7 throws plain PHP exceptions) to constitute enough precedent to disambiguate the meaning of this naming conventions bylaw.

moufmouf:
@Seldaek if you want to raise the issue on the list regarding not suffixing the exception interfaces, that would be most welcome. I'll probably abstain from doing it myself since the last vote we asked has been pretty clear-cut.

PR #849, opened 2016-11-28
the param of groupHug should be from type Huggable[]

tflori:
It is not hard failure in annotation but it allows type hints


michaelcullum:
@php-fig/psr-8

TomasVotruba:
:+1:

PR #852, opened 2016-12-01
Remove out of scope text

joshdifabio:
> Different CacheInterface instances MAY be backed by the same datastore, but MUST be logically independent.

I'm of the opinion that infrastructure and _what a cache object represents_ are both outside the scope of what this standard should address, and that this sentence should be removed.

Furthermore, the second clause, which stipulates that "CacheInterface instances ... MUST be logically independent" is needlessly restrictive and prevents many valid use cases, such as decorator objects.

### DIFF
--- a/proposed/simplecache.md
+++ b/proposed/simplecache.md
@@ -109,8 +109,7 @@ to perform, and lets you perform your operations in a single call to the cache s
 times dramatically.
 
 An instance of CacheInterface corresponds to a single collection of cache items with a single key namespace,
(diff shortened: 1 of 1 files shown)


michaelcullum:
@php-fig/psr-16

Seldaek:
Closing as per the discussion on ML. I think Larry explained the goals sufficiently.

joshdifabio:
I still think that the second clause fails to solve any actual problems while having the potential to create problems, but I respect your judgement. Cheers o/

PR #855, opened 2016-12-31
Update data about Container Interface

schlessera:
Changes:

1. The numerical index now reflects that it is in `Review`.
2. The Coordinator and Sponsor are updated.
3. The link points to the `php-fig` repository now.

Resolves https://github.com/php-fig/php-fig.github.com/issues/219


michaelcullum:
👍

PR #858, opened 2017-01-02
Requires an InvalidArgumentException to be thrown on set methods if an invalid $ttl is provided.

naroga:
If the provided $ttl is neither an integer nor a DateInterval, the Cache must throw an InvalidArgumentException.

### DIFF
--- a/accepted/PSR-16-simple-cache.md
+++ b/accepted/PSR-16-simple-cache.md
@@ -166,7 +166,8 @@ interface CacheInterface
      * @return bool True on success and false on failure.
      *
      * @throws \Psr\SimpleCache\InvalidArgumentException
-     *   MUST be thrown if the $key string is not a legal value.
+     *   MUST be thrown if the $key string is not a legal value
+     *   OR if $ttl is neither an integer or a \DateInterval object.
      */
     public function set($key, $value, $ttl = null);
 
@@ -178,8 +179,7 @@ interface CacheInterface
      * @return bool True if the item was successfully removed. False if there was an error.
      *
      * @throws \Psr\SimpleCache\InvalidArgumentException
-     *   MUST be thrown if the $key string is not a legal value 
-     *   OR if $ttl is neither an integer or a \DateInterval object.
+     *   MUST be thrown if the $key string is not a legal value.
      */
     public function delete($key);
 


accepted/PSR-16-simple-cache.md
```diff
@@ -178,7 +178,8 @@ interface CacheInterface
      * @return bool True if the item was successfully removed. False if there was an error.
      *
      * @throws \Psr\SimpleCache\InvalidArgumentException
-     *   MUST be thrown if the $key string is not a legal value.
+     *   MUST be thrown if the $key string is not a legal value 
+     *   OR if $ttl is neither an integer or a \DateInterval object.
```
Seldaek: There is no $ttl here

accepted/PSR-16-simple-cache.md
```diff
@@ -215,7 +216,8 @@ interface CacheInterface
      *
      * @throws \Psr\SimpleCache\InvalidArgumentException
      *   MUST be thrown if $values is neither an array nor a Traversable,
-     *   or if any of the $values are not a legal value.
+     *   if any of the $values are not a legal value
+     *   OR if $ttl is neither an integer or a \DateInterval object.
```
Seldaek: This one is a valid clarification of the docblocks I guess but I don't think the spec can be changed that it's been voted in..

matthiasmullie:
There is also an (unmerged) discussion about typechecking TTLs for PSR-6 (in a different way): https://github.com/php-fig/fig-standards/pull/787/files

accepted/PSR-16-simple-cache.md
naroga: You are right, I pasted my comment above the wrong method. I've fixed this (changed from delete to the set method)

dragoonis:
I have replied onto the ML about such related questions.

accepted/PSR-16-simple-cache.md
michaelcullum: Correct. It can go in as errata though?

michaelcullum:
@dragoonis Do you have a link to that post? Can this be closed or did you / @Seldaek wish to do an errata?

Crell:
This should be covered by errata, not a change to the spec.  See #915

PR #860, opened 2017-01-04
Markdown Syntax Highlighting

dragoonis:


michaelcullum:
👍

schnittstabil:
Just out of interest; as far as I known and according to [the github markdown documentation](https://help.github.com/articles/creating-and-highlighting-code-blocks/#syntax-highlighting), the correct syntax is as follows, isn't it?

```diff
- ``` php
+ ```php

- ~~~php
+ ```php
```

michaelcullum:
Jekyll requires tildes. @reinink Could explain better

schnittstabil:
I believe this should be better fixed by some jekyll configuration; anyway, thanks @michaelcullum.

PR #861, opened 2017-01-10
<?php tag MUST be followed by a blank line

gmponos:
In my opinion I believe that this new PSR should stricly declare if the `<?php` tag MUST or MUST NOT be followed by a blank line.

### DIFF
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -142,7 +142,8 @@ remainder of the code consists a closing PHP tag and then a mixture of HTML and
 PHP.
 
 When the opening `<?php` tag is on the first line of the file, it MUST be on its
-own line with no other statements unless it is a file containing markup outside of PHP
+own line with no other statements, and it MUST be separated by a blank line from
+the following block unless it is a file containing markup outside of PHP
 opening and closing tags.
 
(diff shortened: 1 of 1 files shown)


gmponos:
Hmm... I just found out this [MR](https://github.com/php-fig/fig-standards/pull/690). I made this change based on the example following the line I have changed which has a blank line after the tag. The other examples that do not have a File-level docblock do not seem to have a blank line. 

So how is this decided? Should this be on the judgement of each developer if he will leave a blank line or not?

samdark:
Currently it's on the judgment of each developer.

gmponos:
So what are your thoughts about it? Should be left as is or force a blank/not blank line rule?

samdark:
There are cases where opinions were contradicting each other:

```php
<?php

namespace my\soft;

class X
{
}
```

and

```<?php
declare(strict_types=1);
```

or even

```<?php declare(strict_types=1);
```

That's why that was left out for now.

michaelcullum:
@php-fig/psr-12

michaelcullum:
It does.

gmponos:
> It does.

What does?

michaelcullum:
>[PSR-12] should strictly declare if the <?php tag MUST or MUST NOT be followed by a blank line.

PR #863, opened 2017-01-11
[PSR-13] Minor fix of title levels

petk:
There is a slight inconsistency in the title levels of PSR-13. This patch fixes that.

### DIFF
--- a/accepted/PSR-13-links.md
+++ b/accepted/PSR-13-links.md
@@ -91,7 +91,7 @@ A relationship that is not defined in one of the above registries or a similar
 public registry is considered "private", that is, specific to a particular
 application or use case.  Such relationships MUST use an absolute URI.
 
-## 1.4 Link Templates
+### 1.4 Link Templates
 
 [RFC 6570](https://tools.ietf.org/html/rfc6570) defines a format for URI templates, that is,
 a pattern for a URI that is expected to be filled in with values provided by a client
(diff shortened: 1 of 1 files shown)


accepted/PSR-13-links.md
```diff
@@ -272,7 +272,7 @@ interface EvolvableLinkInterface extends LinkInterface
 }
 ~~~
 
-#### 3.2 `Psr\Link\LinkProviderInterface`
+### 3.3 `Psr\Link\LinkProviderInterface`
```
schnittstabil: As PSR-13 is already accepted, people referring to these numbers; therefore, it would be unfortunate to change these in my opinion.

accepted/PSR-13-links.md
```diff
@@ -306,7 +306,7 @@ interface LinkProviderInterface
 }
 ~~~
 
-#### 3.3 `Psr\Link\EvolvableLinkProviderInterface`
+### 3.4 `Psr\Link\EvolvableLinkProviderInterface`
```
schnittstabil: Same here.

schnittstabil:
Aww, sorry, never mind – 3.2 is used two times etc.

petk:
Probably it is not too much of an issue. So I'm closing this as resolved. The code is available in the separate repository anyway also and links don't break much things. Thanks for your feedback anyhow.

PR #865, opened 2017-01-14
Fixed some small typos and return type-hint of has()

kocsismate:
I found and corrected some small typos in the PSR as well as changed the return type-hint of the `has()` method to be `bool` instead of its alias `boolean` (I found that other PSRs also prefer the shorter form).

Another thing I wanted to change is the style of code blocks (``` instead of ~~~), because the built-in GitHub editor didn't like it very much (the namespace was lined through). Finally, I changed my mind because I found that other PSR-s also use this convention. Do you have any idea why this is the preferred code block style?

### DIFF
--- a/proposed/container.md
+++ b/proposed/container.md
@@ -12,7 +12,7 @@ interpreted as described in [RFC 2119][].
 
 The word `implementor` in this document is to be interpreted as someone
 implementing the `ContainerInterface` in a dependency injection-related library or framework.
-Users of dependency injections containers (DIC) are referred to as `user`.
+Users of dependency injection containers (DIC) are referred to as `user`.
 
 [RFC 2119]: http://tools.ietf.org/html/rfc2119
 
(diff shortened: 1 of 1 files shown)


schnittstabil:
I've asked the same question the other day at https://github.com/php-fig/fig-standards/pull/860#issuecomment-270690519

proposed/container.md
```diff
@@ -12,7 +12,7 @@ interpreted as described in [RFC 2119][].
 
 The word `implementor` in this document is to be interpreted as someone
 implementing the `ContainerInterface` in a dependency injection-related library or framework.
-Users of dependency injections containers (DIC) are referred to as `user`.
+Users of dependency injection containers (DIC) are referred to as `user`.
```
kocsismate: Wouldn't "referred as" be the correct form?

mnapoli:
I have no idea regarding your question :/

Merging because those are typo fixes, thanks

PR #866, opened 2017-01-14
Alternate markdown syntax is a little less buggy.

judgej:
An unordered list followed by an underline heading syntax results in an extra paragraph element wrapping the last list item. The ## heading syntax avoids this.

The change brings Taylor Otwell back into the list, rather than standing off the end like Billy Nomates :-)

Also the triple back-tick ``` syntax for a code section formats correctly in the github code change previewer and editor, while the tilde \~\~\~ syntax does not (it is parsed as a strikeout in some contexts, and a code block in other contexts).


mnapoli:
:+1:

Thank you.

PR #868, opened 2017-01-14
Update PSR-17 proposal with current consensus

shadowhand:


proposed/http-factory/http-factory-meta.md
```diff
@@ -1,5 +1,5 @@
-PSR-17 Meta Document
-====================
+HTTP Factories
```
weierophinney: Add the verbiage "Meta Document" to the header, plase.

proposed/http-factory/http-factory.md
```diff
@@ -34,7 +34,10 @@ in separate classes.
 Has the ability to create client requests.
 
 ```php
-namespace Psr\Http\Message;
+namespace Psr\Http\Factory;
```
weierophinney: Question: where was the discussion behind the new namespace? I'd typically expect these to be in the same namespace, as they're related to the _messages_ in particular; having them in a `Factory` subnamespace might cause conflicts in the future (e.g., if we ever did a "middleware factory" PSR — which I doubt will happen, but it serves as an example).

proposed/http-factory/http-factory.md
```diff
@@ -163,14 +170,17 @@ fwrite($resource, $body);
...
+use Psr\Http\Message\UploadedFileInterface;
 
 interface UploadedFileFactoryInterface
 {
     /**
      * Create a new uploaded file.
      *
-     * If a string is passed it is assumed to be a file path.
+     * If a string is used to create the file, a temporary resource will be
+     * created with the content of the string.
```
weierophinney: Looking at this, I'm wondering if we should have separate methods for creating from:

- string content
- a `StreamInterface` instance

Thoughts?

schnittstabil: I have not enough knowledge about this PSR, but I'm wondering if a `StreamInterface` method don't/shouldn't need the `$size` parameter? – If that is the case, then I believe, two separate methods make sense.

proposed/http-factory/http-factory-meta.md
shadowhand: :+1:

proposed/http-factory/http-factory.md
shadowhand: Good question... I'm not really sure where that came from, I think it's just because http-interop didn't share the namespace.

proposed/http-factory/http-factory.md
shadowhand: Honestly, I wonder if this interface is needed at all. What's the use case for creating an `UploadedFile` that isn't already provided by `ServerRequest`? It's not possible to use `UploadedFile` for transmitting files (client usage) and PSR-7 requires that `ServerRequest` provide functionality to get uploaded files.

I would certainly be happier if this was dropped from PSR-17, as I have found zero real-world use for it.

weierophinney: I foresee a use case within async (non-SAPI) servers, where they may want to create an array of such instances to push into a `ServerRequestInterface`.

Another possibility is when parsing the request body to build the instances, as might happen in non-POST, non-form-encoded requests; having the factory allows middleware that performs this logic to do so without knowledge of the PSR-7 implementation.

shadowhand: @weierophinney I'm not sure what to do about this. What would accepting a `StreamInterface` get us?

weierophinney: @shadowhand — It would allow a process to create the stream for an uploaded file discovered in the request stream *before* creating the actual `UploadedFileInterface` instances. In particular, when I was working with react, I had to parse the request manually, because it's not tied to a SAPI, and doesn't automatically create temporary filehandles. This approach would allow `php://temp` streams to be used in those instances, which could then be used to seed the `UploadedFileInterface` instances.

shadowhand: @weierophinney I tried to follow through with this and got really tied up trying to deal variations in the parameters between different methods. Largely the same type of confusion we had in trying to define the stream factory methods. It seems like there should actually be several methods:

- create from file path ($_FILES compatibility)
- create from string of content
- create from resource
- create from stream

The `$error` parameter really bothers me. In any place except `$_FILES` compatibility, does it matter? Why wouldn't the result always be `UPLOAD_ERR_OK`?

shadowhand: @weierophinney perhaps a better question is... in a system that does not have `$_FILES` compatibility, why would creating uploaded file instances be necessary at all? Why wouldn't those systems simply use the existing stream to write a file or process in some other way?

proposed/http-factory/http-factory.md
```diff
@@ -105,48 +116,44 @@ Has the ability to create streams for requests and responses.
...
+     * The file MUST be opened using the given mode, which may be any mode
+     * supported by the `fopen` function.
+     *
+     * @param string $file
+     * @param string $mode
      *
      * @return StreamInterface
      */
-    public function createStreamFromResource($body);
+    public function createStreamFromFile($file, $mode = 'r');
```
Isinlor: Should this issue be addressed before merging? https://github.com/http-interop/http-factory/issues/28

proposed/http-factory/http-factory.md
```diff
@@ -77,24 +82,30 @@ Has the ability to create server requests.
...
+     * The method and URI of the request SHOULD be derived from the given
+     * server parameters.
      *
-     * @return ServerRequestInterface
-     */
-    public function createServerRequest($method, $uri);
-
-    /**
-     * Create a new server request from PHP globals.
+     * If the method or URI parameter is passed, these values MUST be used
```
Isinlor: What should happen in this case:
```php
$factory->createServerRequest($server, null, $uri)
```

 I think that the intention is to use `$server` value instead of `null`. Nevertheless `these values MUST be used` is suggesting to me that the `null` must be used, therefore `\InvalidArgumentException` should be raised.

proposed/http-factory/http-factory.md
```diff
@@ -77,24 +82,30 @@ Has the ability to create server requests.
...
+     * instead of the server parameters.
+     *
+     * @param array $server
```
Isinlor: What is the structure of this array?

proposed/http-factory/http-factory.md
Isinlor: Actually I'm not sure about rationale of providing the same parameter twice - once in `$server` and then overwriting it in `$method` and `$uri`. Seems like a bad API design to me.

proposed/http-factory/http-factory.md
```diff
@@ -77,24 +82,30 @@ Has the ability to create server requests.
...
+     * @param string|null $method
+     * @param UriInterface|string|null $uri
+     *
+     * @throws \InvalidArgumentException
```
Isinlor: AFAIK the convention is to throw exception implementing PSR specific interface.

proposed/http-factory/http-factory.md
```diff
@@ -105,48 +116,44 @@ Has the ability to create streams for requests and responses.
...
 
     /**
-     * Create a new stream from a resource.
+     * Create a stream from an existing file.
      *
-     * @param resource $body
+     * The file MUST be opened using the given mode, which may be any mode
+     * supported by the `fopen` function.
+     *
+     * @param string $file
```
Isinlor: Should the `$file` string be interpreted the same way as in the `fopen()` function? http://php.net/manual/en/function.fopen.php#refsect1-function.fopen-parameters

proposed/http-factory/http-factory.md
```diff
@@ -105,48 +116,44 @@ Has the ability to create streams for requests and responses.
...
-    public function createStreamFromResource($body);
+    public function createStreamFromFile($file, $mode = 'r');
 
     /**
-     * Create a new stream from a string.
+     * Create a new stream from an existing resource.
      *
-     * A temporary resource will be created with the content of the string.
-     * The resource will be writable and seekable.
+     * The stream MUST be readable and may be writable.
```
Isinlor: Should an appropriate exception be raised if an implementation is not able to comply?

proposed/http-factory/http-factory.md
weierophinney: I think the `$error` parameter is likely not useful here when creating these instances manually; if you're creating it manually, you were likely able to create the stream, which means the upload was successful. The only reason it might _not_ be is if there was a mismatch between things like content-length and actual stream length.

In terms of a system that does not have `$_FILES`, you may want to be able to drop in middleware for handling file uploads that was written agnostic of the environment it'd be in. This would allow it to be used in either a standard SAPI, or with an async server (which parses the incoming request manually in order to create the request instance). In this latter case, it may be useful to delay parsing of uploaded files unless absolutely needed, in order to reduce both CPU and memory overhead; in such a case, _middleware_ might do the actual parsing.

BTW, `$_FILES` is not relevant even in standard SAPIs in one key circumstance: non-POST requests will not populate it! As such, having middleware that can parse the parts and create uploaded file instances is key.

Isinlor:
I took myself the liberty of making quite nitpicking review. I hope it will be helpful ;) .

proposed/http-factory/http-factory.md
weierophinney: Not necessarily; it's varied from one specification to the next. PSR-7 did not define _any_ exceptions, and only specified the most general SPL exceptions that were contextually correct.

That allows implementations to have their own exceptions, so long as they implement the expected exception type from the specification.

proposed/http-factory/http-factory.md
schnittstabil: IMO, following would make more sense:

```php
 interface RequestFactoryInterface
  {
      /**
      * @param string $method
      * @param UriInterface|string $uri
      *
      * @return RequestInterface
      */
     public function createRequest($method, $uri);
 }

interface ServerRequestFactoryInterface
{
      /**
      * @param string $method
      * @param UriInterface|string $uri
      *
      * @return RequestInterface
      */
     public function createServerRequest($method, $uri);

     public function createServerRequestFromArray(array $server = []);
}
```

Thus `createRequest` and `createServerRequest` would share the same signature and `createServerRequestFromArray` can be used that way:

```php
$factory->createServerRequestFromArray($_SERVER)->withMethod(…)->withUri(…);
```

proposed/http-factory/http-factory.md
schnittstabil: `and may be writable.` – that's useless and should be dropped, shouldn't it?

proposed/http-factory/http-factory.md
shadowhand: Yes.

proposed/http-factory/http-factory.md
shadowhand: I like that @schnittstabil. Will update.

proposed/http-factory/http-factory.md
shadowhand: @Isinlor since the resource is already open, I don't see how it would be possible for the implementation to fail. ELI5?

@schnittstabil  Read-only streams are valid.

shadowhand:
@Isinlor @schnittstabil @weierophinney in the interest of not letting this get stale, I am going to merge this and create a separate PR to address your concerns with file upload factory.

proposed/http-factory/http-factory.md
schnittstabil: > Read-only streams are valid.

Yes, they are valid. But as an `interface` consumer, this information does not provide me any(?) useful information. E.g. the stream might be seekable as well, but as a consumer I cannot rely on that either.

But thinking about it, `may be writeable` explicitly allows writable streams, thus it might be useful for `interface` implementers. :+1:

PR #872, opened 2017-01-18
[PSR-6] Put the actual title of the PSR at the top of the document, instead of the word introduction.

alexanderpas:
This puts the actual title of the PSR at the top of the document, instead of the word introduction.

This fixes errors by automated tools that expect the name of the PSR on the first line of the document, such as seen on http://www.php-fig.org/psr/psr-6/ which displays "PSR-6: # Introduction" instead of the expected "PSR-6: Caching Interface" as the title of the page.

### DIFF
--- a/accepted/PSR-6-cache.md
+++ b/accepted/PSR-6-cache.md
@@ -1,4 +1,4 @@
-## Introduction
+# Caching Interface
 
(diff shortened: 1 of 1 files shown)


michaelcullum:
👍

PR #876, opened 2017-02-02
Objects that are allowed to be null MUST go at the end of the argument list

gmponos:
After the discussion I had [here](https://github.com/squizlabs/PHP_CodeSniffer/issues/1092) I believe that the new PSR must explicitly mention that the Objects that are allowed to be null must also go at the end of the argument list.

### DIFF
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -430,7 +430,7 @@ In the argument list, there MUST NOT be a space before each comma, and there
 MUST be one space after each comma.
 
 Method and function arguments with default values MUST go at the end of the argument
-list.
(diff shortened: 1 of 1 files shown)


schnittstabil:
@gmponos Please take also the already accepted [PHP RFC: Parameter Type Widening](https://wiki.php.net/rfc/parameter-no-type-variance) into account (which will probably land in PHP 7.2).

Sorry, but in conjunction with https://github.com/squizlabs/PHP_CodeSniffer/issues/1092#issuecomment-277064552, your suggestion doesn't make much sense to me.

samdark:
I don't see why it should be added. Arguments hinted with an interface/class and with null as default value are still the same default values.

PR #824, opened 2016-10-18
[PSR-12] Uppercasing key word

Jean85:
Minor typo. The `MUST NOT` keyword should always be uppercase.

PR #877, opened 2017-02-04
Uppercasing key word

samdark:
Minor typo. The `MUST NOT` keyword should always be uppercase.

See #824

### DIFF
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -178,7 +178,7 @@ class FooBar
 
 ~~~
 
-Compound namespaces with a depth of more than two MUST not be used. Therefore the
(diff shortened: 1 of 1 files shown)


KorvinSzanto:
Thanks for this Alex! Looks like you have more than just the described change in there, can you add a commit to this to fix that?

schnittstabil:
@KorvinSzanto Good catch. `<` accidentally(?) changed to `≤`.

samdark:
Done.

renanbr:
hi guys, not sure if here is the right place, but trailing commas (line 190 and 201) for grouped use statement [won't work](http://php.net/manual/en/language.namespaces.importing.php#language.namespaces.importing.group)

samdark:
@renanbr that's definitely wrong place since it's not related. Feel free to make a pull request or start a discussion at mailing list.

PR #879, opened 2017-02-06
[PSR-15] Add syntax highlighting & fix markdown bullet list bug

atanvarno69:
* Add syntax highlighting in section 4.2.1 code block
* Use alternate markdown syntax to prevent bullet list rendering bug, see #866 for the same issue/fix in PSR-11's documentation

### DIFF
--- a/proposed/http-middleware/middleware-meta.md
+++ b/proposed/http-middleware/middleware-meta.md
@@ -146,7 +146,7 @@ with one notable exception.
 [Guzzle middleware][guzzle-middleware] is focused on outgoing (client) requests
 and uses this signature:
 
-```
+```php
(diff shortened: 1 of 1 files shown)


weierophinney:
IIRC, the syntax used by the various PSRs is due to the usage of Jekyll to publish the website. As such, I'm unsure if we can accept the changes.

Pinging @michaelcullum @reinink for clarification...

reinink:
Nope, this should be fine. I'd just say be consistent with what's being done elsewhere in this repo, which I believe this change is doing.

michaelcullum:
@php-fig/psr-15

shadowhand:
:+1:

PR #878, opened 2017-02-06
Incorporate feedback on http-middleware

weierophinney:
This patch incorporates feedback that was provided via #867, including the following:

- Updated the titles of each document:
  - Both now use the verbiage "HTTP Server Middleware"
  - Meta document now advertises itself as such
- Added to the meta document:
  - Minor additions to the discussion on why `__invoke` was not chosen.
  - Added a section on why `process` was chosen.
  - Added a section on why "delegate" was chosen, and another on why `$next` was not.

Additionally, combed through the documents for typos, line wraps, etc.

PR #880, opened 2017-02-07
Unify different technical terms

schnittstabil:
In my opinion, a standard should use as few technical terms as possible to not confuse implementers (this PR supersedes http-interop/http-middleware/pull/59).

It unifies different technical terms (_process_, _handle_, _execute_ and _dispatch_):
1. _requests_ and _responses_ are _processed_
2. middleware dispatchers and delegates (of dispatchers) _dispatch_ middlewares

##### _process_ aligns with the method names and the wording introduced by @weierophinney at `middleware-meta.md` (#878):

> #### Why the term "delegate"?
> […] the delegate then processes the request […]

---

@weierophinney I've replaced some of the wording you've recently introduced, please take a look.

### DIFF
--- a/proposed/http-middleware/middleware-meta.md
+++ b/proposed/http-middleware/middleware-meta.md
@@ -190,7 +190,7 @@ significant issues regarding implementation.
 
 The most severe is that passing an empty response has no guarantees that the
 response is in a usable state. This is further exacerbated by the fact that a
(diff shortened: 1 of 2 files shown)


michaelcullum:
@php-fig/psr-15 @php-fig/psr-17

shadowhand:
:+1:

PR #882, opened 2017-02-08
Adding simple negotiation mechanism.

eaton:
The proposed implementation assumes that all Huggable objects are willing to be hugged by all other Huggable objects. The addition of `willHug()` to the proposed interface facilitates a consent-based negotiation process preceding hugs.

Huggable objects are expected to truthfully answer requests regarding their willingness to hug the passed object, but are not required to hug at any point in the future.

### DIFF
--- a/proposed/psr-8-hug/psr-8-hug.md
+++ b/proposed/psr-8-hug/psr-8-hug.md
@@ -36,7 +36,7 @@ its current willingness to engage in a hug with the Huggable passed as the first
 parameter. However, the return value is not a guarantee that future invocations
 will return the same result.
 
-5. An object MUST NOT take additional actions or modify state when mayHug() is
+5. An object MUST NOT take additional actions or modify state when willHug() is
 invoked.
 
 6. An object MAY take additional actions, including modifying state, when hugged.
--- a/proposed/psr-8-hug/psr-8-hug.md
+++ b/proposed/psr-8-hug/psr-8-hug.md
@@ -22,7 +22,7 @@ This specification defines two interfaces, \Psr\Hug\Huggable and
 1. A Huggable object expresses affection and support for another object by invoking
 its hug() method, passing $this as the first parameter.
 
-2. An object whose hug() method is invoked MUST hug() the calling object back
+2. An object whose hug() method is invoked MAY hug() the calling object back
 at least once.
 
 3. Two objects that are engaged in a hug MAY continue to hug each other back for
@@ -31,7 +31,15 @@ condition that will prevent an infinite loop.  For example, an object MAY be
 configured to only allow up to 3 mutual hugs, after which it will break the hug
 chain and return.
 
-4. An object MAY take additional actions, including modifying state, when hugged.
+4. An object whose mayHug() method is invoked MUST return a boolean representing
+its current willingness to engage in a hug with the Huggable passed as the first
+parameter. However, the return value is not a guarantee that future invocations
+will return the same result.
+
+5. An object MUST NOT take additional actions or modify state when mayHug() is
+invoked.
+
+6. An object MAY take additional actions, including modifying state, when hugged.
 A common example is to increment an internal happiness or satisfaction counter.
 
 ### GroupHuggable objects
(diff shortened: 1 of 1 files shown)


robwierzbowski:
If we're going to use metaphors to make object interaction easier to understand, we should use the right ones 👍

schnittstabil:
I'm not sure about this, I believe it would violate the [Tell-Don't-Ask principle](https://martinfowler.com/bliki/TellDontAsk.html) – hence what about a `HuggableExceptionInterface` instead? :wink:

eaton:
An exception was considered, but the `willHug()` mechanism allows exploratory requests before an actual hug is initiated, while a try-catch approach could result in side effects including state change.

eaton:
As a side note, use of `willHug()` allows a calling GroupHuggable to determine whether the target object is willing to hug individual members of the group before `hug()` is invoked, either aborting the attempt or removing members from the set before the `hug()`.

proposed/psr-8-hug/psr-8-hug.md
```diff
@@ -56,9 +64,21 @@ interface Huggable
...
+     * Asks whether a hug will be reciprocated.
+     *
+     * Well-behaved Huggables MAY use this method to verify that their
+     * invocation of hug() will be handled by target object.
+     *
+     * @param Huggable $h
+     *   The object that desires a hug.
+     * @return boolean
+     */
+    public function willHug(Huggable $h);
```
schnittstabil: Is `willHug` the correct technical term? Or does `isWillingToHug` or `wouldHug` fit better?

eaton: Good question — as can be seen from a few earlier commits, it started as `mayHug()` but that clearly implied the wrong direction of invocation.

`wouldhug()` Is probably the best option. According to the [English Grammar StackExchange](http://english.stackexchange.com/questions/3657/when-should-i-use-would-would-have-will-and-will-have), 'would' is a softer verb that implies future potential or conditional mood rather than commitment. In this case, that seems appropriate.

michaelcullum:
@php-fig/psr-8

Crell:
I've given the pre- or post-check some thought.  The trick is that, depending on context, it may be reasonable to assume that objects are hug-compatible or not.  Both are valid situations.  Perhaps we should go ahead and support both a wouldHug() method and a HugNotAcceptedException?  That way an object can respond negatively in either case, but a caller can determine which to rely on depending on the contextual likelihood of an error.  (The exception is more expensive at runtime, but if it's a rare exception that may be a fine trade-off to make.)

proposed/psr-8-hug/psr-8-hug.md
```diff
@@ -31,7 +31,15 @@ condition that will prevent an infinite loop.  For example, an object MAY be
 configured to only allow up to 3 mutual hugs, after which it will break the hug
 chain and return.
 
-4. An object MAY take additional actions, including modifying state, when hugged.
+4. An object whose wouldHug() method is invoked MUST return a boolean representing
+its current willingness to engage in a hug with the Huggable passed as the first
+parameter. However, the return value is not a guarantee that future invocations
+will return the same result.
+
+5. An object MUST NOT take additional actions or modify state when wouldHug() is
```
Crell: It should not modify the other object's state, certainly, but is it reasonable for it to cache the result of a wouldHug() request?  That seems like a usually-valid performance optimization.

Crell:
@eaton In practice, when implementing a group hug wouldn't by definition it be necessary for all participating objects to respond `true` to wouldHug() with all other objects in the set?  It would be an unstable graph otherwise.  That implies an n-squared check prior to group-hug invocation.  Is there a way we can optimize that?

proposed/psr-8-hug/psr-8-hug.md
eaton: An excellent point. The original intent was to make the `wouldHug()` request "penalty-free" — a way of determining the likely response without triggering any potentially negative side effects. But it's unrealistic that the spec REQUIRE it be completely devoid of side effects. I'd support pulling 5.

justafish:
I didn't find the hug RFC very funny in the first place, and although I very much appreciate the original intent of this pull request (thank you @eaton ! <3 ), it seems to have quickly devolved into trying to codify consent and technical architecture discussions about it. IMO this should just be removed altogether.

https://github.com/php-fig/fig-standards/pull/896

Jean85:
Closing as stale. Feel free to restart the conversation if needed.

PR #883, opened 2017-02-09
[PSR-11] simplify implementors and consumers list

stefanotorresi:
also reword the paragraph to be more conservative and less biased

ref: [ML thread](https://groups.google.com/forum/#!msg/php-fig/Ee-Yh23R0fo/MX5jI1_IBQAJ)

### DIFF
--- a/proposed/container-meta.md
+++ b/proposed/container-meta.md
@@ -304,44 +304,32 @@ try {
 8. Implementations
 ------------------
 
-The following projects already implement the `container-interop` version of the interface and
-therefore would be willing to switch to a Container PSR as soon as it is available.
+At the time of writing, the following projects already implement and/or consume the `container-interop` version of the interface.
 
-### Projects implementing `ContainerInterface`
-
-- [Acclimate](https://github.com/jeremeamia/acclimate-container): Adapters for
-  Aura.Di, Laravel, Nette DI, Pimple, Symfony DI, ZF2 Service manager, ZF2
-  Dependency injection and any container using `ArrayAccess`
-- [Aura.DI](https://github.com/auraphp/Aura.Di) (v3+)
+- [Acclimate](https://github.com/jeremeamia/acclimate-container)
+- [Alias-Container](https://github.com/thecodingmachine/alias-container)
+- [Aura.DI](https://github.com/auraphp/Aura.Di)
+- [Behat](https://github.com/Behat/Behat)
 - [dcp-di](https://github.com/estelsmith/dcp-di)
+- [interop.silex.di](https://github.com/thecodingmachine/interop.silex.di)
+- [League Container](https://github.com/thephpleague/container)
 - [Mouf](http://mouf-php.com)
+- [mindplay/middleman](https://github.com/mindplay-dk/middleman)
 - [Njasm Container](https://github.com/njasm/container)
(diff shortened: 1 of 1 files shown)


proposed/container-meta.md
```diff
@@ -304,44 +304,32 @@ try {
...
 - [Njasm Container](https://github.com/njasm/container)
 - [PHP-DI](http://php-di.org)
+- [PHP-DI Invoker](https://github.com/PHP-DI/Invoker)
 - [PimpleInterop](https://github.com/moufmouf/pimple-interop)
+- [Prefixer-Container](https://github.com/thecodingmachine/prefixer-container)
+- [Prophiler](https://github.com/fabfuel/prophiler)
+- [Silly](https://github.com/mnapoli/silly)
+- [Slim](https://github.com/slimphp/Slim)
+- [Splash](http://mouf-php.com/packages/mouf/mvc.splash-common/version/8.0-dev/README.md)
 - [XStatic](https://github.com/jeremeamia/xstatic)
```
Jean85: Symfony can now be added here too!
https://github.com/symfony/symfony/pull/21265

stefanotorresi: @Jean85 this list shows implementors of the `Interop\Container` vendor to show pre-PSR adoption, while Symfony is actually ahead of time by implementing directly `Psr\Container`, so I don't think it should be included.

mnapoli:
👎  the distinction between projects implementing and consuming container-interop is very important: the "consuming" part shows the need for the interface. The "implementing" part shows the "offer".

stefanotorresi:
@mnapoli I beg to disagree. The distinction is not that clear for some projects (e.g. large frameworks have both implementations and consumers), and you cannot possibly make any noteworthy assessment of how many consumers are out there, because most of them are userland applications, not libraries (not to mention that presenting just two middlewares on a separate section is honestly useless).

More to the point, since the research is not comprehensive or in any way thorough, nor it can be kept current, it is basically irrelevant. It's just sample data, I find that presenting it like a statistic can be regarded highly biased.

At the end of the day, I'd rather convey a more relaxed message *"hey, by the way, here are a bunch of packages using this interface already"* rather than a prententious and needy *"Look ma, I have x implementors, y middlewares and z consumers, and look at all the shiny download numbers! We must be so important!"*, but maybe that's just how I read it.

weierophinney:
I tend to agree that removing the downloads bit makes sense. That said, I agree with separating the projects by implementors and consumers, as I think that demonstrates the different development audiences (those that provide containers (implement the interfaces), and those that consume them (typehint against them)). They represent two separate use cases. It's similar to PSR-7; those implementing are usually far fewer, but relevance is more interesting in terms of those who _consume_.

(I'd also likely move the "middleware" section _into_ the "consumers" section.)

stefanotorresi:
@mnapoli @weierophinney I reinstated the previous categories.
I now just added League Container and Zend ServiceManager to implementors, stripped out descriptions (as it didn't make much sense to only describe some packages and leave others without comment), and reworded intro and outro paragraphs.

mnapoli:
@stefanotorresi thanks for taking the time to explain it, it makes more sense now, I understand what you mean about the download numbers.

The new diff looks good to me!

mnapoli:
👍

PR #884, opened 2017-02-09
[PSR-12] Standardize/simplify the opening brace placement

jsuggs:
Basically this removes the conditional placement of the opening brace and keeps the standard consistent across all of the cases.

```php
<?php

namespace Vendor\Package;

class Foo
{
    // Current Implementation
    public someMethod()
    {  <-- on a new line
    }

    // Current Implementation
    public function someOtherMethod(
        Bar $bar,
        Baz $baz
    ) {  <-- not on a new line
        // body
    }

    // Proposed change
    public function someOtherMethodBetter(
        Bar $bar,
        Baz $baz
    )
    {  <-- consistent :)
        // body
    }
}

### DIFF
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -467,6 +467,15 @@ class ClassName
     {
         // method body
     }
+
+    public function aVeryLongMethodNameWithReturnTypeHint(
+        ClassTypeHint $arg1,
+        &$arg2,
+        array $arg3 = []
+    ): string
+    {
+        // method body
+    }
 }
 ~~~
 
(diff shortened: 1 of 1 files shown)


nopolabs:
And with return type hints?
`
public function someOtherMethodBetter(
        Bar $bar,
        Baz $baz
    ): string
    {
        // body
    }
`
vs.
`
public function someOtherMethodBetter(
        Bar $bar,
        Baz $baz
    ): string {
        // body
    }
`

jsuggs:
Thanks @nopolabs added an example with a return type hint.

michaelcullum:
We don't want to explicitly change anything from PSR-2 in PSR-12, only add and clarify.

PR #887, opened 2017-02-14
[PSR-15] Remove duplicate section

schnittstabil:
To be honest, I dunno why the `Why does the delegate conflict with middleware?` section got duplicated – https://github.com/php-fig/fig-standards/commit/6c878b450fb69c053278d540b02863898cd5a738 should have moved it only!?


Pudge601:
I just noticed this myself and was about to make a PR, then saw that this PR already exists.

Looks like it was a simple mistake doing a merge (c9ffa0d), the section got re-added in its original place.

shadowhand:
👍

PR #888, opened 2017-02-14
[PSR-15] Use broader definition of dispatchers

schnittstabil:
* Dispatcher are not necessarily objects, functions can dispatch middlewares as well.
* Dispatcher usually hold `0`, `1`, `1…n` or `0…n` middleware (components), depending on their purpose and framework design; "multiple" is therefore too restrictive.

**NB:** The _next available middleware_ sentence is still part of the `DelegateInterface` contract (section 2.2).

### DIFF
--- a/proposed/http-middleware/middleware.md
+++ b/proposed/http-middleware/middleware.md
@@ -33,23 +33,19 @@ Middleware using this standard MUST implement the following interface:
 
 - `Psr\Http\ServerMiddleware\MiddlewareInterface`
 
-Middleware dispatching systems using this standard MUST implement the following
-interface:
-
-- `Psr\Http\ServerMiddleware\DelegateInterface`
-
 Legacy middleware implementing a double pass approach MUST be wrapped using an
 object that implements the `MiddlewareInterface`.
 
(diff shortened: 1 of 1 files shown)


KorvinSzanto:
I agree with the reasoning in your pull request comment, but I don't understand why you combined delegates with dispatchers.

From my perspective, dispatchers are the supervisor coordinating the process while delegates are the glue between middlewares. These are two distinct ideas and deserve distinct sections in the document as IMO the state of things currently and after this PSR will lead to confusion and to middleware dispatchers implementing the delegate interface for some reason.

schnittstabil:
> I don't understand why you combined delegates with dispatchers.

I believe I didn't, did I? Or are your concerns only about the header/section allocation?

As far as I know, from the editors point of view, only dispatching systems should be allowed to implement the `DelegateInterface`; thus, I don't see a problem in treating both in one section. 

Anyway, I've created a new section for the delegates – @KorvinSzanto what do you think?

weierophinney:
Closing; the section on "Dispatchers" was replaced by one on "Request Handlers", which serve a different purpose.

PR #889, opened 2017-02-14
Rename the last occurence of NotFoundException

kocsismate:


weierophinney:
:+1: 

This is a clarification, and not a substantive change; falls along the lines of a typographical error. @php-fig/secretaries — I assume this can be done without a vote?

PR #898, opened 2017-03-07
[PSR-11] Add Symfony & Twig  to implementers/consumers

chalasr:
Symfony implemented PSR-11 in https://github.com/symfony/symfony/pull/21265 and already consumes it quite a lot. 
Twig consumes it since https://github.com/twigphp/Twig/pull/2401


michaelcullum:
This only refers to container-interop, it's not a maintained list of PSR-11 compatible projects.

PR #899, opened 2017-03-14
Align PSR-5 @method signatures with PHP7 with regard to return types

rquadling:
Currently, the placement of the return type does not match that of PHP7.

For a new developer, this would seem 'wrong' as there are now 2 different signatures for methods.

In addition, allows for the documentation to support static methods that are processed by `__callStatic()`.

And finally, this change removes the ambiguity where a method may be one of two;

```php
<?php

class Parent
{
    public function __call()
    {

    }

    public static function __callStatic()
    {

    }
}

/**
 * @method static getString()
 */
class Child extends Parent
{
    <...>
}
```

Is `getString()` a static method returning void, or a regular method returning a static instance?

By having the scope and return type separated, the ambiguity is removed.

### DIFF
--- a/proposed/phpdoc.md
+++ b/proposed/phpdoc.md
@@ -479,7 +479,7 @@ An example of this is:
 
 ```php
 /**
- * @method int MyMagicMethod(string $argument1) {
+ * @method MyMagicMethod(string $argument1): int {
  *     This is the summary for MyMagicMethod.
  *
  *     @param string $argument1 Description for argument 1.
@@ -578,7 +578,7 @@ Some tags may even feature an "Inline PHPDoc" as shown in the following example.
 
 ```php
 /**
- * @method int MyMagicMethod(string $argument1) {
+ * @method MyMagicMethod(string $argument1): int {
  *     This is the summary for MyMagicMethod.
  *
  *     @param string $argument1 Description for argument 1.
@@ -1220,18 +1220,19 @@ The @method allows a class to know which 'magic' methods are callable.
 
 #### Syntax
 
-    @method [return type] [name]([type] [parameter], [...]) [description]
+    @method <static|> [name]([type] [parameter], [...])[: return type] [description]
 
 #### Description
 
-The @method tag is used in situation where a class contains the `__call()` magic
-method and defines some definite uses.
+The @method tag is used in situation where a class contains the `__call()` or
+`__callStatic()` magic methods and defines some definite uses.
 
-An example of this is a child class whose parent has a `__call()` to have dynamic
-getters or setters for predefined properties. The child knows which getters and
-setters need to be present but relies on the parent class to use the `__call()`
-method to provide it. In this situation, the child class would have a @method
-tag for each magic setter or getter method.
+An example of this is a child class whose parent has a `__call()` or
+`__callStatic` to have dynamic getters or setters for predefined properties.
+The child knows which getters and setters need to be present but relies on the
+parent class to use the `__call()` or `__callStatic()` method to provide it.
+In this situation, the child class would have a @method tag for each magic
+setter or getter method.
 
 The @method tag allows the author to communicate the type of the arguments and
 return value by including those types in the signature.
@@ -1254,9 +1255,11 @@ class Parent
 }
 
 /**
- * @method string getString()
- * @method void setInteger(int $integer)
+ * @method getString(): string
+ * @method setInteger(int $integer): void
  * @method setString(int $integer)
+ * @method static setStaticString(int $integer)
+ * @method static setStaticStringFluid(int $integer): static
  */
 class Child extends Parent
 {
--- a/proposed/phpdoc.md
+++ b/proposed/phpdoc.md
@@ -1220,7 +1220,7 @@ The @method allows a class to know which 'magic' methods are callable.
 
 #### Syntax
 
-    @method <static|> [name]([type] [parameter], [...])[: return type] [description]
+    @method [static] [name]([type] [parameter], [...])[: return type] [description]
 
 #### Description
 


proposed/phpdoc.md
```diff
@@ -1220,18 +1220,19 @@ The @method allows a class to know which 'magic' methods are callable.
 
 #### Syntax
 
-    @method [return type] [name]([type] [parameter], [...]) [description]
+    @method <static|> [name]([type] [parameter], [...])[: return type] [description]
```
schnittstabil: `<static|>` seems not correct according to [ABNF](https://tools.ietf.org/html/rfc5234). Dunno ABNF in all details, but `[static]` looks more appropriate.

rquadling: Ah. OK. But shouldnt' `[name]` be `name` as it is not optional.

schnittstabil: IMO, at least in this case `<name>` would be even better:
> Unlike original BNF, angle brackets ("<", ">") are not required.
> However, angle brackets may be used around a rule name whenever their
> presence facilitates in discerning the use of a rule name. 

But honestly, I cannot really figure out the rules how PSR-5 mixes ABNF and PHP :confused:

rquadling: I'm just learning this too. I don't know what is more 'correct'.

abacaphiliac:
+1

proposed/phpdoc.md
teohhanhui: This section is not in ABNF. :smile:

rquadling: I'm hoping someone who knows how to express what I need will be able to add the syntax.

schnittstabil: @teohhanhui Dunno, if I understand you correctly, but this section _uses_ ABNF.

Thus, either `name` or `<name>` should be used IMO; and to avoid ambiguity, I'm in favor of `<name>`.

rquadling: How about : `@method ["static"] <name>([<type>] [<parameter>] [, ...])[: <returnType>] [description]`

Optional literal static'.
Required name.
Optional type and parameter and repeatable.
Optional : return type.
Optional description.

teohhanhui: If it's ABNF then it'd be `"@method" ["static"] name "(" ...`

schnittstabil: Which brings us back to the issue:
> But honestly, I cannot really figure out the rules how PSR-5 mixes ABNF and PHP :confused:

Obviously the square brackets in `@method [name](…) [description]` have different meaning – or is `name` really optional – or is `description` mandatory?

IMO, an appropriate solution would look something like:
```
@method [static] <name>([<parameter-list>])[: <type>] [<description>]
```

Where square brackets (`[` and `]`) enclose optional elements, and angle brackets (`<` and `>`) denote ABNF non-terminals.

rquadling:
Hello @michaelcullum. Wondering if you have any suggestions on this as you are the 'required approver'.

michaelcullum:
It needs the PSR-5 team to approve, it just defaults to me when nobody is in that team. We need to setup a PSR-5 WG with @mvriel.

samdark:
The change itself makes perfect sense so I think it has a big chance to be merged.

rquadling:
I've sent an email to @mvriel. Not sure if there's anything else I can do at this stage.

GeeH:
I am putting together a WG to get PSR-5 moving again - would be happy if anyone here would like to join.

SpacePossum:
@GeeH if someone is interested in joining the WG, how would one proceed?

GeeH:
Ideally reply to the thread on the FIG mailing list - I'll post more info
here

GeeH:
If you would like to be involved in the working group, please post here https://groups.google.com/forum/#!topic/php-fig/bcV4KXIW6dQ

tractorcow:
Thanks @GeeH I've put my hand up. :)

rquadling:
My proposal for supporting multiple types in DocBook has been accepted (with amendments) into the DocBook standard. [1], [2], [3].

I'm now looking at getting the PHP Documentation rendering tools to output the PHP 7 syntax for function and method signatures.

This may provide a review of the rfc for union types in PHP [4].

In the meantime, is there anyone ready, willing, able to actually move PSR-5 forward a little in this regard. I know it is a time change, but even having it accepted into the standard would allow for additional tooling to start the task of getting things uptodate.


[1] https://github.com/docbook/docbook/issues/91
[2] https://github.com/docbook/docbook/pull/96
[3] http://docbook.org/xml/5.2b03/
[4] https://wiki.php.net/rfc/union_types

ashnazg:
I'm discussing with the phpDocumentor team about the feasibility of deprecating the old syntax in favor of this new syntax, as I think the parser should be able to distinguish between the two.  I've also broached the subject with the WG.  

Any folks from the IDEs: feedback here would be most useful.

neuro159:
Note
`@method static returnType name()`
where `static` is a static qualifier was introduced by PhpStorm in 2014 due to user requests https://youtrack.jetbrains.com/v2/issue/WI-21935

We don't have metrics on how widely it is now used in codebases out there.

neuro159:
While automatic migration is relatively easy...

I personally fail to see the real benefit of changing @method syntax for EVERYBODY, instead of only those wanting @method static. Much easier change would be @method-static, as @property-read.

ashnazg:
I'm mainly swayed by this argument based on the matching return type syntax of PHP7.

rquadling:
With modern tooling, converting the inconsistent style to match PHP 7's syntax should make this a non-issue. There are many tools that can hook into GitHub and other repos that will push the changes quite happily as a pull request.

I suppose another way to look at this is what SHOULD the standard be?

rquadling:
@ashnazg, the ambiguity is with `@method static foo()`. Is `static` scope or return type?

`@method static foo(): static` is clear and unambiguous.

`@method static static foo()` looks like a cut and paste typo, but is correct.

JanTvrdik:
> We don't have metrics on how widely it is now used in codebases out there.

I actually have the metrics (computed based on top 1000 packages on packagist).

~~~
method tag
  static returnType name + (args)                 549      4.245 %
  static returnType name + ()                     242      1.871 %
  static name + (args)                             30      0.232 %
  static name + ()                                  0      0.000 %
  returnType name + (args)                      11148     86.205 %
  returnType name + ()                            767      5.931 %
  name + (args): returnType                         0      0.000 %
  name + (): returnType                             0      0.000 %
  name + (args)                                    78      0.603 %
  name + ()                                        63      0.487 %
  name                                             23      0.178 %
  other                                            32      0.247 %
~~~

So the `@method static <type> <name>(<args>)` syntax is used by over 6 % of `@method` tags.

rquadling:
So, with that analysis, having tools like PHPCS to fix it to PSR-5 would be relatively simple.

@neuro159 With any change, there's resistance. That's fair. Every point has to be argued and an agreement reached. But the "that's how we've always done it" position is problematic in that it can be used to resist all change. For all those new developers coming to PHP in the next decade, why are we forcing 3 different and incompatible syntaxes on them for what should/could be an easy replacement.

In a decade's time, if unchanged, we'll have more developers and more code not matching. The benefit's aren't just about "consistency". If a new developer comes to the PHP documentation and sees one syntax for the functions and methods and tries to write that in their code, it'll fail. If they try  to use the docblock syntax, their code will fail. How is this of any benefit? Tooling can fix this.

Certainly reaching out to the IDEs would be extremely useful. I'm only a user of PHPStorm, but I can certainly put a request for some comments out there. https://youtrack.jetbrains.com/issue/WI-42107

neuro159:
I just was saying that
to remove ambiguity for 6 % of users you make a new convention that affects 100% users
including
 returnType name + (args)                      11148     86.205 %
  returnType name + ()                            767      5.931 %
by moving return type to the end.

It will take me a day to support this, with inspection and auto converter in PhpStorm, its NOT A PROBLEM for IDE. (thanks for request :)

IMO you adding NO value (except aesthetics) to any of those people (those 6% who use static know how it works anyway) and this is the problem.
Do you understand that If they convert to new syntax they loose support in all existing tool chains including old versions of all IDEs and type checkers they have set up? So they would not.

JanTvrdik:
@rquadling New syntax cannot resolve ambiguity. It can only create new one. Ambiguity can only be resolved by introducing a *resolution rule* or by breaking compatibility. **I'm not actually sure at all, what you're proposing.** For example https://youtrack.jetbrains.com/issue/WI-42107 mentions that the *return type declaration* is optional. Is that an oversight? Or is that something you are actually proposing? Can you explicitly specify what *conflict resolution rules* and what breaking changes are you proposing?

ashnazg:
If https://github.com/php/phd/pull/17 does get accepted, I think it gives this PR a bit more strength.

To reiterate, I like this change not so much for resolving the `static static` use case, but to more fully embrace the PHP7 return type hinting syntax and growing its usage.

JanTvrdik:
> I like this change

What change? Can you explain to me what is actually being proposed? You cannot just propose new stuff without proposing how to handle the old stuff.

ashnazg:
@JanTvrdik :  what's being proposed is that the token positions in the `@method` tag be changed, from its PHP4 beginnings in phpDocumentor, to fit the new PHP7 return typing format.

Old way: `@method [optional return type] functionName()`
New way:  `@method [optional scope modifier] functionName() [: optional return type]`

If a return type is given, it now comes after the function name, delimited by colon... just like PHP7 return types are written on the function signature.

If a scope modifier exists, it is now the only thing to expect between the tag and the function name.

That is the change being proposed.  One provided rationale for the change has been the ` static (modifier) static (return type)` potential ambiguity, but what sways me on the overall worth of the change is instead the matching of the return type hinting in PHP7.

In my discussions within the phpDocumentor team, we're thinking we can successfully deal with recognizing both the old and new formats, and treating the old format as "deprecated".  I get the feeling from @neuro159 's comments that IDE support wouldn't be complicated either.

Also, as @rquadling has mentioned repeatedly, updating old code is likely feasible via code sniffing CS fixers.

ashnazg:
Rereading this with fresh eyes... if there's no consensus on changing the `@method` syntax, then maybe @neuro159 's suggestion of `@method-static` is the only other possibility for resolving the aforementioned `@method [static function modifier] [static return type] foo()` use case.

JanTvrdik:
@ashnazg Can't we just require explicit return type for static methods? i.e. the syntax would be

~~~
MethodTagValue
  = StaticKeyword ReturnType MethodName '(' ... ')'
  / [ReturnType] MethodName '(' ... ')'
~~~

(That's what PHPStan currently does)

ashnazg:
@JanTvrdik yes, that would be a viable option to solve the `static` ambiguity.

ashnazg:
It seems like https://github.com/php/phd/pull/17 might be close to being merged.  If it does, then I might look at doing a Google survey to see how much interest there is in changing to @rquadling 's original request for PHP7 return syntax.  If there's mostly pushback to keep the existing `@method` syntax, then that will be that.  However, given the improvements in tooling the last few years, I can imagine a potential majority of community willing to make the change... had tooling been this good way back, I could believe PSR-2 might have been more acceptable to have been more uniform in its rules.

ashnazg:
As much as I'd like this change, I need to decline it for PSR-19 due to level of BC.  Note that I won't forget it, though, when working on a superceding PSR that adds new concepts (generics et al) later on.

rquadling:
The PHP Documentation is now rendered closer to the standard format : http://docs.php.net/manual/en/function.curl-init.php as an example.

WinterSilence:
@JanTvrdik
> So the @method static <type> <name>(<args>) syntax is used by over 6 % of @method tags.

your stats base on PhpDoc for static methods or at all?

PR #900, opened 2017-03-14
Allow multiline control structure expressions

julienfalque:
Ref: https://groups.google.com/forum/?fromgroups#!topic/php-fig/7MSCs8O5byk

IMO it's better to make explicit that newlines are allowed between the parentheses of control structures.

In the examples I added, all expressions parts are on their own line:

```php
if (
    $foo
    && $bar
) {
    // body
}
```

Would adding the following case make sense too?

```php
if ($foo
    && $bar) {
    // body
}
```

Also, I chose _"indented at least once"_ to allow continuous indent, e.g.:

```php
if ($foo
        && $bar) { // indented twice, can help for readability
    // body
}
```

Does it make sense or should this be forbidden instead?

### DIFF
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -652,6 +652,29 @@ if ($expr1) {
 The keyword `elseif` SHOULD be used instead of `else if` so that all control
 keywords look like single words.
 
+Expressions in parentheses MAY be split across multiple lines, where each
+subsequent line is indented at least once. When doing so, the first condition
+MUST be on the next line. The closing parenthesis and opening brace MUST be
+placed together on their own line with one space between them. Boolean
+operators between conditions MUST always be at the beginning or at the end of
+the line, not a mix of both.
+
+~~~php
+<?php
+
+if (
+    $expr1
+    && $expr2
+) {
+    // if body
+} elseif (
+    $expr3
+    && $expr4
+) {
+    // elseif body
+}
+~~~
+
 ### 5.2 `switch`, `case`
 
 A `switch` structure looks like the following. Note the placement of
@@ -681,6 +704,24 @@ switch ($expr) {
 }
 ~~~
 
+Expressions in parentheses MAY be split across multiple lines, where each
+subsequent line is indented at least once. When doing so, the first condition
+MUST be on the next line. The closing parenthesis and opening brace MUST be
+placed together on their own line with one space between them. Boolean
+operators between conditions MUST always be at the beginning or at the end of
+the line, not a mix of both.
+
+~~~php
+<?php
+
+switch (
+    $expr1
+    && $expr2
+) {
+    // structure body
+}
+
+
 ### 5.3 `while`, `do while`
 
 A `while` statement looks like the following. Note the placement of
@@ -690,6 +731,24 @@ parentheses, spaces, and braces.
 <?php
 
 while ($expr) {
+    // structure body
+}
+~~~
+
+Expressions in parentheses MAY be split across multiple lines, where each
+subsequent line is indented at least once. When doing so, the first condition
+MUST be on the next line. The closing parenthesis and opening brace MUST be
+placed together on their own line with one space between them. Boolean
+operators between conditions MUST always be at the beginning or at the end of
+the line, not a mix of both.
+
+~~~php
+<?php
+
+while (
+    $expr1
+    && $expr2
+) {
     // structure body
 }
 ~~~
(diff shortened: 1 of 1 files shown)


teohhanhui:
Wouldn't it be good to standardize on 1 format? Also, trailing `&&`?

samdark:
@teohhanhui could you elaborate?

teohhanhui:
We can reuse similar wording from these sections:

> ## [4.5 Method and function Arguments](https://github.com/php-fig/fig-standards/blob/0193e4f90a87dea4d0c37660d48fb5a8e36d117c/proposed/extended-coding-style-guide.md#45-method-and-function-arguments)
>
> ...
>
> Argument lists MAY be split across multiple lines, where each subsequent line is indented once. When doing so, the first item in the list MUST be on the next line, and there MUST be only one argument per line.
>
> When the argument list is split across multiple lines, the closing parenthesis and opening brace MUST be placed together on their own line with one space between them.
>
> ## [4.7 Method and Function Calls](https://github.com/php-fig/fig-standards/blob/0193e4f90a87dea4d0c37660d48fb5a8e36d117c/proposed/extended-coding-style-guide.md#47-method-and-function-calls)
>
> ...
>
> Argument lists MAY be split across multiple lines, where each subsequent line is indented once. When doing so, the first item in the list MUST be on the next line, and there MUST be only one argument per line. A single argument being split across multiple lines (as might be the case with an anonymous function or array) does not constitute splitting the argument list itself.

So perhaps:

> Composite conditions MAY be split across multiple lines, where each subsequent line is indented once. When doing so, the first item in the list MUST be on the next line, and there MUST be only one condition per line unless it is a composite condition enclosed in braces. The logical operator MUST be placed at the end of the same line as the condition. A single component in a condition being split across multiple lines (as might be the case with an array) does not constitute splitting the composite conditions itself.
>
> When the composite conditions are split across multiple lines, the closing parenthesis and opening brace MUST be placed together on their own line with one space between them.

teohhanhui:
Example:

```php
if (
    $foo &&
    $bar ||
    ($baz < 0 && false !== strpos('foobarbazqux', $qux)) ||
    in_array($qux, [
        'foo',
        'bar',
    ]) &&
    $baz > 30
) {
    // body
}
```

samdark:
Sounds like a good suggestion. Thanks.

Jean85:
Nice suggestion. But, I would be against trailing operators, I prefer them at the start of the line. Maybe a poll or some kind of investigation over the most common practice is needed here?

samdark:
Probably. I've linked this pull request to PSR-12 editors so when they'll awake they'll check it.

julienfalque:
Nice suggestion altough I prefer having logical operators at the start of the lines too.

schnittstabil:
Both styles have their supporters, but they usually never mix them:

```php
// bad:
if (
    $foo &&
    $bar
    || $baz
) {…}

// ok:
if (
    $foo &&
    $bar ||
    $baz
) {…}

// ok too:
if (
    $foo
    && $bar
    || $baz
) {…}
```

Somewhat related: [Python's PEP8](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator).

Jean85:
Yeah, my reason for preferring operators before instead of trailing are the same of Python's PEP8!
TLDR:
> the operators tend to get scattered across different columns on the screen, and each operator is moved away from its operand and onto the previous line

julienfalque:
@teohhanhui Do you mind if I integrate your suggestion into this PR? Or maybe you could submit a PR on my own fork?

teohhanhui:
@julienfalque I've opened #900 as an alternative. It's more general in nature.

KorvinSzanto:
I like this change, but given the current circumstance, I think it's probably a bad idea to try to assert either leading or trailing operators. Could we make this a little more general as to require either all at the beginning or all at the end so that we can avoid any contention?

imiroslavov:
KorvinSzanto, I'm on the same boat with You. I suggest that both styles should be allowed depending on a developer's personal preferences as long as they are not mixed together in one project.

julienfalque:
This PR is staling. I'm willing to work on it again to improve it but I see alternative PRs have been closed. Should I continue or should we close it too?

Jean85:
The PSR is now on review. I think that we previously reached the conclusion with @KorvinSzanto that this shouldn't be regulated, but I would leave that up to him to decide.

julienfalque:
PR updated:
* operators can be either all at the beginning or all at the end of lines;
* opening and closing parentheses must be on their own line.

julienfalque:
Thanks :)

PR #902, opened 2017-03-23
That's not a double pass middleware.

hannesvdvreken:
readme states: deprecated, use this callable notation instead


hannesvdvreken:
@weierophinney @KorvinSzanto @shadowhand et al.

So if you actually take a decent look at that old deprecated repo of mine: I ditched that proposal in favour of double pass. But it has the same principles that the current PSR-15 proposal has.

shadowhand:
Thanks @hannesvdvreken.

PR #903, opened 2017-03-23
[PSR-15] Remove references to pipelines

shadowhand:
The collection of middleware is better referred to as a "stack".
Pipeline processing pattern is different.

Refs https://groups.google.com/d/msg/php-fig/a-obTBtOWnE/MS-fFw2dBwAJ


proposed/http-middleware/middleware-meta.md
```diff
@@ -123,7 +120,7 @@ Middleware taking this approach generally has the following commonalities:
   the request for processing.
 * The middleware is passed 2 arguments during invocation:
   1. An object that represents an HTTP request.
-  2. A delegate that receives the request to dispatch next middleware in the pipeline.
+  2. A delegate that receives the request to dispatch next middleware in the stack.
```
stefanotorresi: :-1: I think the word `pipeline` is much more appropriate than `stack`.

Isinlor: Could you elaborate why? See my post here: https://groups.google.com/d/msg/php-fig/a-obTBtOWnE/MS-fFw2dBwAJ

stefanotorresi: Yep, I read the post.
While I agree that the reference to the "pipeline processing pattern" wikipedia page may result confusing, I think using `stack` instead of `pipeline` is plain wrong, because there may be no stack involved at all: the data structure used to handle a group of middlewares is an implementation detail.
Now the generic `pipeline` word doesn't suggest any type of specific data structure in this regard, while `stack` does.

Isinlor: Maybe we should get rid of it altogether?

> A delegate that receives the request to dispatch next middleware.

shadowhand: This is what I have done.

mnapoli:
I'm afraid of 2 things with this change:

- it removes an important concept (the pipe) which allows to understand more easily how middlewares connect together: if that was a rename of the concept to, e.g., stack, it would be less of a problem IMO. Without explicitly naming the thing and how middlewares relate to it, the text is more vague
- the name "pipe" is quite widespread today, if we were to rename it it would be very confusing and have a big impact (but one could argue that "stack" is also in the ecosystem with StackPHP…)

I don't see a huge benefit over this change honestly (considering the downsides). Pipe is not perfect, stack is not perfect, etc. We are calling it a middleware pipe because of many reasons, that's fine by me.

Isinlor:
@mnapoli Attempting to define how middleware is dispatched is explicitly stated as a non-goal.

> 3.2 Non-Goals (...) Attempting to define how middleware is dispatched.

Do you want to make it a goal?

mnapoli:
@Isinlor that's a good point! I feel a bit stupid now :)

Well I'll let others chime in I don't have a strong opinion ^^

stefanotorresi:
I mostly agree with @mnapoli. The only change that I support is the removal of the explicit reference to the "pipeline pattern" Wikipedia page, because it goes too deep into details that are not all that pertinent to this document, as @Isinlor pointed out.

Using the generic word `pipeline` in the HTTP context, though, is common practice (e.g. see [.NET](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/middleware) and [OpenStack](https://docs.openstack.org/developer/keystonemiddleware/middlewarearchitecture.html)), and I don't see anything wrong with it. 

It doesn't dictate at all the way middleware is dispatched, it only suggests the idea of a linear process happening in stages and, contrary to @Isinlor, I see this concept quite fitting. That's probably why the resemblance of the pipeline pattern was introduced in the first place.

Isinlor:
For the needs of the discussion I feel that it is important to state the difference between **descriptive** and **prescriptive** approaches. Descriptive approach tries to describe the reality, while prescriptive tries to define it by elevating one variety or manner of language use over another (as it is eloquently stated in the Wikipedia article on [Linguistic prescription](https://en.wikipedia.org/wiki/Linguistic_prescription)).  

In my opinion we should not use terms `pipeline`, `stack`, `queue` etc. in _prescriptive_ parts of the document as it is explicitly stated as a non-goal and leads to the unnecessary conflict about the definition of these terms. This is what this pull request is currently achieving.

However I feel that it may be beneficial to _describe_ how the middleware dispatching is working in existing PHP projects or other technologies eg. the ones linked by @stefanotorresi. Whether and how to do that, in my opinion, should be discussed on the mailing list or in a separate pull-request.

stefanotorresi:
@Isinlor I'm aware of the difference, but the metadoc purpose is to clarify the design decisions that shape the specification itself, so it will inevitably contain descriptive language.

Again, I personally think the word `pipeline` is appropriate to refer to what in the spec is defined as [processing multiple middlewares in sequence](https://github.com/php-fig/fig-standards/blob/master/proposed/http-middleware/middleware.md#11-dispatchers).
I don't see any conflict raised by its definition, but if you think it does, I'd rather add the definition itself to the metadoc rather than stripping it out, otherwise people will need clarifications in the future, and we want to prevent this by being as clear as possible.

As it stands, I don't think this PR makes the metadoc more clarifying, it makes it less so.

proposed/http-middleware/middleware-meta.md
```diff
@@ -123,11 +120,10 @@ Middleware taking this approach generally has the following commonalities:
   the request for processing.
 * The middleware is passed 2 arguments during invocation:
   1. An object that represents an HTTP request.
-  2. A delegate that receives the request to dispatch next middleware in the pipeline.
+  2. A delegate that receives the request to dispatch next middleware.
```
weierophinney: Perhaps this could be: "A delegate by which the middleware may process the request if unable to complete the request alone, or unable to generate a response."

That would get rid of the "next middleware" and "pipeline" and "stack" and whatnot verbiage entirely. We could indicate that delegates often act as queues or stacks, but that these are specific implementation details of specific libraries, and not an architectural consideration for this specification.

stefanotorresi: > Perhaps this could be: "A delegate by which the middleware may process the request if unable to complete the request alone, or unable to generate a response."

Spot on.

I particularly approve adding the indication that any sort of underlying data structure is an implementation detail, but we must remember that the spec itself still implies that a collection of middlewares will be processed in some sort of sequence, which to be honest will probably be the case in 99.9% of the implementations, but the spec doesn't really prevent from using delegates in some funky ways, and this may be worth an explicit reminder, just to be safe.

shadowhand:
Closed in favor of #926

PR #904, opened 2017-03-23
Throw an exception if a file cannot be opened

shadowhand:
Refs http-interop/http-factory#28

### DIFF
--- a/proposed/http-factory/http-factory.md
+++ b/proposed/http-factory/http-factory.md
@@ -144,6 +144,9 @@ interface StreamFactoryInterface
      * @param string $filename
      * @param string $mode
      *
+     * @throws \InvalidArgumentException
+     *  If the file cannot be opened using the mode requested.
+     *
      * @return StreamInterface
      */
     public function createStreamFromFile($filename, $mode = 'r');


shadowhand:
@schnittstabil 👍 ?

schnittstabil:
I believe, what to do best, is mainly an PSR-7 implementation problem. @weierophinney seems to be the most compete contact to answer that.

As far as I can see, the only way to implemented that interface involves a `fopen` call. Thus, the warning have to be _caught_ using `set_error_handler` similar to this: 

```php
$error = null;
set_error_handler(function ($severity, $message, $filename, $lineno) use (&$error) {
    $error = new \ErrorException($message, 0, $severity, $filename, $lineno);
}, E_WARNING);
$stream = fopen($path, $mode);
restore_error_handler();

if ($error) {
    throw new InvalidArgumentException(
        "The file \"$path\" could not be opened ($mode).",
        $error
    );
}
```

I dunno; could that be avoided? Maybe by not explicitly forcing `fopen`:
> supported by the `fopen` function

shadowhand:
It is not necessary to rely on `fopen` generating a warning to ensure this functionality. Checking for `file_exists`, `is_readable`, `is_writable` should suffice for 99% of cases.

schnittstabil:
Mhm, would `RuntimeException` be more appropriate?
* [`InvalidArgumentException`](http://php.net/manual/en/class.invalidargumentexception.php): Exception thrown if an argument is not of the expected type.
* [`RuntimeException`](http://php.net/manual/en/class.runtimeexception.php): Exception thrown if an error which can only be found on runtime occurs.

proposed/http-factory/http-factory.md
```diff
@@ -144,6 +144,9 @@ interface StreamFactoryInterface
      * @param string $filename
      * @param string $mode
      *
+     * @throws \InvalidArgumentException
+     *  If the file cannot be opened using the mode requested.
```
weierophinney: Do we want to limit this to `InvalidArgumentException`? I can think of a few other candidates:

- `DomainException`, as it's related to the HTTP message _domain_ context.
- `UnexpectedValueException`, as we are getting a `false` return value from `fopen()`, and not the expected resource.
- `RuntimeException`, as it's happening during runtime.

If we choose _not_ to define it, the implementation can then choose what to do in such scenarios, including what type of exception to throw. Thoughts?

shadowhand: Probably makes sense to use `Exception` here, or define a specific exception interface.

weierophinney: Defining a specific exception interface would require going back to Draft status; indicating simply `Exception` would be a clarification. How would you like to proceed?

shadowhand: Hmm, reading through it again, I actually think that `InvalidArgumentException` is the right type because it means that either given `$filename` or `$mode` is invalid.

I think `Exception` is probably too loose, `DomainException` is awkward (I only use it for application code), `UnexpectedValueException` doesn't make sense because the parameters are have strict types, and `RuntimeException` is more vague than the current value.

shadowhand: Now that I say that, I think this is a good point https://github.com/php-fig/fig-standards/pull/904#issuecomment-300562216. Perhaps `UnexpectedValueException` is the better type due to the fact that `InvalidArgumentException` is related to type more than `UnexpectedValueException`.

jkufner: RuntimeException is the correct exception to use here. The argument may be correct, but somebody might deleted the file; therefore, InvalidArgumentException is not correct. The InvalidArgumentException should be thrown if $mode is invalid:

`````
 /**
  * ...
  * @throws \RuntimeException If the file cannot be opened.
  * @throws \InvalidArgumentException If the $mode is invalid.
  */
`````

shadowhand: @jkufner smart, somehow I didn't even consider throwing two different exception types. Makes perfect sense to me. How about you @weierophinney ?

weierophinney: I've checked with the core committee, and review status is considered less RC and more beta under the current by-laws.

Additionally, I'd consider this a clarification: it's clarifying intended behavior by implementors to consumers when (a) the mode is incorrect, and/or (b) the file cannot be opened.

So: make the change, and I'll review one last time for merge.

shadowhand: Updated.

PR #905, opened 2017-03-24
[PSR-17] Separate methods for creating uploaded files

shadowhand:
Allowing for files to be uploaded from existing streams allows for
uploaded files to be created when not using `POST` or for other kinds of
raw file upload beyond multipart form data.

Refs php-fig/fig-standards#868

### DIFF
--- a/proposed/http-factory/http-factory.md
+++ b/proposed/http-factory/http-factory.md
@@ -183,30 +183,48 @@ interface UploadedFileFactoryInterface
     /**
      * Create a new uploaded file.
      *
-     * If a string is used to create the file, a temporary resource will be
-     * created with the content of the string.
+     * The file SHOULD be opened in read only mode.
      *
-     * If a size is not provided it will be determined by checking the size of
-     * the file.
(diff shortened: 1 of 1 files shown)


proposed/http-factory/http-factory.md
```diff
@@ -210,6 +209,25 @@ interface UploadedFileFactoryInterface
         $clientFilename = null,
         $clientMediaType = null
     );
+
+    /**
+     * Create a new uploaded file from an existing stream.
+     *
+     * The uploaded file SHOULD report a size equal to the length of the stream.
+     *
+     * The uploaded file SHOULD have an error status of `UPLOAD_ERR_OK`.
```
shadowhand: @weierophinney is this what you were thinking would be constraints associated with creating uploaded files from streams?

weierophinney:
Working Group has decided to only define `createUploadedFile()`, with a `StreamInterface $stream` initial argument, as having an untyped `$file` argument was ambiguous and could easily lead to unwarranted complexity and errors.

PR #906, opened 2017-03-24
[PSR-15] Clarify language related to the middleware design

Isinlor:
A response can not be passed to the delegate.

### DIFF
--- a/proposed/http-middleware/middleware-meta.md
+++ b/proposed/http-middleware/middleware-meta.md
@@ -222,7 +222,7 @@ and a delegate and must return a response. The middleware may:
 - Evolve the request before passing it to the delegate to execute the next
   available middleware.
 - Evolve the response received from the delegate before returning it.
-- Create and return a response without passing it to the delegate, thereby
+- Create and return a response without passing the request to the delegate, thereby
   preventing any further middleware from processing.
 
 #### Why doesn't middleware use `__invoke`?


proposed/http-middleware/middleware-meta.md
```diff
@@ -222,8 +222,8 @@ and a delegate and must return a response. The middleware may:
 - Evolve the request before passing it to the delegate to execute the next
   available middleware.
 - Evolve the response received from the delegate before returning it.
-- Create and return a response without passing it to the delegate, thereby
-  preventing any further middleware from processing.
+- Create and return a response without further process delegation, thereby
```
schnittstabil: Maybe I read that wrong, but I believe this means `a/the process (noun)` will not be delegated. But what `process`? The `request process`?

proposed/http-middleware/middleware-meta.md
```diff
@@ -222,8 +222,8 @@ and a delegate and must return a response. The middleware may:
...
+  preventing any inner middleware from processing.
```
schnittstabil: While `inner` refers to the onion image, it does not always fit to all frameworks. E.g. in [Strategility](https://zendframework.github.io/zend-stratigility/middleware/):

```php
// Hello-World middleware
$app->pipe('/', function ($req, $res, $next) {
    if (/* whatever */) {
        return $next($req, $res); // will dispatch FOO middleware
    }

    $res->getBody()->write('Hello world!');
    return $res;
});

// FOO middleware
$app->pipe('/', function ($req, $res, $next) {
    $res->getBody()->write('FOO!');
    return $res;
});
```

I believe the second middleware cannot be seen as an _inner_ middleware.

Isinlor: The word `inner` is currently in use in the document. See: https://github.com/php-fig/fig-standards/blame/master/proposed/http-middleware/middleware-meta.md#L129

The relationship `inner` `outer` is established by the delegate on a run time. If the request processing is delegated by the `Hello-World middleware` to the `FOO middleware` then the `FOO middleware` is the `inner` middleware. The source code visual organization is not relevant.

It could be good tough to clarify the language related to the middlewares even further by defining this relationships in the document itself.

Isinlor:
I decided to limit the change to the single word.

proposed/http-middleware/middleware-meta.md
weierophinney: > I believe the second middleware cannot be seen as an inner middleware.

Yes, it can. In Stratigility, we represent the pipeline as a queue internally; each successive call to `pipe()` appends to the queue. It acts like an onion: delegating moves to the next layer deeper in the queue. In your example, you end up with the following dispatch flow:

```text
application
  - Hello-World
    - FOO
```

As @Isinlor notes, FOO is inner to Hello-World due to how the developer _composes the pipeline_, which in turn determines execution.

That said, how Stratigility uses delegates is an implementation detail. Not every delegate will be operating in the same way. In fact, we have delegates in Expressive now that simply return responses, and which can be used as "defaults" for when a pipeline completes execution without returning a response. How delegates process a request is entirely up to the library and/or individual developer. Onion-style is common, but not the only possibility.

schnittstabil: Granted, it can be seen as an _inner_ middleware.

> How delegates process a request is entirely up to the library and/or individual developer.

Unfortunately, that is not true: 

```php
```

shadowhand:
👍

PR #912, opened 2017-04-24
Update PSR-1 Headings

MayBeTall:
Using the same format as the other recommended changes, like #910.


MayBeTall:
Quick question - would you guys mind if I reformatted all the approved PSRs? I notice they are coming in one-by-one and the approval for each must be tedious on your side.

KorvinSzanto:
@MayBeTall that's probably a better way to do it. We're waiting on @michaelcullum to approve for this to be merged.

Jean85:
This has conficts with the target branch. @MayBeTall are you willing to fix those?

samdark:
Closing since it's already like that in master.

PR #915, opened 2017-05-08
Add Errata regarding DateTimeInterface handling for pre-5.5 environments, using exceptions

Crell:
Trying this again... Here's an Errata for PSR-6 that addresses the missing type hint on expiresAt() using an exception rather than a simulated type check failure.

### DIFF
--- a/accepted/PSR-6-cache-meta.md
+++ b/accepted/PSR-6-cache-meta.md
@@ -332,8 +332,15 @@ class ExpiresAtInvalidParameterException implements Psr\Cache\InvalidArgumentExc
 
 // ...
 
-if (!(null === $expiration || $expiration instanceof \DateTime || $expiration instanceof \DateTimeInterface)) {
-  $error = sprintf('Argument 1 passed to %s::expiresAt() must be an instance of DateTime or DateTimeImmutable, %s given', get_class($this), gettype($expiration));
-  throw new ExpiresAtInvalidParameterException($error);
+if (! (
+        null === $expiration
+        || $expiration instanceof \DateTime
+        || $expiration instanceof \DateTimeInterface
+)) {
+    throw new ExpiresAtInvalidParameterException(sprintf(
+        'Argument 1 passed to %s::expiresAt() must be an instance of DateTime or DateTimeImmutable; %s given',
+        get_class($this),
+        is_object($expiration) ? get_class($expiration) : gettype($expiration)
+    ));
 }
 ```
--- a/accepted/PSR-6-cache-meta.md
+++ b/accepted/PSR-6-cache-meta.md
@@ -319,7 +319,7 @@ specification.
 Despite that, any value that is not null or an object of `\DateTimeInterface` (either
 `\DateTime` or `\DateTimeImmutable`, or possibly others added to the language in
 the future) MUST be treated as an invalid argument error.  Implementers are
-encouraged to actively reject values that do not implement that interface.
+required to actively reject values that do not implement that interface.
 
 Simulating a failed type check unfortunately varies between PHP versions and thus is not
 recommended.  Instead, implementors SHOULD throw an instance of `\Psr\Cache\InvalidArgumentException`.  
--- a/accepted/PSR-6-cache-meta.md
+++ b/accepted/PSR-6-cache-meta.md
@@ -303,3 +303,37 @@ _**Note:** Order descending chronologically._
 [1]: https://docs.google.com/spreadsheet/ccc?key=0Ak2JdGialLildEM2UjlOdnA4ekg3R1Bfeng5eGlZc1E#gid=0
 [2]: https://docs.google.com/spreadsheet/ccc?key=0AsMrMKNHL1uGdDdVd2llN1kxczZQejZaa3JHcXA3b0E#gid=0
 [3]: https://docs.google.com/spreadsheet/ccc?key=0AsMrMKNHL1uGdEE3SU8zclNtdTNobWxpZnFyR0llSXc#gid=1
+
+8. Errata
+----------
+
+### 8.1 Handling of incorrect DateTime values in expiresAt()
+
+The `CacheItemInterface::expiresAt()` method's `$expiration` parameter is untyped
+in the interface, but in the docblock is specified as `\DateTimeInterface`.  The
+intent is that either a `\DateTime` or `\DateTimeImmutable` object is allowed.
+However, `\DateTimeInterface` and `\DateTimeImmutable` were added in PHP 5.5, and
+the authors chose not to impose a hard syntactic requirement for PHP 5.5 on the
+specification.
+
+Despite that, any value that is not null or an object of `\DateTimeInterface` (either
+`\DateTime` or `\DateTimeImmutable`, or possibly others added to the language in
+the future) MUST be treated as an invalid syntax error.  Implementers are
+encouraged to actively reject values that do not implement that interface.
+
+Simulating a failed type check unfortunately varies between PHP versions and thus is not
+recommended.  Instead, implementors SHOULD throw an instance of `\Psr\Cache\InvalidArgumentException`.  
+The following sample code is recommended in order to enforce the type check on the expiresAt()
+method:
+
+```php
+
+class ExpiresAtInvalidParameterException implements Psr\Cache\InvalidArgumentException {}
+
+// ...
+
+if (!(null === $expiration || $expiration instanceof \DateTime || $expiration instanceof \DateTimeInterface)) {
+  $error = sprintf('Argument 1 passed to %s::expiresAt() must be an instance of DateTime or DateTimeImmutable, %s given', get_class($this), gettype($expiration));
+  throw new ExpiresAtInvalidParameterException($error);
+}
+```
--- a/accepted/PSR-6-cache-meta.md
+++ b/accepted/PSR-6-cache-meta.md
@@ -316,10 +316,9 @@ However, `\DateTimeInterface` and `\DateTimeImmutable` were added in PHP 5.5, an
 the authors chose not to impose a hard syntactic requirement for PHP 5.5 on the
 specification.
 
-Despite that, any value that is not null or an object of `\DateTimeInterface` (either
-`\DateTime` or `\DateTimeImmutable`, or possibly others added to the language in
-the future) MUST be treated as an invalid argument error.  Implementers are
-required to actively reject values that do not implement that interface.
+Despite that, implementers MUST accept only `\DateTimeInterface` or compatible types
+(such as `\DateTime` and `\DateTimeImmutable`) as if the method was explicitly typed.
+(Note that the variance rules for a typed parameter may vary between language versions.)
 
 Simulating a failed type check unfortunately varies between PHP versions and thus is not
 recommended.  Instead, implementors SHOULD throw an instance of `\Psr\Cache\InvalidArgumentException`.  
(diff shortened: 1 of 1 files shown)


Crell:
I need to get around to calling a CC vote on this...  Can't merge without a vote.

accepted/PSR-6-cache-meta.md
```diff
@@ -276,25 +276,59 @@ default control information is all that's possible), and because it's trivial
...
+The `CacheItemInterface::expiresAt()` method's `$expiration` parameter is untyped
+in the interface, but in the docblock is specified as `\DateTimeInterface`.  The
+intent is that either a `\DateTime` or `\DateTimeImmutable` object is allowed.
+However, `\DateTimeInterface` and `\DateTimeImmutable` were added in PHP 5.5, and
+the authors chose not to impose a hard syntactic requirement for PHP 5.5 on the
+specification.
+
+Despite that, any value that is not null or an object of `\DateTimeInterface` (either
+`\DateTime` or `\DateTimeImmutable`, or possibly others added to the language in
+the future) MUST be treated as an invalid syntax error.  Implementers are
```
kelunik: That's not a syntax error, just an error.

accepted/PSR-6-cache-meta.md
```diff
@@ -276,25 +276,59 @@ default control information is all that's possible), and because it's trivial
...
+encouraged to actively reject values that do not implement that interface.
```
kelunik: Encouraged would be a should, not a must as above.

weierophinney: I agree with @kelunik here; this last sentence contradicts the previous. The previous says these values MUST be treated as invalid, while this sentence implies a SHOULD instead.

accepted/PSR-6-cache-meta.md
```diff
@@ -276,25 +276,59 @@ default control information is all that's possible), and because it's trivial
...
+
+Simulating a failed type check unfortunately varies between PHP versions and thus is not
+recommended.  Instead, implementors SHOULD throw an instance of `\Psr\Cache\InvalidArgumentException`.
```
weierophinney: Why is this not a MUST? Other specs that define exception types for implementations to throw specify MUST; what is specifically different here that prevents that?

accepted/PSR-6-cache-meta.md
```diff
@@ -276,25 +276,59 @@ default control information is all that's possible), and because it's trivial
...
+```php
+
+class ExpiresAtInvalidParameterException implements Psr\Cache\InvalidArgumentException {}
+
+// ...
+
+if (!(null === $expiration || $expiration instanceof \DateTime || $expiration instanceof \DateTimeInterface)) {
+  $error = sprintf('Argument 1 passed to %s::expiresAt() must be an instance of DateTime or DateTimeImmutable, %s given', get_class($this), gettype($expiration));
+  throw new ExpiresAtInvalidParameterException($error);
+}
```
weierophinney: Can we please make this more readable, e.g., by splitting the conditional over multiple lines and splitting the `sprintf` over multiple lines? The line length right now makes understanding more difficult.

Additionally, because `$expiration` is un-typed, we should likely test for objects vs. PHP built-ins when generating the error message.

Suggestion:

```php
if (! (
        null === $expiration
        || $expiration instanceof \DateTime
        || $expiration instanceof \DateTimeInterface
)) {
    throw new ExpiresAtInvalidParameterException(sprintf(
        'Argument 1 passed to %s::expiresAt() must be an instance of DateTime or DateTimeImmutable; %s given',
        get_class($this),
        is_object($expiration) ? get_class($expiration) : gettype($expiration)
    ));
}
```

accepted/PSR-6-cache-meta.md
```diff
@@ -276,25 +276,59 @@ default control information is all that's possible), and because it's trivial
 for a particular implementation to include as an add-on should it desire to
 do so.
 
-5. People
+5. Errata
```
weierophinney: Adding a section mid-document means re-flowing all remaining numbers. May I suggest that errata become a new section at the bottom of the document?

accepted/PSR-6-cache-meta.md
stefanotorresi: I think `invalid syntax error` should either become `invalid argument error` or just `error`.

accepted/PSR-6-cache-meta.md
stefanotorresi: yea, `s/encouraged/required/` I guess.

accepted/PSR-6-cache-meta.md
Crell: Hrm. I had thought that's what we'd done elsewhere, but I went back and looked and the only PSR with Errata is PSR-2, which didn't have a metadoc in the first place so it's a bad example.

Crell:
Needed to rebase as this branch was OOOOOOOOLD.

The distinction between MUST and SHOULD is because, as this is errata, I don't think we can add new hard-requirements to the spec.  Viz, we can say "if you accept something other than DateTimeInterface then You're Doing It Wrong" because that was the original spec, just not syntactically enforced.  However, since we never said what you should do instead we can only recommend the way in which you reject the parameter.

Viz, if someone is already rejecting an int or string by throwing a `\InvalidArgumentException`, we cannot now say that they're non-compliant because they should be throwing `\Psr\Cache\InvalidArgumentException` instead.  We can only demand that they do reject it somehow.

I adjusted the language to make it clear that rejecting invalid values is a requirement, but the particular way of doing so is just a recommendation.

kelunik:
If you're strict, you can't enforce that either. Accepting further types is prefectly fine with variance rules.

Crell:
The spec includes the docblock, which makes it clear that DateTimeInterface is the only legal value.  If someone is accepting something else already then they're already violating the spec, just not the language interface, technically.  The first paragraph is only reiterating what the spec already says, and thus is no change.

kelunik:
@Crell No, the spec says that these types must be accepted, it doesn't say anything about other types. Even if the interface declared `?\DateTimeInterface` as type, PHP 7.2 would allow dropping that type in an implementation and accepting any other type. See https://wiki.php.net/rfc/parameter-no-type-variance and https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science)#Contravariant_method_argument_type for more information.

stefanotorresi:
@kelunik has a point.
Could we go back to the rationale behind this errata for a minute?

Why do we need to dictate how to handle these details?

Loosening this requirement, while possible, would just be an implementation detail.

For example, if an implementation accepts, in addition to a `DateTimeInterface`, also a string in ISO format, would that make it non-compliant?

I think compliance is respected as long as consumers are able to pass a `DateTimeInterface`.

The issue with the absence of the type hint is that implementers cannot enforce this particular type, even if they wanted to. As pointed out, such an enforcement is still implementation specific, so if we want to endorse it, we should just state that implementors are encouraged to write this method as if the `DateTimeInterface` typehint was present, meaning that the method must work with *at least* such a type. It's up to them if they want to duck-type the hell out of it or not.

dbu:
> such an enforcement is still implementation specific, so if we want to endorse it, we should just state that implementors are encouraged to write this method as if the DateTimeInterface typehint was present, meaning that the method must work with at least such a type. 

I guess its too late to change this to strict requirements, but there is a point in making the spec strict: interchangeability. If you rely on a PSR-6 implementation accepting something else than the interface, things might stop working when your code is used with a different PSR-6 implementation. When writing a reusable library, this is very important. In an application less so, but it still weakens the value you get from using the PSR abstraction.

I would not encourage to allow other formats but instead explain that for best interoperability, only DateTimeInterface should be used.

Jean85:
I'm weighting in as a secretary, just for interpretation of how the PSR should be followed:

> I guess its too late to change this to strict requirements, but there is a point in making the spec strict: interchangeability. If you rely on a PSR-6 implementation accepting something else than the interface, things might stop working when your code is used with a different PSR-6 implementation. When writing a reusable library, this is very important. In an application less so, but it still weakens the value you get from using the PSR abstraction.

I agree that it's too late, we all agree I think. But the PSR (and this errata) says that you SHOULD accept only `\DateTimeInterface`; since other kind of parameters are not specified, if your lib accepts them is on you, and it's something out of compat from the PSR; so the end user cannot rely on that as a PSR-compliant behavior; so the burden of what you've describe falls on the well-meaning of the implementors.

I know that that's not ideal, but it's late to fix that. IMHO this errata is the best that we can do in this situation.

stefanotorresi:
@dbu 
> If you rely on a PSR-6 implementation accepting something else than the interface, things might stop working when your code is used with a different PSR-6 implementation.
Every implementation will have its own details, we can't prevent that, and honestly we don't even want that, because otherwise we would be shipping libraries, not interfaces. If a consumer decide to rely on implementation details, that's on him, it's not the spec fault.

> When writing a reusable library, this is very important. 
Yes, of course we must be very clear about the intent, but we can't prevent implementors to do their own thing while being compatible at the same time.

@Crell I reiterate my questions to you, the editor:
What's the intent of the errata?
Is it to clarify the absence of the type hint and more clearly define what it should've been and why it has be omitted, or is it to forbid any kind of type variance, even if php 7.2 introduced contravariance?

stefanotorresi:
LGTM

Crell:
There was further slack discussion leading to the changes in the latest commit, which most parties here agreed were acceptable. I'm calling a new vote shortly.

accepted/PSR-6-cache-meta.md
```diff
@@ -303,3 +303,43 @@ _**Note:** Order descending chronologically._
...
+### 8.1 Handling of incorrect DateTime values in expiresAt()
+
+The `CacheItemInterface::expiresAt()` method's `$expiration` parameter is untyped
+in the interface, but in the docblock is specified as `\DateTimeInterface`.  The
+intent is that either a `\DateTime` or `\DateTimeImmutable` object is allowed.
+However, `\DateTimeInterface` and `\DateTimeImmutable` were added in PHP 5.5, and
+the authors chose not to impose a hard syntactic requirement for PHP 5.5 on the
+specification.
+
+Despite that, implementers MUST accept only `\DateTimeInterface` or compatible types
```
kelunik: I still think that this should also be a `SHOULD`.

Crell: The problem with accepting other things is that it encourages non-standard behavior.  If a consumer uses a library that also accepts strings, let's say, even though it's type hinting against PSR-6 interfaces, then it's now unwittingly coupled itself to that implementation and created a hidden dependency.  That's what specs like this are intended to not do.

The goal here is to approximate as well as possible if we had targeted PHP 5.5 and just included the type hint in the first place, which is what we should have done in the first place. 🤷‍♂️

kelunik: Hence it's a should. A should is still a strong recommendation. And as said, even with a type declaration, PHP 7.2 allows dropping that.

Jean85: @kelunik that would be implicit due to the language behavior, the next paragraph says

> [...] as if the method was explicitly typed.
> (Note that the variance rules for a typed parameter may vary between language versions.)

... to cover exactly that edge case.

accepted/PSR-6-cache-meta.md
```diff
@@ -303,3 +303,43 @@ _**Note:** Order descending chronologically._
...
+
+```php
+
+class ExpiresAtInvalidParameterException implements Psr\Cache\InvalidArgumentException {}
+
+// ...
+
+if (! (
+        null === $expiration
+        || $expiration instanceof \DateTime
```
sgolemon: Unless the goal is to support PHP < 5.5, this line is redundant given the Interface check below.  If the goal is to support < 5.5, then ignore this comment.

Crell: Support for PHP < 5.5 is the whole reason we're in this mess in the first place. :smile:

Crell:
The vote has passed with 11 Yes, 0 no, 1 not voting.  Merging!

PR #917, opened 2017-05-08
PSR 9 and 10 are inverted

jjanvier:
PSR 9 and 10 are inverted
Also, PSR 10 has a wrong title


michaelcullum:
PSR-9 is about security advisories disclosure. PSR-10 is about the reporting processs so it's currently correct.

PR #918, opened 2017-05-17
Add rules on splitting a compound boolean expression / multi-line condition

teohhanhui:
Replaces #900


teohhanhui:
Yeah, any suggestions on making this DRY?

teohhanhui:
"A compound boolean expression used as the condition in a control structure statement ..."?

teohhanhui:
The problem is how to capture the differences / additional rules when a CBE used as a condition is split across multiple lines. The "base" rules for a CBE being split across multiple lines would still apply.

KorvinSzanto:
There hasn't been any movement on this PR for quite a while so I'm thinking we should close this PR unless someone else would like to take it over and try to make it more approachable.

PR #923, opened 2017-07-13
equal signs.

hafah:
image

Sometimes I see this;
```
public $foo=null;
```


samdark:
PSR-2 amendments aren't accepted anymore. Consider PSR-12 that is in the works and will replace PSR-2.

PR #924, opened 2017-07-13
[psr-12] equal sign

hafah:
image

Sometimes I see this;
```
public $foo=null;
```

### DIFF
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -391,6 +391,8 @@ The `var` keyword MUST NOT be used to declare a property.
 
 There MUST NOT be more than one property declared per statement.
 
+There MUST be one space after and before the equal sign.
+
 Property names MUST NOT be prefixed with a single underscore to indicate
 protected or private visibility. That is, an underscore prefix explicitly has
 no meaning.


proposed/extended-coding-style-guide.md
```diff
@@ -391,6 +391,8 @@ The `var` keyword MUST NOT be used to declare a property.
 
 There MUST NOT be more than one property declared per statement.
 
+There MUST be one space after and before the equal sign.
```
Jean85: I agree that this should be the norm, as I see this being practiced everywhere in PSR 1 and 2 code examples, but I don't think we should restrict this rule to the assignment operator only, but to all operators. Also, I think there is the odd one, the equals sign used in declare statements, since it's now widely more popular due to the strict type use in PHP 7.

samdark:
It's actually already there in PSR-12:

> All binary and ternary (but not unary) operators MUST be preceded and followed by at least one space. This includes all arithmetic, comparison, assignment, bitwise, logical (excluding ! which is unary), string concatenation, type operators, trait operators (insteadof and as), and the single pipe operator (e.g. ExceptionType1 | ExceptionType2 $e).

PR #926, opened 2017-07-24
[PSR-15] Replace delegate with handler interface

shadowhand:
Request handler is a fundamental operation that middleware depends on.
The handler can be used with or without middleware.

As per [this discussion](https://groups.google.com/d/msg/php-fig/mjoqnByUrTI/QScOmHZkAgAJ).

### DIFF
--- a/proposed/http-middleware/middleware-meta.md
+++ b/proposed/http-middleware/middleware-meta.md
@@ -1,56 +1,75 @@
-# HTTP Server Middleware Meta Document
+HTTP Server Request Handlers Meta Document
+==========================================
 
 ## 1. Summary
 
-The purpose of this PSR is to provide an interface that defines the formal
-method signature for HTTP Server Middleware ("middleware") that is compatible
-with HTTP Messages, as defined in [PSR-7][psr7].
+The purpose of this PSR is to define formal interfaces for HTTP server request
+handlers ("request handlers") and HTTP server request middleware ("middleware")
+that are compatible with HTTP messages as defined in [PSR-7][psr7].
 
-_**Note:** Any references to "middleware" in this document are specific to
-**server middleware**._
+_Note: All references to "request handlers" and "middleware" are specific to
+**server request** processing._
 
 [psr7]: http://www.php-fig.org/psr/psr-7/
 
 ## 2. Why Bother?
 
-The HTTP Messages specification does not contain any reference to HTTP Middleware.
+The HTTP messages specification does not contain any reference to request
+handlers or middleware.
 
-The design pattern used by middleware has existed for many years as the
-[pipeline pattern][pipeline], or more specifically, "linear pipeline processing".
-The general concept of reusable middleware was popularized within PHP by
-[StackPHP][stackphp]. Since the release of the HTTP Messages standard, a number
-of frameworks have adopted middleware that use HTTP Message interfaces.
+Request handlers are a fundamental part of any web application. The handler is
+the component that receives a request and produces a response. Nearly all code
+that works with HTTP messages will have some kind of request handler.
 
-Agreeing on a formal  middleware interface eliminates several problems and
-provides a number of benefits:
+[Middleware][middleware] has existed for many years in the PHP ecosystem. The
+general concept of reusable middleware was popularized by [StackPHP][stackphp].
+Since the release of HTTP messages as a PSR many frameworks have adopted
+middleware the use HTTP message interfaces.
 
-* Provides a formal standard for middleware developers to commit to.
+Agreeing on formal request handler and middleware interfaces eliminates several
+problems and provides a number of benefits:
+
+* Provides a formal standard for developers to commit to.
+* Enables any middleware component to run in any compatible framework.
 * Eliminates duplication of similar interfaces defined by various frameworks.
 * Avoids minor discrepancies in method signatures.
-* Enables any middleware component to run in any compatible framework.
 
-[pipeline]: https://en.wikipedia.org/wiki/Pipeline_(computing)
+[middleware]: https://en.wikipedia.org/wiki/Middleware
 [stackphp]: http://stackphp.com/
 
 ## 3. Scope
 
 ### 3.1 Goals
 
-* Create a middleware interface that uses HTTP Messages.
-* Ensure that middleware will not be coupled to a specific implementation of HTTP Messages.
-* Implement a middleware signature that is based on best practices.
+* Create a request handler interface that uses HTTP messages.
+* Create a middleware interface that uses HTTP messages.
+* Implement request handler and middleware signatures that are based on
+  best practices.
+* Ensure that request handlers and middleware will be compatible with any
+  implementation of HTTP messages.
 
 ### 3.2 Non-Goals
 
-* Attempting to define how middleware is dispatched.
-* Attempting to define interfaces for client/asynchronous middleware.
 * Attempting to define the mechanism by which HTTP responses are created.
+* Attempting to define interfaces for client/asynchronous middleware.
+* Attempting to define how middleware is dispatched.
 
-## 4. Approaches
+## 4. Request Handler Approaches
 
-There are currently two common approaches to middleware that use HTTP Messages.
+There are many approaches to request handlers that use HTTP messages. However,
+the general process is the same in all of them:
 
-### 4.1 Double Pass
+Given an HTTP request, produce an HTTP response for that request.
+
+The internal requirements of that process will vary from framework to framework
+and application to application. This proposal makes no effort to determine what
+that process should be.
+
+## 5. Middleware Approaches
+
+There are currently two common approaches to middleware that use HTTP messages.
+
+### 5.1 Double Pass
 
 The signature used by most middleware implementations has been mostly the same
 and is based on [Express middleware][express], which is defined as:
@@ -76,14 +95,14 @@ A significant number of projects provide and/or use exactly the same interface.
 This approach is often referred to as "double pass" in reference to both the
 request and response being passed to the middleware.
 
-#### 4.1.1 Projects Using Double Pass
+#### 5.1.1 Projects Using Double Pass
 
 * [mindplay/middleman v1](https://github.com/mindplay-dk/middleman/blob/1.0.0/src/MiddlewareInterface.php#L24)
 * [relay/relay v1](https://github.com/relayphp/Relay.Relay/blob/1.0.0/src/MiddlewareInterface.php#L24)
 * [slim/slim v3](https://github.com/slimphp/Slim/blob/3.4.0/Slim/MiddlewareAwareTrait.php#L66-L75)
 * [zendframework/zend-stratigility v1](https://github.com/zendframework/zend-stratigility/blob/1.0.0/src/MiddlewarePipe.php#L69-L79)
 
-#### 4.1.2 Middleware Implementing Double Pass
+#### 5.1.2 Middleware Implementing Double Pass
 
 * [bitexpert/adroit](https://github.com/bitExpert/adroit)
 * [akrabat/rka-ip-address-middleware](https://github.com/akrabat/rka-ip-address-middleware)
@@ -106,9 +125,9 @@ request and response being passed to the middleware.
 * [relay/middleware](https://github.com/relayphp/Relay.Middleware)
 
 The primary downside of this interface is that the while the interface itself is
-a callable, there is currently no way to type hint a closure in a similar way.
+a callable, there is currently no way to strictly type a closure.
 
-### 4.2 Single Pass (Lambda)
+### 5.2 Single Pass (Lambda)
 
 The other approach to middleware is much closer to [StackPHP][stackphp] style
 and is defined as:
@@ -122,19 +141,19 @@ Middleware taking this approach generally has the following commonalities:
 * The middleware is defined with a specific interface with a method that takes
   the request for processing.
 * The middleware is passed 2 arguments during invocation:
-  1. An object that represents an HTTP request.
-  2. A delegate that receives the request to dispatch next middleware in the pipeline.
+  1. A HTTP request message.
+  2. A request handler to which the middleware can delegate the responsibility
+     of producing an HTTP response message.
 
 In this form, middleware has no access to a response until one is generated by
-innermost middleware. Middleware can then modify the response before returning
-back up the stack.
+the request handler. Middleware can then modify the response before returning it.
 
 This approach is often referred to as "single pass" or "lambda" in reference to
 only the request being passed to the middleware.
 
-#### 4.2.1 Projects Using Single Pass
+#### 5.2.1 Projects Using Single Pass
 
-There are fewer examples of this approach within projects using HTTP Messages,
+There are fewer examples of this approach within projects using HTTP messages,
 with one notable exception.
 
 [Guzzle middleware][guzzle-middleware] is focused on outgoing (client) requests
@@ -144,9 +163,9 @@ and uses this signature:
 function (RequestInterface $request, array $options): ResponseInterface
 ```
 
-#### 4.2.2 Additional Projects Using Single Pass
+#### 5.2.2 Additional Projects Using Single Pass
 
-There are also significant projects that predate HTTP Messages using this approach.
+There are also significant projects that predate HTTP messages using this approach.
 
 [StackPHP][stackphp] is based on [Symfony HttpKernel][httpkernel] and supports
 middleware with this signature:
@@ -155,7 +174,7 @@ middleware with this signature:
 function handle(Request $request, $type, $catch): Response
 ```
 
-*__Note__: While Stack has multiple arguments, a response object is not included.*
+_Note: While Stack has multiple arguments, a response object is not included._
 
 [Laravel middleware][laravel-middleware] uses Symfony components and supports
 middleware with this signature:
@@ -168,16 +187,16 @@ function handle(Request $request, callable $next): Response
 [httpkernel]: https://symfony.com/doc/2.0/components/http_kernel/introduction.html
 [laravel-middleware]: https://laravel.com/docs/master/middleware
 
-### 4.3 Comparison of Approaches
+### 5.3 Comparison of Approaches
 
 The single pass approach to middleware has been well established in the PHP
 community for many years. This is most evident with the large number of packages
 that are based around StackPHP.
 
 The double pass approach is much newer but has been almost universally used by
-early adopters of PSR-7 (HTTP Messages).
+early adopters of HTTP messages (PSR-7).
 
-### 4.4 Chosen Approach
+### 5.4 Chosen Approach
 
 Despite the nearly universal adoption of the double-pass approach there are
 significant issues regarding implementation.
@@ -212,18 +231,49 @@ being passed implements a middleware signature, which reduces runtime safety.
 
 **Due to these significant issues the lambda approach has been choosen for this proposal.**
 
-## 5. Design Decisions
+## 6. Design Decisions
 
-### 5.1 Middleware Design
+### 6.1 Request Handler Design
 
-The `MiddlewareInterface` defines a single method that accepts a server request
-and a delegate and must return a response. The middleware may:
+The `RequestHandlerInterface` defines a single method that accepts a request and
+MUST return a response. The request handler MAY delegate to another handler.
 
-- Evolve the request before passing it to the delegate to execute the next
-  available middleware.
-- Evolve the response received from the delegate before returning it.
-- Create and return a response without passing the request to the delegate, thereby
-  preventing any further middleware from processing.
+#### Why is a server request required?
+
+To make it clear that the request handler can only be used in a server side context.
+In an client side context a [promise][promises] would likely be returned instead
+of a response.
+
+[promises]: https://promisesaplus.com/
+
+#### Why the term "handler"?
+
+The term "handler" means something designated to manage or control. In terms of
+request processing, a request handler is the point where the request must be
+acted upon to create a response.
+
+As opposed to the term "delegate", which was used in a previous version of this
+specification, the internal behavior of the this interface is not specified.
+As long as the request handler ultimately produces a response, it is valid.
+
+#### Why doesn't request handler use `__invoke`?
+
+Using `__invoke` is less transparent than using a named method. It also makes
+it easier to call the request handler when it is assigned to a class variable,
+without using `call_user_func` or other less common syntax.
+
+_See "discussion of FrameInterface" in [relevant links](#8-relevant-links) for
+ additional information._
+
+### 6.2 Middleware Design
+
+The `MiddlewareInterface` defines a single method that accepts an HTTP request
+and a request handler and must return a response. The middleware may:
+
+- Evolve the request before passing it to the request handler.
+- Evolve the response received from the request handler before returning it.
+- Create and return a response without passing the request to the request handler,
+  thereby handling the request itself.
 
 #### Why doesn't middleware use `__invoke`?
 
@@ -233,9 +283,9 @@ forwards compatibility with this specification.
 
 #### Why the name `process()`?
 
-We reviewed a number of existing MVC and middleware frameworks to determine
-what method(s) each defined for processing incoming requests. We found the
-following were commonly used:
+We reviewed a number of existing monolithic and middleware frameworks to
+determine what method(s) each defined for processing incoming requests. We found
+the following were commonly used:
 
 - `__invoke` (within middleware systems, such as Slim, Expressive, Relay, etc.)
 - `handle` (in particular, software derived from Symfony's [HttpKernel][HttpKernel])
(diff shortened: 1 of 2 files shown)


proposed/http-middleware/middleware-meta.md
```diff
@@ -123,7 +130,7 @@ Middleware taking this approach generally has the following commonalities:
   the request for processing.
 * The middleware is passed 2 arguments during invocation:
   1. An object that represents an HTTP request.
-  2. A delegate that receives the request to dispatch next middleware in the pipeline.
+  2. A handler that receives the request to dispatch next middleware in the pipeline.
```
weierophinney: This implies architecture: that the handler passed necessarily composes middleware and/or has a queue or stack of middleware. The handler _could_ simply return a response.

Perhaps another way to phrase it: "A handler that may be called in order to produce a response for the middleware."

proposed/http-middleware/middleware-meta.md
```diff
@@ -270,75 +276,37 @@ additional information._
...
-
-Additionally, since we are defining an object, we chose to use a noun instead of
-a verb to name the interface.
+#### Why is a server request required?
 
-Further, we did not choose the term `next` for the action delegates invoke, as
-that verb implies a queue or stack. The delegate is not required to implement
-either pattern internally in order to do its work; its only job is to _process_
-the request to return a response.
+To make it clear that the handler can only be used in a synchronous, server
```
weierophinney: Is the word "synchronous" needed here? If so (I think it is), we need to explain _why_, and how it might differ from an async architecture (where a promise would likely be returned instead of a response).

proposed/http-middleware/middleware.md
```diff
@@ -31,12 +31,11 @@ and the creation of a resulting HTTP Response, as defined by PSR-7.
...
 
-- `Psr\Http\ServerMiddleware\MiddlewareInterface`
+- `Psr\Http\Server\ServerMiddlewareInterface`
 
-Middleware dispatching systems using this standard MUST implement the following
-interface:
+Handlers using this standard MUST implement the following interface:
 
-- `Psr\Http\ServerMiddleware\DelegateInterface`
+- `Psr\Http\Server\HandlerInterface`
```
weierophinney: In reading the rest of the document, I'm wondering if this should be `RequestHandlerInterface` or `ServerRequestHandlerInterface` instead? That would make it far more clear _what_ it handles.

proposed/http-middleware/middleware.md
```diff
@@ -46,7 +45,7 @@ object that implements the `MiddlewareInterface`.
 An HTTP middleware dispatcher is an object that holds multiple middleware
 components that can be used to process one or more requests in sequence.
 
-The middleware dispatcher MUST pass the request and a delegate to each
+The middleware dispatcher MUST pass the request and a delegate handler to each
```
weierophinney: This is now the first time the term `delegate` crops up. I suggest rephrasing from "delegate handler" to "request handler". The following line would then replace "delegate" as well.

proposed/http-middleware/middleware-meta.md
mnapoli: How about:

> A handler to which the middleware can delegate the responsibility of producing a response.

I think having the term "delegate" in the sentence helps connecting the concepts?

proposed/http-middleware/middleware.md
```diff
@@ -73,34 +72,34 @@ interface MiddlewareInterface
...
      * @param ServerRequestInterface $request
-     * @param DelegateInterface $delegate
+     * @param HandlerInterface $handler
      *
      * @return ResponseInterface
      */
     public function process(
         ServerRequestInterface $request,
-        DelegateInterface $delegate
+        HandlerInterface $handler
```
mnapoli: With that new signature it is now even less clear why a handler is injected here. All this time I thought the new signature would be:

```php
function process(ServerRequestInterface $request, HandlerInterface $delegate)
```

How about this signature:

```php
function process(ServerRequestInterface $request, HandlerInterface $next)
```

?

That way we come back to what everyone is familiar with, and `next` is very clear to what it means (it's the next handler in the stack).

proposed/http-middleware/middleware.md
```diff
@@ -109,6 +108,3 @@ interface DelegateInterface
     public function process(ServerRequestInterface $request);
```
mnapoli: I think this method name should be renamed right? Else a class cannot implement both interfaces.

`__invoke()` was mentioned on Slack but then it is contradicted by a section of the document (which is being removed in this PR):

> #### Why doesn't delegate use `__invoke`?
>
> Classes that define `__invoke` can be more loosely type hinted as `callable`, which results in less strict typing. This is generally undesirable, especially when the method uses strict typing.

proposed/http-middleware/middleware-meta.md
```diff
@@ -270,75 +276,37 @@ additional information._
...
-additional information._
-
-#### Why doesn't delegate use `__invoke`?
+### 5.2 Handler Design
 
-Classes that define `__invoke` can be more loosely type hinted as
-`callable`, which results in less strict typing. This is generally undesirable,
-especially when the method uses strict typing.
+The `HandlerInterface` defines a single method that accepts a request and
+MUST return a response. The handler MAY delegate to another handler.
```
mnapoli: I think the sentence "The handler MAY delegate to another handler." is not necessary, a handler can do anything, the implementation details are not part of the contract.

proposed/http-middleware/middleware-meta.md
```diff
@@ -270,75 +276,37 @@ additional information._
...
-In terms of middleware design, a delegate is called upon by middleware when the
-middleware is unable to process the request itself; the delegate then processes
-the request _for the original middleware_ in order to return a response.
-
-#### Why isn't the delegate a `callable`?
-
-Using an interface type hint improves runtime safety and IDE support.
-
-_See "discussion of FrameInterface" in [relevant links](#8-relevant-links) for
-additional information._
```
mnapoli: Should this section stay for the `HandlerInterface`?

proposed/http-middleware/middleware-meta.md
```diff
@@ -270,75 +276,37 @@ additional information._
...
-The `DelegateInterface` defines a single method that accepts a request and
-returns a response. The delegate interface must be implemented by any middleware
-dispatcher that uses middleware implementing `MiddlewareInterface`.
-
-#### Why the term "delegate"?
-
-The term "delegate" means something designated to act for or represent another.
-In terms of middleware design, a delegate is called upon by middleware when the
-middleware is unable to process the request itself; the delegate then processes
-the request _for the original middleware_ in order to return a response.
```
mnapoli: The point of the meta document is to be able to explain the logic behind the choices. If we decide to drop the `delegate` name, we should explain why. (because if we can't that means that we maybe shouldn't)

proposed/http-middleware/middleware.md
mnapoli: Personally I think it's really clear as it is.

Also for practical reasons PSR-7 interfaces are already extremely painful to use (especially when writing middlewares - `ServerRequestInterface` argh I hate that), that would be even worse.

proposed/http-middleware/middleware.md
```diff
@@ -31,12 +31,11 @@ and the creation of a resulting HTTP Response, as defined by PSR-7.
 
 Middleware using this standard MUST implement the following interface:
 
-- `Psr\Http\ServerMiddleware\MiddlewareInterface`
+- `Psr\Http\Server\ServerMiddlewareInterface`
 
-Middleware dispatching systems using this standard MUST implement the following
-interface:
+Handlers using this standard MUST implement the following interface:
```
mnapoli: Should we make it explicit that middlewares should implement `ServerMiddlewareInterface`, even those that do not need the `$handler` (aka `$delegate`) parameter.

The reason for that is that if they don't, interoperability with other middlewares through "dispatchers" (as called in this document, aka pipes/stacks) is lost.

proposed/http-middleware/middleware.md
shadowhand: Neither of those is necessarily accurate, as it depends entirely on context. Both `$delegate` and `$next` imply that the handler is not going to produce a response by itself, and that is an unknown.

Basically this boils down to "if the middleware cannot produce a response by itself, it will call the handler to produce one" and I can't think of any term besides `$handler` that is more clear on that point.

proposed/http-middleware/middleware.md
shadowhand: Whoops, mistake on my part, this should have become `__invoke`.

proposed/http-middleware/middleware-meta.md
shadowhand: The only reason I included that was to make it clear that a handler does not necessarily have to produce a response itself, eg: it can act as a delegate.

proposed/http-middleware/middleware-meta.md
shadowhand: Yes, good point.

proposed/http-middleware/middleware.md
shadowhand: Isn't that implied by the fact that an implementation of an interface method cannot modify the method signature (even with parameter type widening)?

shadowhand:
@weierophinney @mnapoli changes have been made and additional details added. Please review as you are able.

proposed/http-middleware/middleware.md
weierophinney: I'm agreeing with @shadowhand here; `$next` implies that the handler has some sort of stack/queue behind it. `$delegate` would work, but would raise questions about why the typehint is `HandlerInterface` (or the counter-proposal I did of `RequestHandlerInterface`/`ServerRequestHandlerInterface`). Having it as `$handler`, with clear text in the proposal as to its purpose (_to handle a request and produce a response if the middleware cannot produce one on its own_) should be sufficient.

proposed/http-middleware/middleware-meta.md
weierophinney: I'm both pro- and con- this.

The term `$delegate` was originally coined for this particular proposal. If it doesn't make it into the final version, it's just a bit of historical trivia.

That said, documenting the decision making process (e.g., double-pass to single-pass, delegate to handler) may also answer some questions later — and, more importantly, potentially _prevent_ some.

Thoughts, @shadowhand ?

shadowhand: See next comment.

proposed/http-middleware/middleware-meta.md
```diff
@@ -270,75 +286,48 @@ additional information._
...
+request processing, a handler is the point where the request must be acted upon
+to create a response.
 
-Further, we did not choose the term `next` for the action delegates invoke, as
-that verb implies a queue or stack. The delegate is not required to implement
-either pattern internally in order to do its work; its only job is to _process_
-the request to return a response.
+As opposed to the term "delegate", which was used in a previous version of this
+specification, the internal behavior of the this interface is not specified.
+As long as the handler ultimately produces a response, it is considered valid.
```
shadowhand: @weierophinney I think this correctly addresses the historical reasons for `handler` as opposed to `delegate`. Thoughts?

proposed/http-middleware/middleware.md
mnapoli: Right I think I see things a bit differently now: the middleware decorates a handler. Whether there are other middlewares involved is a detail.

👍  for me then thanks

proposed/http-middleware/middleware-meta.md
mnapoli: With all the explanations I got in this thread that's good for me!

proposed/http-middleware/middleware.md
```diff
@@ -1,114 +1,123 @@
...
+It is RECOMMENDED that any middleware or request handler that generates a response
+will use HTTP factories as defined in [PSR-17][psr17] in order to prevent dependence
+on a specific HTTP message implementation.
+
+[psr17]: https://github.com/php-fig/fig-standards/tree/master/proposed/http-factory
 
 ## 2. Interfaces
 
-### 2.1 Psr\Http\ServerMiddleware\MiddlewareInterface
+### 2.1 Psr\Http\Server\MiddlewareInterface
```
renanbr: Interfaces sub-sections titles and defined namespaces looks not congruent.

Here in 2.1
Psr\Http\\**Server**\MiddlewareInterface
Psr\Http\\**ServerStrategy**\MiddlewareInterface

Idem for HandlerInterface in 2.2

shadowhand:
PR has been updated to reflect discussed namespace corrections.

proposed/http-middleware/middleware-meta.md
```diff
@@ -2,43 +2,50 @@
...
 
 ## 3. Scope
 
 ### 3.1 Goals
 
 * Create a middleware interface that uses HTTP Messages.
+* Create a handler interface that uses HTTP Messages.
 * Ensure that middleware will not be coupled to a specific implementation of HTTP Messages.
-* Implement a middleware signature that is based on best practices.
+* Implement handler and middleware signatures that are based on best practices.
```
weierophinney: `s/handler/request handler/` (based on slack discussion)

proposed/http-middleware/middleware-meta.md
```diff
@@ -122,12 +129,12 @@ Middleware taking this approach generally has the following commonalities:
 * The middleware is defined with a specific interface with a method that takes
   the request for processing.
 * The middleware is passed 2 arguments during invocation:
-  1. An object that represents an HTTP request.
-  2. A delegate that receives the request to dispatch next middleware in the pipeline.
+  1. A HTTP request message.
+  2. A handler to which the middleware can delegate the responsibility of
```
weierophinney: `s/handler/request handler/` (for consistency)

proposed/http-middleware/middleware-meta.md
```diff
@@ -122,12 +129,12 @@ Middleware taking this approach generally has the following commonalities:
...
+     producing an HTTP response message.
 
 In this form, middleware has no access to a response until one is generated by
-innermost middleware. Middleware can then modify the response before returning
-back up the stack.
+the handler. Middleware can then modify the response before returning it.
```
weierophinney: Here as well (request handler).

proposed/http-middleware/middleware-meta.md
```diff
@@ -217,13 +224,12 @@ being passed implements a middleware signature, which reduces runtime safety.
 ### 5.1 Middleware Design
 
 The `MiddlewareInterface` defines a single method that accepts a server request
-and a delegate and must return a response. The middleware may:
+and a handler and must return a response. The middleware may:
```
weierophinney: Here as well (request handler).

proposed/http-middleware/middleware-meta.md
```diff
@@ -217,13 +224,12 @@ being passed implements a middleware signature, which reduces runtime safety.
...
 
-- Evolve the request before passing it to the delegate to execute the next
-  available middleware.
-- Evolve the response received from the delegate before returning it.
-- Create and return a response without passing the request to the delegate, thereby
-  preventing any further middleware from processing.
+- Evolve the request before passing it to the handler.
+- Evolve the response received from the handler before returning it.
+- Create and return a response without passing the request to the handler, thereby
+  acting as a handler itself.
```
weierophinney: Here as well (request handler; all three bullet points).

proposed/http-middleware/middleware-meta.md
```diff
@@ -249,6 +255,16 @@ themselves as middleware (or middleware compatible with this specification),
 and, as such, needed to choose a name not in common usage. As such, we chose
 `process`, to indicate _processing_ a request.
 
+#### Why isn't the handler a `callable`?
+
+Despite the fact that the handler uses the `__invoke` method, it should not be
```
weierophinney: Here as well (request handler; this sentence and in title).

proposed/http-middleware/middleware.md
```diff
@@ -1,114 +1,123 @@
 HTTP Server Middleware
```
weierophinney: Should the name of the proposal be updated, since we cover both middleware and request handlers? Perhaps "HTTP Server Components" (to borrow verbiage from the opening paragraph)?

proposed/http-middleware/middleware.md
```diff
@@ -1,114 +1,123 @@
...
-Middleware dispatching systems using this standard MUST implement the following
-interface:
+Middleware using this standard MUST implement the following interface:
 
-- `Psr\Http\ServerMiddleware\DelegateInterface`
+- `Psr\Http\Server\MiddlewareInterface`
 
-Legacy middleware implementing a double pass approach MUST be wrapped using an
-object that implements the `MiddlewareInterface`.
+### 1.2 Request Handlers
```
weierophinney: I think we need to invert sections 1.1 and 1.2, as the discussion of middleware relies on request handlers.

proposed/http-middleware/middleware.md
```diff
@@ -1,114 +1,123 @@
...
     public function process(
         ServerRequestInterface $request,
-        DelegateInterface $delegate
+        RequestHandlerInterface $handler
     );
 }
 ```
 
-### 2.2 Psr\Http\ServerMiddleware\DelegateInterface
+### 2.2 Psr\Http\Server\RequestHandlerInterface
```
weierophinney: Same comment here as for sections 1.1/1.2 - invert these, so that the `RequestHandlerInterface` is introduced first.

shadowhand:
@weierophinney 
> change the title to "HTTP Server Components"

How about "HTTP Server Request Handlers"?

proposed/http-middleware/middleware-meta.md
```diff
@@ -2,43 +2,50 @@
...
+Agreeing on formal middleware and handler interfaces eliminates several
+problems and provides a number of benefits:
 
 * Provides a formal standard for middleware developers to commit to.
 * Eliminates duplication of similar interfaces defined by various frameworks.
 * Avoids minor discrepancies in method signatures.
 * Enables any middleware component to run in any compatible framework.
 
-[pipeline]: https://en.wikipedia.org/wiki/Pipeline_(computing)
+[middleware]: https://en.wikipedia.org/wiki/Middleware
```
shadowhand: The purpose of the usage is to reference what "middleware" is as a general concept, not just "http middleware".

proposed/http-middleware/middleware.md
shadowhand: We have decided to use `Psr\Http\Server` as the root namespace.

shadowhand:
@weierophinney reorganized sections and ensured that "request handler" is written out when applicable.

atanvarno69:
I don't see any explanation of why `__invoke()` is the chosen method name for `HandlerInterface`. Since the handler handles a request, `handle()` would seem more descriptive.

Defining `__invoke()` as the interface method precludes an implementer from using that method name (and hence callable functionality for the class) for something else. Whereas, using a normal (non-magic) method name allows an implementer to proxy `__invoke()` to it if they want.

While this design choice is left unexplained I can't tell if it is good and has benefits over a normal method name that I'm not seeing.

shadowhand:
@atanvarno69 that is a valid concern. My answer would be:

> The request handler uses `__invoke` to indicate that the handler should only have a single purpose: to transform the request into a response. Using `__invoke` implies that the request handler will be used in a functional way.
>
> Additionally, due to [parameter type widening](https://wiki.php.net/rfc/parameter-no-type-variance) the handler MAY be replaced by another callable that serves the same purpose.

Would that be a satisfactory answer?

Thoughts @php-fig/psr-15 ?

mickadoo:
> Additionally, due to parameter type widening the handler MAY be replaced by another callable that serves the same purpose.

I can see the point here, for example in some framework where you want to replace the main handler with a callable

However I think it might be not so clear how the handler should be call if it is a property of a class, for example

```
  public function doSomething(Request $request) {
    $handler = $this->handler;
    return $handler($request); // works
    return ($this->handler)($request); // works
    return $this->handler->__invoke($request); // works
    return $this->handler($request); // does not work. treats handler as method call
  }
```

I would say that requiring 2 lines, as the first option shows, is not practical. The second option works, but is a new syntax provided in PHP7 and might not be obvious. The third option seems wrong because it's explicitly calling magic methods. To be honest, the 4th option was my first attempt, but it caused an error.

This is my first comment on this repo, but in this case I'm not sure what the goal is. I would have thought it is to provide useful, well-defined interfaces for the PHP community, but when changing the descriptive method name `handle` to `__invoke` I think it's focusing more on the implementation side of things.

weierophinney:
@mickadoo Honestly, this has been hashed out over and over and over again for months; head over to the [http-interop/http-middleware](https://github.com/http-interop/http-middleware) repo and browse through the issues and pull requests for the copious threads.

Whatever choice we make here will be "wrong" for a significant number of people. Some, like you, feel that a named method is the better choice; others are adamant that __invoke() makes the most sense.

At some point, we need to make a decision, and that's ultimately up to the editor of the specification (@shadowhand in this case). I will back whatever decision is made; there are ample reasons for and against either approach, and each is supportable.

atanvarno69:
@shadowhand yes, that is exactly the sort of explanation that should be in the meta document.

@mickadoo unfortunately @weierophinney is correct. Personally, I am against `__invoke`, but the argument has dragged on long enough. I want to get it into review (where the method name will probably be discussed again at length) more than I want my preferred method name.

Other notes on the spec: it is silent on what exceptions implementations can throw. I expect the answer is "none". This could be made explicit otherwise untrusting devs will have to wrap a fair amount of logic in try/catch blocks.

Also, PSR-11 suggests implementers should mark they provide an implementation in their composer.json. I think this would be useful for PSR-15 too: compliant middle packages provide `psr/middleware-implementation` and require `psr/middleware-handler-implementation` (or whatever names) which are provided by packages that provide handlers for middleware only (rather than using the handler interface more generally).

weierophinney:
> Other notes on the spec: it is silent on what exceptions implementations can throw. I expect the answer is "none". This could be made explicit otherwise untrusting devs will have to wrap a fair amount of logic in try/catch blocks.

They can certainly throw exceptions. Typical middleware stacks will employ one or more error handling middleware that wrap calls to the delegate/handler in a try/catch block in order to log, report, etc.

atanvarno69:
Sorry, it's late and I was not specific (or thinking particularly clearly). I was thinking about the handler not originating an exception. Though I suppose in a stack it all bubbles up regardless of origin.

Anyway, my main point was I found it odd exceptions weren't mentioned at all. One sentence to set expectations seems a good idea. Or a docblock `@throws \Exception This method MAY throw an exception.`

shadowhand:
I don't think the interfaces need to define exceptions. It seems reasonable to have a sentence that states something like:

> A request handler MAY throw an exception if request conditions prevent the creation of a response. The type of exception is dependent on implementation and/or application code.

shadowhand:
@atanvarno69 I have updated the PR to include the reasoning for request handler using `__invoke` in the meta document and wording about exceptions in the interface document.

atanvarno69:
@shadowhand 👍 from me.

I think a Package section à la [PSR-11](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-11-container.md) would be a nice to have, but hardly a deal breaker.

shadowhand:
@atanvarno69 a Package section will be added once package names are decided. Thanks!

mindplay-dk:
FWIW, I remain very firmly against the use of `__invoke()` in interfaces - nothing requires (or entitles) this PSR (or any other) to dictate, block, or reserve the use of this magic method for this or any specific purpose.

Magic methods do not belong in interfaces. It's simply not what magic methods are for. It's a complete misunderstanding of this language feature.

Please consider the inclusion of standard closure-adapter for the interface instead - that is the normal, natural way to deal with OO issues of this sort.

I oppose this PSR and any PSR that attempts to hijack magic methods for their own purposes.

mnapoli:
@mindplay-dk yes that's a good point. Should we use `handle()` instead? That would match the class name, and would make that interface very similar to Symfony's HttpKernelInterface.

mindplay-dk:
@mnapoli in my own daily terminology, handlers *process* an incoming request into a response, and previously the name was `process()`, so I'd prefer we just keep it that way.

Of course, I'd prefer we didn't have this interface in the first place - I'd rather we standardize on [request handlers](https://github.com/http-request-handler-interop/fig-standards) first, which would also negate the discussion about changing the name of this PSR because of this interface, which isn't middleware-specific at all; the same reason I pushed for the request-handler PSR to begin with. I know it's a tedious process, but it would be cleaner this way.

mnapoli:
`process()` will conflict with `MiddlewareInterface`. Some classes will need to implement both. For example an `Application` class will probably implement `HandlerInterface` (like Symfony's HttpKernelInterface) but implementing `MiddlewareInterface` will also be a good thing because that will allow to integration the whole application into another as a middleware.

Same for routers, pipes/stacks, etc.

> I'd rather we standardize on request handlers first

This is included in this PR.

shadowhand:
I don't feel like having this argument all over again. Will replace `Handler::__invoke` with `Handler::handle`.

flik:
The point of the meta document is to be able to explain the logic behind the choices. Is it right ?

shadowhand:
@flik yes.

mindplay-dk:
> `process()` will conflict with `MiddlewareInterface`. Some classes will need to implement both.

Gotcha :-)

> > I'd rather we standardize on request handlers first

> This is included in this PR.

I know. I meant, as a separate PSR.

Handlers have broad, general applications outside of the middleware domain.

That's the primary motivation for the separate PSR.

weierophinney:
@mindplay-dk The plan currently is to ship the handler interface via a separate package from the middleware interface, which will allow you to use it standalone. I think this should suffice for your purposes.

PR #931, opened 2017-09-05
Fix markdown syntax for heading levels

borisceranic:
Minor issues with the markdown syntax causes 1st level headings not to be rendered properly.

Also, increase levels of headings for X.Y sections from 3rd level to 2nd level, where they belong.


michaelcullum:
Thanks!

PR #932, opened 2017-09-07
Added @min tag link, description and example.

HRADigital:


proposed/phpdoc.md
```diff
@@ -1264,7 +1265,45 @@ class Child extends Parent
 }
 ```
 
-### 7.12. @package
+### 7.12. @min
+
+The @min allows a developer to tag a structure with a minimum PHP version requirment.
```
samdark: Does that make any sense? Nowadays we have Composer that allows you to specify requirements. You won't get any code if your requirements aren't met.

HRADigital: I'm sorry to say so, but IMHO that makes even less sense.
This would allow you to "calculate" in real-time, what you would set on the composer file, not the other way around.

If the maximum "@min" for the project is 7.1, then that's what you'll set on composer.

You can reflect all the objects in a project, calculate the "@min" value for the (ex:.) class, based on the maximum "@min" value of its methods.
The maximum "@min" from all classes, would become the minimum required (PHP) version for your project. That's what you'll set on composer.

HRADigital: Note: I have seen projects with PHP7 syntax, when it had as a minimum requirement in composer 5.6.

samdark: Do you mean that developer have to specify `@min` for each method and class?

HRADigital: Just method, if (s)he wants to be compliant with the standard.
Automatic code validation/formatting tools can do the rest.
Those tools can even have "strict" or "permissive" modes, but that's up to them.

There are a lot of other tags people don't use either.
Some people don't even comment/document.

samdark: It is unlikely that developer who cares not about PHP version in his `composer.json` would care about `@min` tag.

HRADigital: Nor would (s)he care about "@param", "@return"...
But that doesn't mean the rest of us don't care.

samdark: `@return` and `@param` are used by IDEs to autocomplete and inspect code and in many cases it's not possible to generate these automatically.

samdark: Actually... I think it is possible to create a tool that will scan the project and determine minimum required PHP version to be put into `composer.json`.

samdark: But not if syntax is the same but behavior is different i.e. a bug was fixed in a certain PHP version and you don't want to implement a fix in the code for older versions.

HRADigital: Yet, "@return" and "@param" are in the standard...
IDEs will implement whatever the standard tells them to.

The "@min" tag would only facilitate the process.

HRADigital: If by automated tool, you mean some sort of manual version mapping, then the "automation" becomes relative. Someone will need to maintain that mapping (1737 native).
Also, a similar tool would already exist, as there's already a "@deprecated" tag in the standard.

This tag would allow the developers to maintain their own code, and the tool to be a lot simpler and faster.

ashnazg: A tag like this implies to me that different files in one codebase are not uniformly designed to work together as a unit.  Presumably one given codebase should have such metadata ("minimum PHP version") at the overall package level, not on a per-file or per-class level.

-1 from me on this, I'm afraid.

HRADigital: @ashnazg , If you use (ex:.) random_int() in one of your classes, does that mean that you'll need to change all of your classes to include at least one PHP7 function?
Of course you'll definitely have different minimum (PHP) versions in different classes. That's a given. Nonetheless, that doesn't invalidate my point.

If you have PHP5.6 compliant code in all of your product, and add random_int() inside one of your class methods, that method becomes a "@min 7.0.0".
An automated tool could also mark that class as a "@min 7.0.0", due to that method, but not all the others.
That same automated tool, can print for you in the command line "7.0.0", just due to the fact that one of your project's classes uses "@min 7.0.0". That makes it, by definition, a PHP7 project, not a 5.6.
You will know what to do next with your composer file, I presume.

-1 back to you.

PS: I can only imagine your version tracking, if there's more than just you developing on the same project. I'm sure you'll have this locked.

samdark: @HRADigital that doesn't mean that you have to add PHP 7 everywhere but that does mean that the codebase overall would require PHP 7 in order to be used successfully.

samdark: And it makes sense to add such requirement to `composer.json` right away. It sounds quite weird to forget raising `composer.json` requirement but putting `@min` tag in the doc at the same time. You won't be running a project  with PHP 7 only parts with PHP 5.6, right?

ashnazg: Are you saying that it's preferable for a *human* to denote (via `@min`) what the definitive "minimum PHP version that can run this method" is?  And said human will keep such information up to date?  And that other developers need to scan all `@min` tags in a candidate library to figure out if they can use it, based on their own PHP version limitations?

I think the majority of developers are going to look for one definitive declaration at the library's metadata level for such information.  If I'm reading the documentation for a method in order to see how to use it, I'm not concerned that the method alone might need a PHP version I don't have... because I looked at the library's metadata to see if the library's PHP requirement works for me or not.

I think there's a place and time for such "minimum usable PHP for this method" information, but I think that place is in a static analysis tool, and the time is when the developers of that code run such a tool against their overall package in order to know what "one minimum PHP version is required for this package overall"... and that resulting minimum version is what they denote at the metadata level.

HRADigital: @samdark , you would obviously just add "@min 7.0.0" to the method that will use random_int(), not all around the project. The rest of the project will remain as it was.
"@min 7.0.0" becomes a minimum version for that specific method, which will make the class that contains that method (by definition/omission) a 7.0.0 class, which will make the project...
You get the idea!

You can add to the composer straight away, or you may not.
It's easier for an automated tool (in "strict" mode) to "flag" a CI and/or code reviewer.

samdark: Well, makes no sense to me. Why would you have a PHP 7.0 only method in a codebase that runs on PHP 5.6?

HRADigital: @ashnazg , I'm not sure you're looking at this the right way...

If you're using a 3rd party extension/package, you shouldn't even care if there's a "@min" tag or not. It's not for you.
The package has a version, and that should be your concern.

If you're talking about one of the projects you're working on, then look at this possibility:

1. The project is a "@min 7.0.0".
You develop with 7.0.0 in mind.
Suddenly, you'll need to add a native function, only available in PHP7.1. You'll either make the decision to use it, or request someone else for permission to use it.
If granted, you'll just add "@min 7.1.0" to the method's DocBlock.

2. Another developer for the same project, now sees (automatically?) that the project became a PHP7.1 project. (S)he will use 7.1 functionality, or not. Lets imagine not.

3. Later on, you decide to remove that 7.1 function, and use an alternative (7.0 / internal).
You don't know if someone else started to use 7.1 as well.

4. If your method reverts back to "@min 7.0", and your project suddenly reverts back as well, you'll have your answer.

Nonetheless, it's just an optional tag.
You'll either use it, or not.
You'll either see its usefulness, or not.
It's up to you, and/or your project's reviewer.

PS: I'm positive you don't actually use all the tags available in the standard. This is just another (useful) option.

ashnazg: Understand, I'm **not** saying there's *zero* value in the information you're suggesting be captured by such a tag.

What I *am* saying is that developers are more likely to put such information at *one* higher level, not at *all* the lower levels.

Also, that static analysis can determine this reliably, rather than having humans do it manually.  Packages like @llaville's php-compat-info have been doing so since the PEAR days.

HRADigital: @samdark , or a 7.1 method in a 7.0 project, right?

jkufner:
You can use php -l (lint) to check if the project conforms to your 
version of PHP. And if you have the minimal desired version installed, 
you can use that version for the check.

Nobody is going to maintain these annotations manually.

proposed/phpdoc.md
HRADigital: @ashnazg , but is that analysis validated?
Is it version mapping of (native) functions? Does it support other language features?
My system accuses >1700 count of native functions alone. Someone needs to maintain that.
"@min" at your method's level, is easier and more direct.

Another example can be, if you're working on a bigger project, with more developers, and your project's version increases, could you easily identify the cause/origin?

HRADigital: Just to terminate this discussion, which is already too long.
"@min" tag is just a tag.
You either use it or not.
You either require it or not.
You either validate it or not.
This being a standard for PHP, I don't see why it shouldn't have a possible solution for this.

It's just like any other tag that you probably don't use.

ashnazg: I could take your use case description, alter it slightly to be `@tested passed` or `@tested failed` based on the last run of the CI build, and the narrative seems the same.

I can see a narrow use for it, yes... but I don't think its use case is significant enough to earn its way into this standard.  Using it as a custom tag is certainly permitted by the standard, but I don't think it qualifies as a standard-level tag.

samdark: @HRADigital 

> or a 7.1 method in a 7.0 project, right?

Doesn't matter. If you have a part of the project that runs on 7.1 you certainly won't run the project on 7.0 because part of it won't work. You're not using multiple PHP versions to run different parts of the same codebase, right?

> Another example can be, if you're working on a bigger project, with more developers, and your project's version increases, could you easily identify the cause/origin?

Yes. Find a commit when `composer.json` PHP version was increased and check what else was changed.

samdark:
@jkufner syntax only. Not behavior.

proposed/phpdoc.md
HRADigital: @samdark ,
> Doesn't matter. If you have a part of the project that runs on 7.1 you certainly won't run the project on 7.0 because part of it won't work.

Locally and for you, you mean?

> Find a commit when composer.json PHP version

What if it's not updated? What if it's no longer needed?

@ashnazg ,
> I can see a narrow use for it, yes (...) but I don't think it qualifies as a standard-level tag.

I can't argue with that.
If it's not, then it's not...

samdark: > Locally and for you, you mean?

No matter if it's locally or not if the project requires PHP 7.1 in order for all its parts to function then PHP 7.1 should be used both locally and in production.

> What if it's not updated? What if it's no longer needed?

If it's not updated then it will be exactly the same situation as not updating `@min` tags.

PR #935, opened 2017-09-29
[PSR12] Add `new` to lower case keyword list.

JABirchall:
The `new` keyword should be enforced to lowercase to prevent usages such as `New Class()` or `NEW Class()`.

### DIFF
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -109,7 +109,7 @@ tabs for indenting.
 
 ### 2.5 Keywords and Types
 
-All PHP reserved types and keywords [[1]][keywords][[2]][types] MUST be in lower case.
+All PHP reserved keywords and types [[1]][keywords][[2]][types] MUST be in lower case.
 
 Short form of type keywords MUST be used i.e. `bool` instead of `boolean`,
 `int` instead of `integer` etc.
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -111,7 +111,7 @@ tabs for indenting.
 
 PHP [keywords][] MUST be in lower case.
 
-The PHP types and keywords `array`, `int`, `true`, `object`, `float`, `false`, `mixed`,
+The PHP types and keywords `new`, `array`, `int`, `true`, `object`, `float`, `false`, `mixed`,
 `bool`, `null`, `numeric`, `string`, `void` and `resource` MUST be in lower case.
 
 Short form of type keywords MUST be used i.e. `bool` instead of `boolean`,
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -109,10 +109,7 @@ tabs for indenting.
 
 ### 2.5 Keywords and Types
 
-PHP [keywords][] MUST be in lower case.
-
-The PHP types and keywords `new`, `array`, `int`, `true`, `object`, `float`, `false`, `mixed`,
-`bool`, `null`, `numeric`, `string`, `void` and `resource` MUST be in lower case.
+All PHP reserved types and keywords [[1]][keywords][[2]][types] MUST be in lower case.
 
 Short form of type keywords MUST be used i.e. `bool` instead of `boolean`,
 `int` instead of `integer` etc.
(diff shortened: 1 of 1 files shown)


KorvinSzanto:
I'm confused, `new` is not a type keyword. Type keywords are things like `bool` `int` `void` etc. They are not listed with the other reserved keywords which is why they are listed separately in this link. 

I think the statement above the line you changed is already covering `new`, just click through the link:

> PHP [keywords](https://secure.php.net/manual/en/reserved.keywords.php) MUST be in lower case.

proposed/extended-coding-style-guide.md
```diff
@@ -111,7 +111,7 @@ tabs for indenting.
 
 PHP [keywords][] MUST be in lower case.
```
Jean85: This line already covers `new`, but...

proposed/extended-coding-style-guide.md
```diff
@@ -111,7 +111,7 @@ tabs for indenting.
...
 
-The PHP types and keywords `array`, `int`, `true`, `object`, `float`, `false`, `mixed`,
+The PHP types and keywords `new`, `array`, `int`, `true`, `object`, `float`, `false`, `mixed`,
```
Jean85: ..maybe we can improve this line, linking to https://secure.php.net/manual/en/reserved.other-reserved-words.php, so we can indirectly list them all, even future ones.

KorvinSzanto:
We discussed this change a little more and decided that it'd be better to defer to the documentation for the list of type keywords rather than try to itemize them ourselves. The reason they were originally itemized is that this document predates any documentation for type keywords.

Here's how I think this should end up:

```diff --git a/proposed/extended-coding-style-guide.md b/proposed/extended-coding-style-guide.md
index 1bddf60..e7430f4 100644
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -112,10 +112,7 @@ tabs for indenting.
 
 ### 2.5 Keywords and Types
 
-PHP [keywords][] MUST be in lower case.
-
-The PHP types and keywords `array`, `int`, `true`, `object`, `float`, `false`, `mixed`,
-`bool`, `null`, `numeric`, `string`, `void` and `resource` MUST be in lower case.
+PHP reserved keywords [[1]][keywords1][[2]][keywords2] MUST be in lower case.
 
 Short form of type keywords MUST be used i.e. `bool` instead of `boolean`,
 `int` instead of `integer` etc.
@@ -918,7 +915,8 @@ $instance = new class extends \Foo implements
 
 [PSR-1]: http://www.php-fig.org/psr/psr-1/
 [PSR-2]: http://www.php-fig.org/psr/psr-2/
-[keywords]: http://php.net/manual/en/reserved.keywords.php
+[keywords1]: http://php.net/manual/en/reserved.keywords.php
+[keywords2]: https://secure.php.net/manual/en/reserved.other-reserved-words.php
 [arithmetic]: http://php.net/manual/en/language.operators.arithmetic.php
 [assignment]: http://php.net/manual/en/language.operators.assignment.php
 [comparison]: http://php.net/manual/en/language.operators.comparison.php
```

JABirchall:
Maybe a secondary improvement ontop of @KorvinSzanto's comment would be to reference All keywords.

`All PHP reserved keywords [[1]][keywords1][[2]][keywords2] MUST be in lower case`

KorvinSzanto:
@JABirchall are any of those keywords not covered by the two included links?

JABirchall:
@KorvinSzanto No, PHP does well to keep their docs uptodate.

`All PHP reserved types and keywords [[1]][keywords1][[2]][keywords2] MUST be in lower case.` would suffice.

JABirchall:
I have added another commit with stricter wording: https://github.com/php-fig/fig-standards/pull/935/commits/29969c43e5d61651b29195bd865caccd0ea2ced4

JABirchall:
Also ontop of @Jean85's comment

>I agree with that. BTW we should consider if we should list the keywords or just reference them all pointing to the docs.

I think it would be best to list all the keywords in the PSR, incase PHP decide to change the page keywords are on as to not lose the integrity of the PSR.

samdark:
@JABirchall not listing keywords and linking the page instead would be a bit more future-proof.

PR #938, opened 2017-10-02
Add a section for compound boolean statements

Jean85:
This PR is an attempt to rework #918.

 * I tried to simplify it by creating a dedicated section
 * I split the rules with practical examples
 * I also changed the rule about internal elements that are splitted in multiple lines: IMHO it should count anyway as a splitted compound statement, and we should split everything at that point, for the sake of readability
 * I do not enforce a new line for the first element of the expression: this is useful if it's a return or assignment statement

### DIFF
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -928,7 +928,7 @@ at the beginning of each line.
 If the inner elements of compound expression are to be splitted, the parenthesis 
 MUST stay on the same line of the logical operator, and each inner boolean
 expression MUST be splitted with the same rules, but with an additional
-level of indentation; the closing parenthesis MUST stay on their own line.
+level of indentation; the closing parenthesis MUST stay on its own line.
 
 ~~~php
     return $firstBoolean
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -905,6 +905,68 @@ $instance = new class extends \Foo implements
 };
 ~~~
 
+## 9. Compound boolean statements
+
+A compound boolean expression MAY be split across multiple lines. When splitted, 
+there MUST be only one boolean expression per line, or an other compound boolean
+expression that MUST be enclosed by parentheses. The logical operator MUST be placed 
+at the beginning of each line. 
+
+~~~php
+    return $firstBoolean
+        && ($secondBoolean || $thirdBoolean)
+        && $fourthBoolean;
+~~~
+~~~php
+    $otherBoolean = in_array($foo, [
+            'foo',
+            'bar',
+        ]) 
+        && $baz;
+~~~
+
+If the inner elements of compound expression are to be splitted, the parenthesis 
+MUST stay on the same line of the logical operator, and each inner boolean
+expression MUST be splitted with the same rules, but with an additional
+level of indentation; the closing parenthesis MUST stay on their own line.
+
+~~~php
+    return $firstBoolean
+        && (
+            $secondBoolean 
+            || $thirdBoolean
+        )
+        && $fourthBoolean;
+~~~
+
+When used in a control structure, splitted compound boolean expressions MUST 
+start on a new line, leaving the opening and closing parentheses of the structure
+on their own lines.
+
+~~~php
+if (
+    $foo
+    && $bar
+    || ($baz < 0 && false !== strpos('foobarbazqux', $qux))
+    || in_array($qux, [
+        'foo',
+        'bar',
+    ])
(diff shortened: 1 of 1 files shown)


proposed/extended-coding-style-guide.md
```diff
@@ -905,6 +905,68 @@ $instance = new class extends \Foo implements
 };
 ~~~
 
+## 9. Compound boolean statements
+
+A compound boolean expression MAY be split across multiple lines. When splitted,
```
JABirchall: > When splitted,

I think you mean 'When split,'

proposed/extended-coding-style-guide.md
```diff
@@ -905,6 +905,68 @@ $instance = new class extends \Foo implements
...
+## 9. Compound boolean statements
+
+A compound boolean expression MAY be split across multiple lines. When splitted, 
+there MUST be only one boolean expression per line, or an other compound boolean
+expression that MUST be enclosed by parentheses. The logical operator MUST be placed 
+at the beginning of each line. 
+
+~~~php
+    return $firstBoolean
+        && ($secondBoolean || $thirdBoolean)
```
samdark: Aren't there *two* boolean expressions per line?

proposed/extended-coding-style-guide.md
```diff
@@ -905,6 +905,68 @@ $instance = new class extends \Foo implements
...
+            'foo',
+            'bar',
+        ]) 
+        && $baz;
+~~~
+
+If the inner elements of compound expression are to be splitted, the parenthesis 
+MUST stay on the same line of the logical operator, and each inner boolean
+expression MUST be splitted with the same rules, but with an additional
+level of indentation; the closing parenthesis MUST stay on their own line.
```
julienfalque: _... the closing parenthesis MUST stay on **its** own line._

julienfalque:
> I also changed the rule about internal elements that are splitted in multiple lines: IMHO it should count anyway as a splitted compound statement, and we should split everything at that point, for the sake of readability

I'm not sure I understand. Does that mean any multiline element (e.g. an array, like in one of the examples) forces the compound boolean expression itself to be multiline?

> I do not enforce a new line for the first element of the expression: this is useful if it's a return or assignment statement

I prefer forcing the new line for the first element just like with control structure, but why not? :)

proposed/extended-coding-style-guide.md
Jean85: Right! Sorry, I'm not a native speaker, fixing it now!

proposed/extended-coding-style-guide.md
Jean85: Yes but:

> there MUST be only one boolean expression per line, **or an other compound boolean expression** that MUST be enclosed by parentheses.

I don't want to force the splitting for every single element, that example is more than readable, and it emphasize that the inner compound element is different from the others (it uses `||` instead of `&&`)

Jean85:
> I'm not sure I understand. Does that mean any multiline element (e.g. an array, like in one of the examples) forces the compound boolean expression itself to be multiline?

Yes, exactly. I'm suggesting that because I find the following example extremely unreadable:

```php
if (in_array($foo, [
    'foo',
    'bar',
]) && $baz) {
    // body
}
```

julienfalque:
> Well, forcing the first new line should not be forbidden by this verbiage, so everybody is happy :smile:

Right :+1: 

In your example, swapping the conditions makes it readable IMO:
```php
if ($baz && in_array($foo, [
    'foo',
    'bar',
])) {
    // body
}
```
so I would drop that rule as well. One may still force newlines if desired, just like one may still force newlines for mulitline conditions in a `return` statement.

Jean85:
I still would be against relaxing that rule, it would allow a lot of unreadable stuff. But it's not up to me to decide :D anyone else wants to weight in?

proposed/extended-coding-style-guide.md
samdark: OK.

KorvinSzanto:
Thanks for writing this up @Jean85! I started a discussion in slack but I figured I should post here too.

I am thinking we should leave this out of scope. None of the examples here are particularly more readable to me with the significant amount of complexity this adds to this documentation. Further I'm thinking if your conditionals are getting as complicated as some of the examples, your code has bigger issues.

IMO Implementors can make a decision on how they'd like to format their conditionals on their own.

Jean85:
Thanks Korvin. I undestand your concerns, and I agree that this looks like more trouble than its worth. I'm closing this, at least we tried and we know that it's not feasible.

PR #942, opened 2017-10-21
Added note about async non-goal

Nyholm:
This will replace https://github.com/php-http/fig-standards/pull/31

### DIFF
--- a/proposed/http-client/http-client-meta.md
+++ b/proposed/http-client/http-client-meta.md
@@ -37,7 +37,7 @@ specify the default behaviours.
 
 The reason asynchronous support is a non-goal is because a PSR for HTTP clients 
 should not define Promises. That should be a separate PSR. At the time the HTTP 
-client PSR was written there where no PSR for Promises. 
+client PSR was written there was no PSR for Promises. 
 
 A separate interface for asynchronous requests can be defined in a separate PSR 
 once the promise PSR is accepted. The method signature for asynchronous requests 
--- a/proposed/http-client/http-client-meta.md
+++ b/proposed/http-client/http-client-meta.md
@@ -41,7 +41,8 @@ would delay the HTTP client PSR.
 
 A separate interface for asynchronous requests can be defined in a separate PSR 
 once the promise PSR is accepted. The method signature for asynchronous requests 
-has to be different.
+has to be different from the method signature for synchronous requests because 
+the return types of asynchronous calls will be a Promise.
 
 ## Approaches
 
--- a/proposed/http-client/http-client-meta.md
+++ b/proposed/http-client/http-client-meta.md
@@ -36,8 +36,8 @@ specify the default behaviours.
 #### Asynchronous HTTP client
 
 The reason asynchronous support is a non-goal is because a PSR for HTTP clients 
-should not define Promises. That should be a separate PSR. Waiting for a such PSR 
-would delay the HTTP client PSR.
+should not define Promises. That should be a separate PSR. At the time the HTTP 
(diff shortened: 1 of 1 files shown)


michaelcullum:
:+1:

proposed/http-client/http-client-meta.md
```diff
@@ -33,6 +33,15 @@ client will allow libraries to be decoupled from an implementation such as Guzzl
...
 
+#### Asynchronous HTTP client
+
+The reason asynchronous support is a non-goal is because a PSR for HTTP clients 
+should not define Promises. That should be a separate PSR. Waiting for a such PSR 
+would delay the HTTP client PSR.
+
+A separate interface for asynchronous requests can be defined in a separate PSR 
+once the promise PSR is accepted. The method signature for asynchronous requests 
+has to be different.
```
dbu: maybe:
...has to be different from the method signature for synchronous requests because the return type has to be different anyways (Promise vs Response).

Nyholm: Great. Thanks1

proposed/http-client/http-client-meta.md
```diff
@@ -33,6 +33,16 @@ client will allow libraries to be decoupled from an implementation such as Guzzl
 specify the default behaviours.
 * The purpose is not to be opinionated about the use of middlewares (PSR-15).
 
+#### Asynchronous HTTP client
+
+The reason asynchronous support is a non-goal is because a PSR for HTTP clients 
+should not define Promises. That should be a separate PSR. Waiting for a such PSR
```
sagikazarmark: Thinking about a wait part: it only really matters while the PSR is WIP, but afterwards it's not an information. The fact that it should be a separate PSR and at the time of writing there was no such PSR should be enough to be mentioned IMO.

Nyholm:
@sagikazarmark, better now?

proposed/http-client/http-client-meta.md
```diff
@@ -33,6 +33,16 @@ client will allow libraries to be decoupled from an implementation such as Guzzl
 specify the default behaviours.
 * The purpose is not to be opinionated about the use of middlewares (PSR-15).
 
+#### Asynchronous HTTP client
+
+The reason asynchronous support is a non-goal is because a PSR for HTTP clients 
+should not define Promises. That should be a separate PSR. At the time the HTTP 
+client PSR was written there where no PSR for Promises.
```
sagikazarmark: there *was* no such PSR

sagikazarmark:
Definitely. Just one last small thing.

proposed/http-client/http-client-meta.md
Nyholm: oops, was way to quick there. Thanks

proposed/http-client/http-client-meta.md
```diff
@@ -33,6 +33,16 @@ client will allow libraries to be decoupled from an implementation such as Guzzl
...
+#### Asynchronous HTTP client
+
+The reason asynchronous support is a non-goal is because a PSR for HTTP clients 
+should not define Promises. That should be a separate PSR. At the time the HTTP 
+client PSR was written there was no PSR for Promises. 
+
+A separate interface for asynchronous requests can be defined in a separate PSR 
+once the promise PSR is accepted. The method signature for asynchronous requests 
+has to be different from the method signature for synchronous requests because 
+the return types of asynchronous calls will be a Promise.
```
dbu: "the return type of asynchronous calls" (return type should be singular here, not plural)

PR #943, opened 2017-10-21
Remove `Http` prefix from `HttpClient`.

Nyholm:
This will replace https://github.com/php-http/fig-standards/pull/42

@dbu or @sagikazarmark, could you approve this again?

### DIFF
--- a/proposed/http-client/http-client.md
+++ b/proposed/http-client/http-client.md
@@ -30,7 +30,7 @@ should not follow redirect nor throw exceptions on HTTP responses with status 4x
 The following interfaces MAY be implemented together within a single class or
 in separate classes.
 
-### ClientInterface
+### Client interface
 
 ```php
 namespace Psr\Http\Client;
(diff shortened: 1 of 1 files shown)


proposed/http-client/http-client.md
```diff
@@ -30,15 +30,15 @@ should not follow redirect nor throw exceptions on HTTP responses with status 4x
...
+### ClientInterface
 
 ```php
 namespace Psr\Http\Client;
 
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
 
-interface HttpClient
+interface Client
```
dbu: is PSR standard naming with Interface suffix or without? i think we said we do without suffix, but then the title of this section should be `### Client interface` i think.

Nyholm: Thank you

sagikazarmark: Ah, I was hoping noone notices :trollface:

michaelcullum: Will need to be `ClientInterface` yep

Nyholm: Thank you

PR #944, opened 2017-10-22
Added more clear specification.

Nyholm:
Im trying to address @joelwurtz [comment](https://github.com/php-http/fig-standards/issues/24) and part of https://github.com/php-http/fig-standards/pull/34

### DIFF
--- a/proposed/http-client/http-client.md
+++ b/proposed/http-client/http-client.md
@@ -23,11 +23,17 @@ encoding and adjust the Content-Length header.
 
 An implementing library MUST implement `Psr\Http\Client\ClientException` for each exception it throws. 
 
-When there is an error with the network or the remote server cannot be reached, the
-HTTP client MUST thrown a `Psr\Http\Client\Exception\NetworkException`. If the 
-request is invalid and cannot be sent the HTTP client MUST throw a 
-`Psr\Http\Client\Exception\RequestException`. Smaller issues like wrong HTTP version
-is not blocking the HTTP client to send the request and MUST not cause any exception. 
+When the client get a request that is invalid and cannot be sent, the HTTP client 
+MUST throw a `Psr\Http\Client\Exception\RequestException`. If there is an error
+with the network or the remote server cannot be reached, the HTTP client MUST throw
+a `Psr\Http\Client\Exception\NetworkException`. 
+
+Smaller issues like wrong HTTP version is not blocking the HTTP client to send the
+request and MUST not cause any exception. 
+
+No exception should be thrown if the remote server answers with an response that can
+be parsed onto a PSR-7 Response. For example a 400 or 500 response should not cause 
+an exception.
 
 ## Goal
 
--- a/proposed/http-client/http-client.md
+++ b/proposed/http-client/http-client.md
@@ -24,10 +24,10 @@ encoding and adjust the Content-Length header.
 An implementing library MUST implement `Psr\Http\Client\ClientException` for each exception it throws. 
 
 When there is an error with the network or the remote server cannot be reached, the
-HTTP client SHOULD thrown a `Psr\Http\Client\Exception\NetworkException`. If the 
-request is invalid and cannot be sent the HTTP client SHOULD throw a 
+HTTP client MUST thrown a `Psr\Http\Client\Exception\NetworkException`. If the 
+request is invalid and cannot be sent the HTTP client MUST throw a 
 `Psr\Http\Client\Exception\RequestException`. Smaller issues like wrong HTTP version
-is not blocking the HTTP client to send the request and SHOULD not cause any exception. 
+is not blocking the HTTP client to send the request and MUST not cause any exception. 
 
 ## Goal
 
--- a/proposed/http-client/http-client.md
+++ b/proposed/http-client/http-client.md
@@ -3,16 +3,27 @@ HTTP Client
 
 This document describes common interfaces for sending HTTP messages.
 
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
 
 ## Specification
 
+### Client
+
 An HTTP client has the responsibility to send a PSR-7 request and return a PSR-7
-response. When there is an error during sending the request or an error with network an
-exception should be thrown. 
+response. The HTTP client MAY modify the response from the server before returning 
+the PSR-7 response. 
+
+### Exceptions
 
 An implementing library MUST implement `Psr\Http\Client\ClientException` for each exception it throws. 
-It SHOULD implement exceptions for `Psr\Http\Client\Exception\NetworkException` and
-`Psr\Http\Client\Exception\RequestException`.
+
+When there is an error with the network or the remote server cannot be reached, the
+HTTP client SHOULD thrown a `Psr\Http\Client\Exception\NetworkException`. If the 
+request is invalid and cannot be sent the HTTP client SHOULD throw a 
+`Psr\Http\Client\Exception\RequestException`. Smaller issues like wrong HTTP version
+is not blocking the HTTP client to send the request and SHOULD not cause any exception. 
 
 ## Goal
 
--- a/proposed/http-client/http-client.md
+++ b/proposed/http-client/http-client.md
@@ -17,7 +17,7 @@ from the user/server. In this case the request and the response MUST be consiste
 between the body and headers. For example a 
 server may return a gzip encoded body and the client may know how to decode this, 
 when it decodes the body the client MUST also remove the header that specifies the 
-encoding.
+encoding and adjust the Content-Length header. 
 
 ### Exceptions
 
--- a/proposed/http-client/http-client.md
+++ b/proposed/http-client/http-client.md
@@ -12,11 +12,11 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
 ### Client
 
 An HTTP client has the responsibility to send a PSR-7 request and return a PSR-7
-response. The HTTP client MAY modify the request, given by the user and sent to 
-the server or the response from the server before returning. In this case the 
-request and the response must be consistent between the body and headers. I.e. a 
+response. Under the hood the HTTP client MAY modify the request/response received 
+from the user/server. In this case the request and the response MUST be consistent
+between the body and headers. For example a 
 server may return a gzip encoded body and the client may know how to decode this, 
-when it decodes the body the client MUST also remove the header that specifies this 
+when it decodes the body the client MUST also remove the header that specifies the 
 encoding.
 
 ### Exceptions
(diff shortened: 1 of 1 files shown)


proposed/http-client/http-client.md
```diff
@@ -3,16 +3,27 @@ HTTP Client
...
 
 ## Specification
 
+### Client
+
 An HTTP client has the responsibility to send a PSR-7 request and return a PSR-7
-response. When there is an error during sending the request or an error with network an
-exception should be thrown. 
+response. The HTTP client MAY modify the response from the server before returning 
+the PSR-7 response.
```
joelwurtz: Something like this (minus typo ^^) would be better IMO:

"The HTTP client MAY modify the request, given by the user and sent to the server or the response from the server before returning. In this case the request and the response must be consistent between the body and headers. I.e. a server may return a gzip encoded body and the client may know how to decode this, when it decodes the body the client MUST also remove the header that specify this encoding."

dbu: if this is what we mean by this, then yes lets give this example to illustrate what we actually mean.

proposed/http-client/http-client.md
```diff
@@ -3,16 +3,31 @@ HTTP Client
...
+### Client
+
 An HTTP client has the responsibility to send a PSR-7 request and return a PSR-7
-response. When there is an error during sending the request or an error with network an
-exception should be thrown. 
+response. The HTTP client MAY modify the request, given by the user and sent to 
+the server or the response from the server before returning. In this case the 
+request and the response must be consistent between the body and headers. I.e. a 
+server may return a gzip encoded body and the client may know how to decode this, 
+when it decodes the body the client MUST also remove the header that specify this
```
dbu: i think this line mixes singular and plural. either "remove the headers that specify" or "remove the header that specifies". i think the second makes more sense, the encoding example would have just one header.

Nyholm: thanks

proposed/http-client/http-client.md
```diff
@@ -12,8 +12,12 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
 ### Client
 
 An HTTP client has the responsibility to send a PSR-7 request and return a PSR-7
-response. The HTTP client MAY modify the response from the server before returning 
-the PSR-7 response. 
+response. The HTTP client MAY modify the request, given by the user and sent to
```
sagikazarmark: "Under the hood the HTTP client MAY modify the request/response received from the user/server."

How does that sound?

This part tries to make it clear that under the contract surface anything could happen to request/response objects and there is no assurance that they will remain the same objects on the instance level, right?

proposed/http-client/http-client.md
```diff
@@ -12,8 +12,12 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
 ### Client
 
 An HTTP client has the responsibility to send a PSR-7 request and return a PSR-7
-response. The HTTP client MAY modify the response from the server before returning 
-the PSR-7 response. 
+response. The HTTP client MAY modify the request, given by the user and sent to 
+the server or the response from the server before returning. In this case the 
+request and the response must be consistent between the body and headers. I.e. a
```
sagikazarmark: I.e. -> "For example," ?

proposed/http-client/http-client.md
```diff
@@ -12,8 +12,12 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
 ### Client
 
 An HTTP client has the responsibility to send a PSR-7 request and return a PSR-7
-response. The HTTP client MAY modify the response from the server before returning 
-the PSR-7 response. 
+response. The HTTP client MAY modify the request, given by the user and sent to 
+the server or the response from the server before returning. In this case the 
+request and the response must be consistent between the body and headers. I.e. a 
+server may return a gzip encoded body and the client may know how to decode this, 
+when it decodes the body the client MUST also remove the header that specify this
```
sagikazarmark: specify **the** encoding?

proposed/http-client/http-client.md
```diff
@@ -3,16 +3,31 @@ HTTP Client
...
+encoding.
+
+### Exceptions
 
 An implementing library MUST implement `Psr\Http\Client\ClientException` for each exception it throws. 
-It SHOULD implement exceptions for `Psr\Http\Client\Exception\NetworkException` and
-`Psr\Http\Client\Exception\RequestException`.
+
+When there is an error with the network or the remote server cannot be reached, the
+HTTP client SHOULD thrown a `Psr\Http\Client\Exception\NetworkException`. If the
```
michaelcullum: `thrown` -> `throw`

Generally as a rule, `SHOULD` is recommended against as it cannot be relied upon by a client anymore. Is there rationale as to why this isn't a `MUST`?

proposed/http-client/http-client.md
```diff
@@ -3,16 +3,31 @@ HTTP Client
...
 ## Specification
 
+### Client
+
 An HTTP client has the responsibility to send a PSR-7 request and return a PSR-7
-response. When there is an error during sending the request or an error with network an
-exception should be thrown. 
+response. The HTTP client MAY modify the request, given by the user and sent to 
+the server or the response from the server before returning. In this case the 
+request and the response must be consistent between the body and headers. I.e. a
```
michaelcullum: `I.e.` doesn't make much sense in this context. Perhaps use `For example` instead.

proposed/http-client/http-client.md
```diff
@@ -3,16 +3,31 @@ HTTP Client
...
+### Client
+
 An HTTP client has the responsibility to send a PSR-7 request and return a PSR-7
-response. When there is an error during sending the request or an error with network an
-exception should be thrown. 
+response. The HTTP client MAY modify the request, given by the user and sent to 
+the server or the response from the server before returning. In this case the 
+request and the response must be consistent between the body and headers. I.e. a 
+server may return a gzip encoded body and the client may know how to decode this, 
+when it decodes the body the client MUST also remove the header that specifies this
```
michaelcullum: using a `MUST` inside an example doesn't make much sense. Suggestion:
`and the client may know how to decode this, however the client should then also mutate the headers that specify the encoding`

dbu: hm, the MUST is important here. should we move that out of the example to say that if the client modifies the request or response, it MUST also adjust the headers accordingly? (i'd assume that if it gzips a request, it must also adjust the content-length header, for example)

proposed/http-client/http-client.md
dbu: in some circumstances it might be difficult to know what kind of error happened. there will probably always be some fuzzyness around the exception type. however, we are specific as to which things cause a network exception, so i think MUST should be fine here.

proposed/http-client/http-client.md
michaelcullum: 👍

proposed/http-client/http-client.md
```diff
@@ -12,11 +12,11 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
...
-response. The HTTP client MAY modify the request, given by the user and sent to 
-the server or the response from the server before returning. In this case the 
-request and the response must be consistent between the body and headers. I.e. a 
+response. Under the hood the HTTP client MAY modify the request/response received 
+from the user/server. In this case the request and the response MUST be consistent
+between the body and headers. For example a 
 server may return a gzip encoded body and the client may know how to decode this, 
-when it decodes the body the client MUST also remove the header that specifies this 
+when it decodes the body the client MUST also remove the header that specifies the 
 encoding.
```
joelwurtz: and update the content length header.

Nyholm: Thank you

proposed/http-client/http-client.md
dbu: can you fix the thrown throw please?

ok if we say MUST here?

proposed/http-client/http-client.md
```diff
@@ -3,16 +3,31 @@ HTTP Client
...
+encoding and adjust the Content-Length header. 
+
+### Exceptions
 
 An implementing library MUST implement `Psr\Http\Client\ClientException` for each exception it throws. 
-It SHOULD implement exceptions for `Psr\Http\Client\Exception\NetworkException` and
-`Psr\Http\Client\Exception\RequestException`.
+
+When there is an error with the network or the remote server cannot be reached, the
+HTTP client MUST thrown a `Psr\Http\Client\Exception\NetworkException`. If the
```
dbu: s/thrown/throw

proposed/http-client/http-client.md
```diff
@@ -3,16 +3,31 @@ HTTP Client
...
+### Exceptions
 
 An implementing library MUST implement `Psr\Http\Client\ClientException` for each exception it throws. 
-It SHOULD implement exceptions for `Psr\Http\Client\Exception\NetworkException` and
-`Psr\Http\Client\Exception\RequestException`.
+
+When there is an error with the network or the remote server cannot be reached, the
+HTTP client MUST thrown a `Psr\Http\Client\Exception\NetworkException`. If the 
+request is invalid and cannot be sent the HTTP client MUST throw a 
+`Psr\Http\Client\Exception\RequestException`. Smaller issues like wrong HTTP version
```
dbu: i would swap the 2 points, so they are more in order of when they happen (invalid request is detected first, network problem only later)

proposed/http-client/http-client.md
```diff
@@ -3,16 +3,31 @@ HTTP Client
...
+when it decodes the body the client MUST also remove the header that specifies the 
+encoding and adjust the Content-Length header. 
+
+### Exceptions
 
 An implementing library MUST implement `Psr\Http\Client\ClientException` for each exception it throws. 
-It SHOULD implement exceptions for `Psr\Http\Client\Exception\NetworkException` and
-`Psr\Http\Client\Exception\RequestException`.
+
+When there is an error with the network or the remote server cannot be reached, the
```
dbu: what about an invalid response, should we mention that here? like not proper http or such.

Nyholm:
Thank you @dbu

proposed/http-client/http-client.md
```diff
@@ -3,16 +3,37 @@ HTTP Client
...
+when it decodes the body the client MUST also remove the header that specifies the 
+encoding and adjust the Content-Length header. 
+
+### Exceptions
 
 An implementing library MUST implement `Psr\Http\Client\ClientException` for each exception it throws. 
-It SHOULD implement exceptions for `Psr\Http\Client\Exception\NetworkException` and
-`Psr\Http\Client\Exception\RequestException`.
+
+When the client get a request that is invalid and cannot be sent, the HTTP client
```
dbu: When the HTTP client is passed a request that ... sent , the client

proposed/http-client/http-client.md
```diff
@@ -3,16 +3,37 @@ HTTP Client
...
+
+When the client get a request that is invalid and cannot be sent, the HTTP client 
+MUST throw a `Psr\Http\Client\Exception\RequestException`. If there is an error
+with the network or the remote server cannot be reached, the HTTP client MUST throw
+a `Psr\Http\Client\Exception\NetworkException`. 
+
+Smaller issues like wrong HTTP version is not blocking the HTTP client to send the
+request and MUST not cause any exception. 
+
+No exception should be thrown if the remote server answers with an response that can
```
dbu: If the remote server answers with a response that can be parsed into a PSR-7 response, the client MUST NOT throw an exception. For example, response status codes in the 400 and 500 range MUST NOT cause an exception.

Nyholm:
I've addressed their issues. Im merging this to move forward. Feel free to open issues or PRs to make additional changes.

PR #945, opened 2017-10-22
A HTTP client must never throw exception on 4xx or 5xx

Nyholm:
This will address issues in https://github.com/php-http/fig-standards/pull/34, https://github.com/php-http/fig-standards/issues/41 and https://github.com/php-http/fig-standards/issues/28

### DIFF
--- a/proposed/http-client/http-client-meta.md
+++ b/proposed/http-client/http-client-meta.md
@@ -41,8 +41,7 @@ specify the default behaviours.
 The intention of this PSR is to provide library developers with HTTP clients that 
 have a well defined behaviour. A library should be able to use any compliant client
 without special code to handle client implementation details (Liskov substitution
-principle). The PSR does not try to restrict nor define configuration interface 
-for HTTP clients. 
+principle). The PSR does not try to restrict nor define how to configure HTTP clients. 
 
 An alternative approach would be to pass configuration to the client. That approach
 would have a few drawbacks: 
--- a/proposed/http-client/http-client-meta.md
+++ b/proposed/http-client/http-client-meta.md
@@ -39,10 +39,8 @@ specify the default behaviours.
 ### Default behavior
 
 The intention of this PSR is ensure library developers that all HTTP clients have the same 
-**default behavior**. That means that all HTTP clients MUST follow Liskov substitution principle
-when no configuration is provided. The PSR does not try to restrict nor define configuration for 
-HTTP clients. An implementing library is free to be configured by the application author to follow
-redirects, to throw exceptions or any other possible setting.  
+behavior. That means that all HTTP clients MUST follow Liskov substitution principle. 
+The PSR does not try to restrict nor define configuration for HTTP clients. 
 
 An alternative approach would be to pass configuration to the client. That approach would have
 a few drawbacks: 
@@ -53,9 +51,9 @@ a few drawbacks:
 
 ### Exceptions
 
-The domain exceptions `NetworkException`, `RequestException` and `HttpException` define
+The domain exceptions `NetworkException` and `RequestException` define
 a contract very similar to each other. The chosen approach is to not let them extend each other
-because inheritance does not make sense in the domain model. A `RequestException` is not a
+because inheritance does not make sense in the domain model. A `RequestException` is simply not a
 `NetworkException`.
 
 Allowing exception to extend a `RequestAwareException` and/or `ResponseAwareException` interface
--- a/proposed/http-client/http-client-meta.md
+++ b/proposed/http-client/http-client-meta.md
@@ -38,12 +38,14 @@ specify the default behaviours.
 
 ### Default behavior
 
-The intention of this PSR is ensure library developers that all HTTP clients have the same 
-behavior. That means that all HTTP clients MUST follow Liskov substitution principle. 
-The PSR does not try to restrict nor define configuration for HTTP clients. 
+The intention of this PSR is to provide library developers with HTTP clients that 
+have a well defined behaviour. A library should be able to use any compliant client
+without special code to handle client implementation details (Liskov substitution
+principle). The PSR does not try to restrict nor define configuration interface 
+for HTTP clients. 
 
-An alternative approach would be to pass configuration to the client. That approach would have
-a few drawbacks: 
+An alternative approach would be to pass configuration to the client. That approach
+would have a few drawbacks: 
 
 * Configuration must be defined by the PSR
 * All client must support the defined configuration
--- a/proposed/http-client/http-client.md
+++ b/proposed/http-client/http-client.md
@@ -41,12 +41,9 @@ use Psr\Http\Message\ResponseInterface;
 interface ClientInterface
 {
     /**
-     * Sends a PSR-7 request.
-     *
-     * If a request is sent without any prior configuration, an exception MUST NOT be thrown
-     * when a response is recieved, no matter the HTTP status code.
-     *
-     * If a request is sent without any prior configuration, a HTTP client MUST NOT follow redirects.
+     * Sends a PSR-7 request and returns a PSR-7 response. No response will get a special treatment. Ie: 
+     *   - Redirects will not be followed
+     *   - No exceptions will be thrown on 4xx or 5xx responses 
      *
      * The client MAY do modifications to the Request before sending it. Because PSR-7 objects are
      * immutable, one cannot assume that the object passed to Client::sendRequest will be the same
@@ -136,39 +133,5 @@ interface NetworkException extends ClientException
 }
 ```
 
-
-### HttpException
-
-```php
-namespace Psr\Http\Client\Exception;
-
-use Psr\Http\Message\RequestInterface;
-use Psr\Http\Client\ClientException;
-
-/**
- * Thrown when a response was received but the request itself failed.
- *
- * This exception MAY be thrown on HTTP response codes 4xx and 5xx.
- * This exception MUST NOT be thrown when using the client's default configuration.
- */
-interface HttpException extends ClientException
-{
-    /**
-     * Returns the request.
-     *
-     * The request object MAY be a different object from the one passed to Client::sendRequest()
-     *
-     * @return RequestInterface
-     */
-    public function getRequest();
-
-    /**
-     * Returns the response.
-     *
-     * @return ResponseInterface
-     */
-    public function getResponse();
-}
-```
 
 [Liskov]: https://en.wikipedia.org/wiki/Liskov_substitution_principle
--- a/proposed/http-client/http-client.md
+++ b/proposed/http-client/http-client.md
@@ -41,7 +41,7 @@ use Psr\Http\Message\ResponseInterface;
 interface ClientInterface
 {
     /**
-     * Sends a PSR-7 request and returns a PSR-7 response. Every technically correct HTTP respons 
+     * Sends a PSR-7 request and returns a PSR-7 response. Every technically correct HTTP response 
      * MUST be returned as is, even if it represents a HTTP error response or a redirect instruction.
      *
      * The client MAY do modifications to the Request before sending it. Because PSR-7 objects are
--- a/proposed/http-client/http-client.md
+++ b/proposed/http-client/http-client.md
@@ -44,8 +44,8 @@ interface ClientInterface
      * Sends a PSR-7 request and returns a PSR-7 response. 
      * 
      * Every technically correct HTTP response MUST be returned as is, even if it represents a HTTP 
-     * error response or a redirect instruction. The only exception is 100 request which MUST be 
-     * put together in the HTTP client. 
+     * error response or a redirect instruction. The only exception is 1xx responses which MUST be 
+     * assembled in the HTTP client. 
      *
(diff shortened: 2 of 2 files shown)


proposed/http-client/http-client-meta.md
```diff
@@ -39,10 +39,8 @@ specify the default behaviours.
 ### Default behavior
 
 The intention of this PSR is ensure library developers that all HTTP clients have the same 
-**default behavior**. That means that all HTTP clients MUST follow Liskov substitution principle
-when no configuration is provided. The PSR does not try to restrict nor define configuration for 
-HTTP clients. An implementing library is free to be configured by the application author to follow
-redirects, to throw exceptions or any other possible setting.  
+behavior. That means that all HTTP clients MUST follow Liskov substitution principle.
```
sagikazarmark: This is kind of obvious to me, I mean that's why we want to create this standard. Also, the next sentence kinda goes against this statement.

So I would rather say something about not restricting nor defining a configuration interface, but clients need to follow the contract by default (which might definitely be more than just fulfilling the interface, see exception handling, etc) even if they offer options that might break it.

Not sure though if it's a good idea to mention breaking the contract at all, so we could just leave this section out?

proposed/http-client/http-client.md
```diff
@@ -41,12 +41,9 @@ use Psr\Http\Message\ResponseInterface;
 interface ClientInterface
 {
     /**
-     * Sends a PSR-7 request.
-     *
-     * If a request is sent without any prior configuration, an exception MUST NOT be thrown
-     * when a response is recieved, no matter the HTTP status code.
-     *
-     * If a request is sent without any prior configuration, a HTTP client MUST NOT follow redirects.
+     * Sends a PSR-7 request and returns a PSR-7 response. No response will get a special treatment. Ie:
```
sagikazarmark: > No response will get a special treatment.

Not sure what this means.

proposed/http-client/http-client-meta.md
```diff
@@ -53,9 +51,9 @@ a few drawbacks:
 
 ### Exceptions
 
-The domain exceptions `NetworkException`, `RequestException` and `HttpException` define
+The domain exceptions `NetworkException` and `RequestException` define
 a contract very similar to each other. The chosen approach is to not let them extend each other
-because inheritance does not make sense in the domain model. A `RequestException` is not a
+because inheritance does not make sense in the domain model. A `RequestException` is simply not a
 `NetworkException`.
```
kelunik: So one always needs two catch blocks?

proposed/http-client/http-client-meta.md
```diff
@@ -39,10 +39,8 @@ specify the default behaviours.
 ### Default behavior
 
 The intention of this PSR is ensure library developers that all HTTP clients have the same
```
dbu: i think "... PSR is ensure library developers" this is not correct english.

How about we do it like this: (also includes a proposal for the Liskov substitution):

 "The intention of this PSR is to provide library developers with HTTP clients that have a well defined behaviour. A library should be able to use any compliant client without special code to handle client implementation details (Liskov substitution principle)."

proposed/http-client/http-client-meta.md
dbu: i like the phrase below as it is now, solves the dilemma well for me.

proposed/http-client/http-client.md
dbu: maybe "Every technically correct HTTP respons MUST be returned as is, even if it represents a HTTP error response or a redirect instruction."

the prosa text should somewhere explain that the PSR client is low level and to guarantee defined behaviour responses may not be acted upon by default (e.g. throwing exceptions for http status codes or redirection). by saying "by default" we do tell that custom configuration, plugins, whatever can violate this.

proposed/http-client/http-client-meta.md
dbu: should we have a HttpClientExceptionInterface or something? i agree that an exception marking all exceptions of the PSR would be good design.

although in this case, the RequestException implies a programming error in the consumer (or possibly the client) or at least missing input validation. NetworkException can mean a misconfiguration (wrong server name) or a temporary network problem. RequestException are thus more likely to require attention from the development team. I might not want to log NetworkException even, but for example retry the action later.

Nyholm: No, only one catch block is enough. There is the `ClientException` that "Every HTTP client related Exception MUST implement this interface."

Nyholm:
Thank you for the reviews. I've updated the PR.

proposed/http-client/http-client-meta.md
dbu: right, sorry, forgot about that. would it make sense to mention this here again? like "(However, both exceptions implement the ClientExceptionInterface if you want to generically catch exceptions while sending a request.)"

proposed/http-client/http-client.md
```diff
@@ -41,9 +41,8 @@ use Psr\Http\Message\ResponseInterface;
 interface ClientInterface
 {
     /**
-     * Sends a PSR-7 request and returns a PSR-7 response. No response will get a special treatment. Ie: 
-     *   - Redirects will not be followed
-     *   - No exceptions will be thrown on 4xx or 5xx responses 
+     * Sends a PSR-7 request and returns a PSR-7 response. Every technically correct HTTP respons
```
joelwurtz: typo respons -> response

proposed/http-client/http-client.md
```diff
@@ -41,9 +41,8 @@ use Psr\Http\Message\ResponseInterface;
 interface ClientInterface
 {
     /**
-     * Sends a PSR-7 request and returns a PSR-7 response. No response will get a special treatment. Ie: 
-     *   - Redirects will not be followed
-     *   - No exceptions will be thrown on 4xx or 5xx responses 
+     * Sends a PSR-7 request and returns a PSR-7 response. Every technically correct HTTP respons 
+     * MUST be returned as is, even if it represents a HTTP error response or a redirect instruction.
```
joelwurtz: Even for the 100 continue ?

Nyholm: Yes, I think so. By the same arguments as we use on 301 responses. 
What do you think?

kelunik: No, definitely not for 100 continue responses, it doesn't make any sense. I have mentioned this earlier.

proposed/http-client/http-client-meta.md
```diff
@@ -38,24 +38,24 @@ specify the default behaviours.
...
 
-The intention of this PSR is ensure library developers that all HTTP clients have the same 
-**default behavior**. That means that all HTTP clients MUST follow Liskov substitution principle
-when no configuration is provided. The PSR does not try to restrict nor define configuration for 
-HTTP clients. An implementing library is free to be configured by the application author to follow
-redirects, to throw exceptions or any other possible setting.  
+The intention of this PSR is to provide library developers with HTTP clients that 
+have a well defined behaviour. A library should be able to use any compliant client
+without special code to handle client implementation details (Liskov substitution
+principle). The PSR does not try to restrict nor define configuration interface
```
dbu: `... define configuration interface for ...` is grammatically not correct. it could be `... define a configuration interface for ...` or `... define how to configure ...`. i would prefer the later, we do not restrict or define how to configure HTTP clients.

proposed/http-client/http-client.md
```diff
@@ -41,12 +41,8 @@ use Psr\Http\Message\ResponseInterface;
 interface ClientInterface
 {
     /**
-     * Sends a PSR-7 request.
-     *
-     * If a request is sent without any prior configuration, an exception MUST NOT be thrown
-     * when a response is recieved, no matter the HTTP status code.
-     *
-     * If a request is sent without any prior configuration, a HTTP client MUST NOT follow redirects.
+     * Sends a PSR-7 request and returns a PSR-7 response. Every technically correct HTTP response
```
kelunik: This should have two newlines before `Every technically...`.

proposed/http-client/http-client.md
dbu: so we would make the exception for status 100 that MUST be handled, but keep 3xx as MUST NOT be handled? i would agree with that. the meta should explain why (and probably mention that plugins then can handle the 3xx to follow redirects)

kelunik: We just have to link to https://tools.ietf.org/html/rfc7231#section-6.2.

dbu:
do you mean to say that clients have to check for the 'Expertise:
100-continue' header?

or just as reasoning why handling 1xx status makes sense?


Am 17.11.2017 4:48 nachm. schrieb "Niklas Keller" <notifications@github.com

proposed/http-client/http-client.md
```diff
@@ -41,12 +41,10 @@ use Psr\Http\Message\ResponseInterface;
...
-     * Sends a PSR-7 request.
-     *
-     * If a request is sent without any prior configuration, an exception MUST NOT be thrown
-     * when a response is recieved, no matter the HTTP status code.
-     *
-     * If a request is sent without any prior configuration, a HTTP client MUST NOT follow redirects.
+     * Sends a PSR-7 request and returns a PSR-7 response. 
+     * 
+     * Every technically correct HTTP response MUST be returned as is, even if it represents a HTTP 
+     * error response or a redirect instruction.
```
dbu: should we be explicit about the 1xx status? i think we should, and we should mandate that the default configuration does resolve 1xx itself. i guess not many people will write low-level debugging tools with php, and when they do they probably either use sockets directly or a specific client that they can tightly control in the tool.

Nyholm:
PR is updated. The client must assemble 100 requests.

proposed/http-client/http-client.md
```diff
@@ -41,12 +41,11 @@ use Psr\Http\Message\ResponseInterface;
...
-     * Sends a PSR-7 request.
-     *
-     * If a request is sent without any prior configuration, an exception MUST NOT be thrown
-     * when a response is recieved, no matter the HTTP status code.
-     *
-     * If a request is sent without any prior configuration, a HTTP client MUST NOT follow redirects.
+     * Sends a PSR-7 request and returns a PSR-7 response. 
+     * 
+     * Every technically correct HTTP response MUST be returned as is, even if it represents a HTTP 
+     * error response or a redirect instruction. The only exception is 100 request which MUST be
```
dbu: The only exception is 1xx responses which MUST be assembled in the HTTP client.

Nyholm: Thank you, I've updated the PR

dbu:
@kelunik @sagikazarmark @joelwurtz can you please check again too, so that we hopefully can merge this soon?

proposed/http-client/http-client.md
```diff
@@ -41,12 +41,11 @@ use Psr\Http\Message\ResponseInterface;
...
-     *
-     * If a request is sent without any prior configuration, an exception MUST NOT be thrown
-     * when a response is recieved, no matter the HTTP status code.
-     *
-     * If a request is sent without any prior configuration, a HTTP client MUST NOT follow redirects.
+     * Sends a PSR-7 request and returns a PSR-7 response. 
+     * 
+     * Every technically correct HTTP response MUST be returned as is, even if it represents a HTTP 
+     * error response or a redirect instruction. The only exception is 1xx responses which MUST be 
+     * assembled in the HTTP client.
```
kelunik: I'm not a native speaker, but `assembled` seems like the wrong word to me. I'd word is something like that:

```
/**
 *
 * Clients MUST NOT differ in behavior depending on the HTTP status code, i.e. they MUST NOT
 * follow redirects or throw exceptions for 4xx or 5xx responses.
 *
 * Note: 1xx responses are intermediate responses and MUST be handled transparently by the
 * client as defined by the HTTP protocol.
 */
```

I do not really understand what you mean with "MUST be handled **transparently** by the client". 
When I read this it seams like you want to show the 100 response. But I know you are arguing for the opposite.

kelunik: If a client handles things *transparently*, the consumer doesn't know about these things. The term is used for [transparent proxies](https://en.wikipedia.org/wiki/Proxy_server#Transparent_proxy) for example, where the client might not even be able to see that a proxy is used.

Nyholm:
Thank you for the reviews. 

@kelunik Im not sure you are correct. However, Im merging this PR since the topic is "A HTTP client must never throw exception on 4xx or 5xx" and 1xx responses is kind of out of scope. 
Please feel free to open a PR with your suggestion and we will discuss it there.

proposed/http-client/http-client.md
dbu: i agree that you moved on tobias, but please do a pull request to refine the wording here @kelunik. maybe try to formulate the part about transparently in a way that does not assume everybody is aware of the definition (it would make sense to me, but from a natural language point of view its not obvious). the way of putting it in a note is also slightly confusing, as before we indicate to never handle status codes. maybe its easier to understand if we say: 

1xx responses are intermediate responses and MUST be handled transparently by the client as defined by the HTTP protocol. The consumer will not notice whether an intermediate 1xx response happened. For all other HTTP status codes, clients MUST NOT differ in behavior. i.e. they MUST NOT follow redirects nor throw exceptions for 4xx or 5xx responses.

PR #950, opened 2017-11-06
Add return type enforcement to PSR-15

shadowhand:
Enforce return types, targeting PHP >= 7.0. Also renames files to reflect the new PSR title.

PR #952, opened 2017-11-06
Adds missing `;` terminating signature definitions

weierophinney:
PR #950 updated the interfaces to use PHP 7 return type hints; however, in doing so, it accidentally dropped the tailing semi-colons in the interface method signatures. This patch simply adds them back.


shadowhand:
Thanks!

PR #959, opened 2017-11-24
nit

acinader:


proposed/http-factory/http-factory-meta.md
```diff
@@ -142,7 +142,7 @@ of `Request`. The implementation listed above is an implementation of `ServerReq
 [slim-uri]: https://github.com/slimphp/Slim/blob/30cfe3c07dac28ec1129c0577e64b90ba11a54c4/Slim/Http/Uri.php#L112-L121
 
 Of the compared approaches, Slim is most different from the proposed factories.
-Most notably, the `Request` implementation contains requirements that specific
+Most notably, the `Request` implementation contains requirements specific
```
Jean85: Maybe it was `that are`?

acinader: Maybe, but my $.02 is that nothing is lost but filler words by not adding `that are`.  As proposed the sentence is shorter without any loss of meaning.

PR #964, opened 2017-11-28
Added paragraph about the discussion of throwing exceptions for [45]xx or not.

Nyholm:
I think this short paragraph summarise the discussion.

### DIFF
--- a/proposed/http-client/http-client-meta.md
+++ b/proposed/http-client/http-client-meta.md
@@ -73,7 +73,17 @@ catch the specific exceptions and handle them accordingly.
 
 One could be more granular when defining exception. For example, `TimeOutException` and `HostNotFoundException`
 could be subtypes of `NetworkException`. The chosen approach is not to define such subtypes because
-the exception handling in a consuming library would not be different between those exceptions. 
+the exception handling in a consuming library would not be different between those exceptions.
+
+#### Throwing exceptions for 4xx and 5xx responses
+
+The initial idea was to allow the client to be configured to throw exceptions for responses
+with HTTP status 4xx and 5xx. That approach is not desired because consuming libraries would 
+have to check for 4xx and 5xx responses twice. First by simply verifying the status code and 
+then to catch possible exceptions. 
+
+To make the specification more predictable, it was decided that HTTP clients never will throw
+exceptions for 4xx and 5xx responses.   
 
 ### Background
 


proposed/http-client/http-client-meta.md
```diff
@@ -73,7 +73,17 @@ catch the specific exceptions and handle them accordingly.
...
+
+#### Throwing exceptions for 4xx and 5xx responses
+
+The initial idea was to allow the client to be configured to throw exceptions for responses
+with HTTP status 4xx and 5xx. That approach is not desired because consuming libraries would 
+have to check for 4xx and 5xx responses twice. First by simply verifying the status code and 
+then to catch possible exceptions. 
+
+To make the specification more predictable, it was decided that HTTP clients never will throw
+exceptions for 4xx and 5xx responses.
```
dbu: i think we should mention that applications may set up their client in a way that transforms those stati into exception if they so desire, but a client configured in that way MUST NOT be passed to a library?

kelunik: No, I'd not mention that here.

dbu: i just feel there is some implicit assumption in that topic, as its not conceptually wrong to do that in an application, and can be convenient when building an application. do we mention the topic of throwing exceptions for http status for an application anywhere?

Nyholm: I agree with @dbu, that is an implicit assumption. But the as @kelunik says, it should not be mentioned as it can never be a part of the specification that you are allowed to break it =)

kelunik: I'd probably not do it in an application myself. It's better to have something like that in an application:

```php
<?php

final class AppHttpClient implements Psr\HttpClient {
    private $httpClient;

    public function __construct(Psr\HttpClient $httpClient) {
        $this->httpClient = $httpClient;
    }

    public function request(Request $request): Response {
        return $this->httpClient->request($request);
    }

    public function requestOrThrow(Request $request): Response {
        $response = $this->httpClient->request($request);

        if (...) {
            throw ...;
        }

        return $response;
    }
}
```

dbu: so you would do your own decorator. makes sense as well.

dbu:
are you also happy to merge this as its now @kelunik ?

PR #965, opened 2017-12-01
spelling and other cosmetic changes for the HTTP client PSR

dbu:


proposed/http-client/http-client.md
```diff
@@ -12,39 +12,38 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
...
 
 ## Goal
 
 The goal of this PSR is to allow developers to create libraries decoupled from HTTP client
-implementations. This would make libraries more stable since the reduced number of
-dependencies and the likelihood to get in version conflicts is reduced.
-
-The second goal is that all HTTP clients should follow the [Liskov substitutions principle][Liskov].
-This means that all clients should act the same when sending a request. By default a HTTP client
-should not follow redirect nor throw exceptions on HTTP responses with status 4xx or 5xx.
```
dbu: this is stated on the interface with more clarity. rather than repeat the whole thing here, i think we can simply drop this here.

proposed/http-client/http-client.md
```diff
@@ -62,8 +61,8 @@ interface ClientInterface
      * Sends a PSR-7 request and returns a PSR-7 response. 
      * 
      * Every technically correct HTTP response MUST be returned as is, even if it represents a HTTP 
-     * error response or a redirect instruction. The only exception is 1xx responses which MUST be 
-     * assembled in the HTTP client. 
+     * error response or a redirect instruction. The only exception are 1xx responses which MUST be
```
Nyholm: Hm, Im not sure you are correct here. The verb ("be") i referring to "the only exception" which is singular.

Please, any native speaker, could you correct this?

ashnazg: `The only exceptions are 1xx responses, which MUST be `

michaelcullum: `exceptions are` or `exception is`. You could use either of those two this case.

proposed/http-client/http-client.md
```diff
@@ -62,8 +61,8 @@ interface ClientInterface
      * Sends a PSR-7 request and returns a PSR-7 response. 
      * 
      * Every technically correct HTTP response MUST be returned as is, even if it represents a HTTP 
-     * error response or a redirect instruction. The only exception is 1xx responses which MUST be 
-     * assembled in the HTTP client. 
+     * error response or a redirect instruction. The only special case is 1xx responses, which MUST
+     * be assembled in the HTTP client.
```
dbu: i actually don't like the word exception here, as its not about throwing an Exception... (thinking about non-native speakers who might get confused by the wording). does it make sense like this? both grammatically and semantically?

dbu:
Do PSR have a rule on line length? i notice that its currently a bit arbitrary, should we break after 80 or 100 characters? i guess we should do that completely separately from other editing, as it will massively mess up the diff.

PR #968, opened 2017-12-03
Added declare strict

Nyholm:
Does it makes sense having `declare(strict_types=1);` on interfaces?


kelunik:
No, it doesn't make sense, as it only affects the caller and there are no calls in these files.

Nyholm:
Thank you

PR #971, opened 2017-12-05
Extend Exception, not Throwable

weierophinney:
Per the [PHP Manual](http://php.net/Throwable), PHP classes cannot extend `Throwable`, but must instead extend `Exception`.

### DIFF
--- a/proposed/http-client/http-client.md
+++ b/proposed/http-client/http-client.md
@@ -90,7 +90,7 @@ namespace Psr\Http\Client;
 /**
  * Every HTTP client related Exception MUST implement this interface.
  */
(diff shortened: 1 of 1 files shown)


riveraj:
An interface can only extend another interface though, and `Throwable` is an interface; `Exception` is not.

Nyholm:
Thank you for the PR.

Yes, two interfaces `extends` each other. One interface cannot `implement` the other. Nor can an interface `extend` a class. 

Closing this as it gives syntax error. Also, have a look at https://groups.google.com/forum/?fromgroups#!topic/php-fig/7PFR-KAAcMk

PR #977, opened 2017-12-06
PSR-18 is HTTP Client, not HTTP Factories

riveraj:
I was confused as to why there are two HTTP Factories PSRs. This will not stand! 😄

### DIFF
--- a/index.md
+++ b/index.md
@@ -31,7 +31,7 @@ As also described in the [PSR Workflow Bylaw][workflow]. The Editor, or editors,
 
 | Num | Title                                | Editor(s)                      |
 |:---:|--------------------------------------|--------------------------------|
-| 18  | [HTTP Client][psr18]              | Tobias Nylhom                  |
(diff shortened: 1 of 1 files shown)


index.md
```diff
@@ -31,7 +31,7 @@ As also described in the [PSR Workflow Bylaw][workflow]. The Editor, or editors,
 
 | Num | Title                                | Editor(s)                      |
 |:---:|--------------------------------------|--------------------------------|
-| 18  | [HTTP Factories][psr18]              | Tobias Nylhom                  |
+| 18  | [HTTP Client][psr18]              | Tobias Nylhom                  |
```
dbu: should the table be aligned?

riveraj: Oops, thanks for the catch! Fixed!

dbu:
@michaelcullum the pullapprove thingy seems to not work out. can you somehow restart it? or is that a manual approval process?

Jean85:
Nope it's correct: it's a general file, so it requires an approval from one of us secretaries.

PR #979, opened 2017-12-08
PSR-18: Change introduction to be more specific to clients

kelunik:
Servers also send HTTP messages. This PSR is about sending requests and receiving responses.

### DIFF
--- a/proposed/http-client/http-client.md
+++ b/proposed/http-client/http-client.md
@@ -1,7 +1,7 @@
 HTTP Client
 ===========
 
-This document describes common interfaces for sending HTTP messages.
+This document describes a common interface for requesting HTTP resources.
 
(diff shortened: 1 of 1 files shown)


Nyholm:
Thank you

sagikazarmark:
I'm not quite sure I completely agree with this change. I agree with the intent, but I think something like `sending HTTP requests and receiving HTTP responses` describes better what an HTTP client does.

dbu:
depends on whether you look from the implementor point of view, or from the using application pov.

for the implementor, its about sending requests and receiving responses. for the user, its requesting resources. (hm, except for POST/PUT) when the main thing is triggering something on the other end...)

i think i like your proposal even better, mark. can you please do a new PR to adjust this again, so we can discuss that?

sagikazarmark:
Sure #986

PR #980, opened 2017-12-08
PSR-18: Move goal above the specification

kelunik:
I think that's the better order.

### DIFF
--- a/proposed/http-client/http-client.md
+++ b/proposed/http-client/http-client.md
@@ -7,6 +7,16 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
 "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
 interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
 
+## Goal
+
+The goal of this PSR is to allow developers to create libraries decoupled from HTTP client
+implementations. This will make libraries more reusable as it reduces the number of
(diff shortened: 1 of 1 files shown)


kelunik:
No idea, TBH. Just had a look at some PSRs, the all didn't contain a goal.

proposed/http-client/http-client.md
```diff
@@ -7,6 +7,16 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
 
+## Goal
+
+The goal of this PSR is to allow developers to create libraries decoupled from HTTP client
+implementations. This will make libraries more reusable as it reduces the number of
+dependencies and lowers the likelihood to get version conflicts.
+
+The second goal is that HTTP clients can be replaced as per the
+[Liskov substitutions principle][Liskov]. This means that all clients MUST behave in the
+same when sending a request.
```
kelunik: Actually, I think this paragraph can be dropped. It's a general principle and not specific to this PSR.

dbu: but for this PSR its particularly relevant. its the reason why we want to define eg 3xx behaviour and not leave defaults to the choice of the implementations.

kelunik: Yes, but e.g. the cache PSR also doesn't mention it. You could argue the same with cache misses there.

stefanotorresi:
What is listed here as the main goal, is usually stated in the very first section of the spec, after the title and before the RFC2119 reference. Other than that, goals and non-goals are usually listed more extensively in the meta document.

PR #981, opened 2017-12-09
grammar edits

ashnazg:
add commas after introductory clauses;
adjust some words to avoid repeated usage in the same sentence;
"forward" is adjective, whereas "forwards" is adverb;


weierophinney:
Thanks, @ashnazg !

PR #984, opened 2017-12-10
Clarify blank line after `<?php`

PowerKiKi:
Blank line after `<?php` opening should be standardized clearly and be code samples should abide to the rules.

Now all header blocks are treated equally with a unique and simple rule for all of them.

See discussion: https://groups.google.com/forum/#!topic/php-fig/gKb63vV7k_4

### DIFF
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -38,6 +38,7 @@ This example encompasses some of the rules below as a quick overview:
 
 ~~~php
 <?php
+
 declare(strict_types=1);
 
 namespace Vendor\Package;
@@ -121,6 +122,7 @@ each of the blocks below MUST be separated by a single blank line, and MUST NOT 
 a blank line. Each block MUST be in the order listed below, although blocks that are
 not relevant may be omitted.
 
+* Opening `<?php` tag.
 * File-level docblock.
 * One or more declare statements.
 * The namespace declaration of the file.
@@ -255,6 +257,7 @@ line.
 
 ~~~php
 <?php
+
 namespace Vendor\Package;
 
 use FooClass;
@@ -273,6 +276,7 @@ list MUST be on the next line, and there MUST be only one interface per line.
 
 ~~~php
 <?php
+
 namespace Vendor\Package;
 
 use FooClass;
@@ -295,6 +299,7 @@ declared on the next line after the opening brace.
 
 ~~~php
 <?php
+
 namespace Vendor\Package;
 
 use Vendor\Package\FirstTrait;
@@ -310,6 +315,7 @@ one-per-line, and each inclusion MUST have its own `use` import statement.
 
 ~~~php
 <?php
+
 namespace Vendor\Package;
 
 use Vendor\Package\FirstTrait;
@@ -329,6 +335,7 @@ closing brace MUST be on the next line after the `use` import statement.
 
 ~~~php
 <?php
+
 namespace Vendor\Package;
 
 use Vendor\Package\FirstTrait;
@@ -343,6 +350,7 @@ Otherwise it MUST have a blank line after the `use` import statement.
 
 ~~~php
 <?php
+
 namespace Vendor\Package;
 
 use Vendor\Package\FirstTrait;
@@ -391,6 +399,7 @@ A property declaration looks like the following.
 
 ~~~php
 <?php
+
 namespace Vendor\Package;
 
 class ClassName
@@ -417,6 +426,7 @@ parentheses, commas, spaces, and braces:
 
 ~~~php
 <?php
+
 namespace Vendor\Package;
 
 class ClassName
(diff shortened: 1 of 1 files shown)


proposed/extended-coding-style-guide.md
```diff
@@ -38,6 +38,7 @@ This example encompasses some of the rules below as a quick overview:
 
 ~~~php
 <?php
+
```
samdark: Declare usually goes w/o blank line as was noted in discussion at mailing list.

PowerKiKi: In the discussion Woody mentions what is commonly found in existing code, not what we should aim for.

Why should we make an exceptional case for `declare` ? How is it simpler to say, sometimes we must write:

```php
<?php
declare(strict_types=1);

use namespace Foo;

$a = 123;
```

But then sometime we must write:

```php
<?php

/**
 * Some comment
 */

declare(strict_types=1);

use namespace Foo;

$a = 123;
```

Allowing no blank line before `declare` in **some** cases, make the rules harder to remember, harder to use, and look inconsistent across files of the same project.

Crell: Making `declare` special makes little to no sense to me.

I MIGHT be convinced that "the first thing after the php tag" could omit the newline" is a worthwhile exception, but only if that's the case regardless of what that thing is.  Making an exception only sometimes just makes it harder for the human parser and machine parser.

drupol:
I also agree with @Crell here.

proposed/extended-coding-style-guide.md
drupol: I also agree with @Crell here.

gmponos:
I am really confused.

Quoted from the mailing list: 
> I just found that whether there should be one blank line  after opening `<?php` is not clarified now

> I agree that it has to be standardized and not left open to interpretation.


Quoted from the comments here: https://github.com/php-fig/fig-standards/pull/861#issuecomment-280863937

> It does.

>> What does?

>>> [PSR-12] should strictly declare if the <?php tag MUST or MUST NOT be followed by a blank line.

PowerKiKi:
@gmponos what is your question ?

gmponos:
Aren't those two statements above contradicting? I think both of them were stated from members of this repo, aren't they?

Look, I am not a native English speaker neither I use English all the time. So when I read something and a member comes and says "It does" and closes the issue I just rub my eyes and try to see where. 

Another member could have said something like "Look you are wrong, it does not but never the less we want to leave it like this" and since no-one else said anything about it I just thought that probably I am lost in translation and let's back off a little. 

Now someone comes and says "it does not". It's really discouraging contributions this way.

Just stating my mind here.

PowerKiKi:
I don't about the discussion you had on that other thread, maybe best to ask over there for clarification.

But one thing is now sure. Because this PR was merged, a blank line MUST follow the `<?php` opening tag.

proposed/extended-coding-style-guide.md
donquixote: Maybe people used to regard this line break as a break before the namespace (or any other code), not after the `<?php`. Or as a separation between regular PHP code and "redundant header stuff".

Since `<?php` and `declare(strict_types=1)` will usually be the same in all files across a package, people have seen it as part of the "opening shebang", which needs to be separated from the rest.

Not having a line break before `declare()` makes it easier to visually skip this part and focus on the things that are specific to this file.

PowerKiKi: @donquixote, the "opening shebang", as you call it, would most likely be the PHP opening tag immediately followed by a file-level docblock in many open source projects, because it is common practice to include some boilerplate about license. A `declare` statement would only come after those.

If you are willing to add boilerplate headers of several lines (which again seems to be a common practice and brings absolutely no value in terms of pure coding), and somehow be able mentally "skip" those, you might as well tolerate a single more blank line.

donquixote:
On 26 December 2017 at 06:36, Adrien Crivelli <notifications@github.com>
wrote:

> *@PowerKiKi* commented on this pull request.
> ------------------------------
>
> In proposed/extended-coding-style-guide.md
> <https://github.com/php-fig/fig-standards/pull/984#discussion_r158669018>:
>
> > @@ -38,6 +38,7 @@ This example encompasses some of the rules below as a quick overview:
>
>  ~~~php
>  <?php
> +
>
> @donquixote <https://github.com/donquixote>, the "opening shebang", as
> you call it, would most likely be the PHP opening tag immediately followed
> by a file-level docblock in many open source projects, because it is common
> practice to include some boilerplate about license. A declare statement
> would only come after those.
>
For my taste, the file docblock could come after the declare() part.
Has this order been defined elsewhere?

PowerKiKi:
The order has indeed [been defined](https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#3-declare-statements-namespace-and-import-statements), and is not what you seem to expect.

PR #987, opened 2017-12-11
Add clarifications and corrections

shadowhand:
Based on feedback from the mailing list review.

https://groups.google.com/d/msg/php-fig/l2hdE0DTgYw/oZWOwe-6BAAJ

### DIFF
--- a/proposed/http-handlers/request-handlers-meta.md
+++ b/proposed/http-handlers/request-handlers-meta.md
@@ -5,7 +5,8 @@ HTTP Server Request Handlers Meta Document
 
 The purpose of this PSR is to define formal interfaces for HTTP server request
 handlers ("request handlers") and HTTP server request middleware ("middleware")
-that are compatible with HTTP messages as defined in [PSR-7][psr7].
+that are compatible with HTTP messages as defined in [PSR-7][psr7] or subsequent
+replacement PSRs.
 
 _Note: All references to "request handlers" and "middleware" are specific to
 **server request** processing._
@@ -275,6 +276,10 @@ and a request handler and must return a response. The middleware may:
 - Create and return a response without passing the request to the request handler,
   thereby handling the request itself.
 
+It is expected that a middleware dispatching system will use the request handler
+to delegate response creation to the next available middleware. The last middleware
+would then act as a domain gateway to execute application code.
+
 #### Why doesn't middleware use `__invoke`?
 
 Doing so would conflict with existing middleware that implements the double-pass
@@ -305,8 +310,8 @@ To make it clear that the middleware can only be used in a synchronous, server
 side context.
 
 While not all middleware will need to use the additional methods defined by the
-server request interface, external requests are typically processed asynchronously
-and would usually return a [promise][promises] of a response. (This is primarily
+server request interface, outbound requests are typically processed asynchronously
+and would typically return a [promise][promises] of a response. (This is primarily
 due to the fact that multiple requests can be made in parallel and processed as
 they are returned.) It is outside the scope of this proposal to address the needs
 of asynchronous request/response life cycles.
--- a/proposed/http-handlers/request-handlers.md
+++ b/proposed/http-handlers/request-handlers.md
@@ -3,7 +3,8 @@ HTTP Server Request Handlers
 
 This document describes common interfaces for HTTP server request handlers
 ("request handlers") and HTTP server middleware components ("middleware")
-that use HTTP messages as described by [PSR-7][psr7].
+that use HTTP messages as described by [PSR-7][psr7] or subsequent
+replacement PSRs.
 
 HTTP request handlers are a fundamental part of any web application. Server side
 code receives a request message, processes it, and produces a response message.
@@ -58,10 +59,9 @@ Middleware using this standard MUST implement the following interface:
 ### 1.3 Generating Responses
 
 It is RECOMMENDED that any middleware or request handler that generates a response
-will use HTTP factories as defined in [PSR-17][psr17] in order to prevent dependence
-on a specific HTTP message implementation.
-
-[psr17]: https://github.com/php-fig/fig-standards/tree/master/proposed/http-factory
+will either compose a prototype of a PSR-7 `ResponseInterface` or a factory capable
+of generating a `ResponseInterface` instance in order to prevent dependence on a
+specific HTTP message implementation.
 
 ### 1.4 Handling Exceptions
 
@@ -83,17 +83,17 @@ use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 
 /**
- * An HTTP request handler process an HTTP request and produces an HTTP response.
- * This interface defines the methods required to use the request handler.
+ * Handles a server request and produces a response
+ *
+ * An HTTP request handler process an HTTP request in order to produce an
+ * HTTP response.
  */
 interface RequestHandlerInterface
 {
     /**
-     * Handle the request and return a response.
+     * Handles a request and produces a response
      *
-     * @param ServerRequestInterface $request
-     *
-     * @return ResponseInterface
+     * May call other collaborating code to generate the response.
      */
     public function handle(ServerRequestInterface $request): ResponseInterface;
 }
@@ -110,20 +110,20 @@ use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 
 /**
+ * Participant in processing a server request and response
+ *
  * An HTTP middleware component participates in processing an HTTP message,
- * either by acting on the request or the response. This interface defines the
- * methods required to use the middleware.
+ * either by acting on the request, generating the response, or forwarding the
+ * request to a subsequent middleware and possibly acting on its response.
  */
 interface MiddlewareInterface
 {
     /**
-     * Process an incoming server request and return a response, optionally delegating
-     * response creation to a handler.
+     * Process an incoming server request
      *
-     * @param ServerRequestInterface $request
-     * @param RequestHandlerInterface $handler
-     *
-     * @return ResponseInterface
+     * Processes an incoming server request in order to produce a response.
+     * If unable to produce the response itself, it may delegate to the provided
+     * request handler to do so.
      */
     public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface;
 }


shadowhand:
cc @Crell

proposed/http-handlers/request-handlers.md
```diff
@@ -83,17 +84,16 @@ use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 
 /**
+ * Handles a server request and produces a response
+ *
  * An HTTP request handler process a HTTP request and produces an HTTP response.
```
weierophinney: Two things on this line (yes, I realize you didn't change it!):

- need to consistently use the `an HTTP` phrasing, not `a HTTP`.
- I'd rephrase `and produces` to `in order to produce`.

proposed/http-handlers/request-handlers.md
```diff
@@ -110,20 +110,18 @@ use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 
 /**
+ * Participant in processing a server request and response
+ *
  * An HTTP middleware component participates in processing an HTTP message,
- * either by acting on the request or the response. This interface defines the
- * methods required to use the middleware.
+ * either by acting on the request or the response. The middleware may create a
```
weierophinney: "or the response" should become "or the response returned by the provided request handler".

proposed/http-handlers/request-handlers.md
```diff
@@ -110,20 +110,18 @@ use Psr\Http\Message\ResponseInterface;
...
-     * Process an incoming server request and return a response, optionally delegating
-     * response creation to a handler.
-     *
-     * @param ServerRequestInterface $request
-     * @param RequestHandlerInterface $handler
+     * Processes a request and the resulting response
      *
-     * @return ResponseInterface
+     * May delegate creating the response to the request handler.
      */
```
weierophinney: I suggest the following:

```php
/**
 * Process an incoming server request
 *
 * Processes an incoming server request in order to produce a response.
 * If unable to produce the response itself, it may delegate to the provided
 * request handler to do so.
 */
```

proposed/http-handlers/request-handlers.md
shadowhand: Would that be accurate though? Considering that the response *might* be generated by the middleware itself?

proposed/http-handlers/request-handlers.md
shadowhand: I don't think I agree with this wording. It implies that the primary purpose of the middleware is to produce a response, when that is actually the purpose of the request handler.

mnapoli: > It implies that the primary purpose of the middleware is to produce a response

It is. It's not incompatible with the fact that a request handler can also do that.

shadowhand: Fair enough.

proposed/http-handlers/request-handlers.md
```diff
@@ -83,17 +84,17 @@ use Psr\Http\Message\ResponseInterface;
...
+ * Handles a server request and produces a response
+ *
+ * An HTTP request handler process an HTTP request in order to produce an
+ * HTTP response.
  */
 interface RequestHandlerInterface
 {
     /**
-     * Handle the request and return a response.
+     * Handles a request and produces a response
```
Crell: s/Handle/Handles/ :+1: 
Removing the period, :-1:

proposed/http-handlers/request-handlers-meta.md
```diff
@@ -275,6 +276,10 @@ and a request handler and must return a response. The middleware may:
 - Create and return a response without passing the request to the request handler,
   thereby handling the request itself.
 
+It is expected that a middleware dispatching system will use the request handler
+to delegate response creation to the next available middleware. The last middleware
+would then act as a domain gateway to execute application code.
```
Crell: This still doesn't help me.  I need to see (probably in the meta doc) a code sample of how I'd put 2-3 middlewares and response handlers together.  Right now, the middlware signature wants a response handler for its 3rd parameter, and so I don't know how I'd string together a series of middlewares akin to what I'd do in Stack PHP et al.  I see a type mismatch and my brain segfaults. :smile:

proposed/http-handlers/request-handlers.md
Crell: Suggested: "either by acting on the request, generating the response, or forwarding the request to a subsequent middleware and possibly acting on its response."

mnapoli: I like your suggestion @Crell, it doesn't close any door and it's explicit.

Do we need to explicitly say that a middleware can also simply have side effects and not do anything else (well except forwarding the request)? (e.g. write to a log)

Or we could also simply say:

> An HTTP middleware takes an HTTP request and returns an HTTP response. The middleware can delegate generating the response to a request handler.

What a middleware does and how it works is an implementation detail (and the world of middlewares out there is really diverse, which is its strength). We don't want to accidentally narrow it.

> if conditions prevent the request from being processed further.

This is IMO a problem, some middlewares may return a response just because it does, e.g. *if* you write the router as a middleware (and assuming it acts as a dispatcher as well) it's fine that it returns a response. Same if you write controllers as middlewares, or if you integrate several applications together into a pipe or router (e.g. you put a Slim and a Zend Expressive app in a router or a pipe). Not the classic use cases but still they should be allowed since these extremes are what make the middlewares so powerful.

proposed/http-handlers/request-handlers.md
shadowhand: The [phpdoc standard disagrees](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#51-summary):

> A Summary MUST end with either
> 
> * a full stop (.) followed by a line break
> * or two sequential line breaks.

Only a single-line summary ends with a period. If a description is included, it does not.

proposed/http-handlers/request-handlers-meta.md
shadowhand: The short answer is: this is really hard to explain simply. This is a job better left to vendor implementations. Composing a middleware stack involves having a handler that points back to the stack, as well as a "fail condition" response.

https://github.com/northwoods/broker/blob/1.0.0/src/RequestHandler.php

proposed/http-handlers/request-handlers.md
shadowhand: This change has been made.

proposed/http-handlers/request-handlers-meta.md
Jean85: Since to pass the PSR we need 2 working implementation, maybe we can link those here in the meta?

weierophinney: @Crell — regarding this:

>  I need to see (probably in the meta doc) a code sample of how I'd put 2-3 middlewares and response handlers together. 

The problem is that _there's more than one way to do it_. In fact, there are a ton of ways to do it.

The most common way is to create a handler that acts as a queue:

```php
class Queue implements RequestHandlerInterface
{
    private $middleware = [];
    private $fallbackResponse;
    
    public function __construct(ResponseInterface $fallbackResponse)
    {
        $this->fallbackResponse = $fallbackResponse;
    }
    
    public function add(MiddlewareInterface $middleware)
    {
        $this->middleware[] = $middleware;
    }
    
    public function handle(ServerRequestInterface $request) : ResponseInterface
    {
        if (0 === count($this->middleware)) {
            return $this->fallbackResponse;
        }
        
        $middleware = array_shift($this->middleware);
        return $middleware->process($request, $this);
    }
}
```

An application bootstrap might then look like this:

```php
// Create instance
$app = new Queue(new Response());

// add one or more middleware:
$app->add(new CorsMiddleware());
$app->add(new FastRouteMiddleware());

// execute it:
$response = $app->handle(ServerRequestFactory::fromGlobals());
```

Each middleware then is passed the `$app` instance as the request handler, and it then determines if it has more middleware to execute, or if it should return its fallback response.

However, you could also write middleware that accepts other middleware via the constructor, and executes it directly:

```php
class Foo implements MiddlewareInterface
{
    private $app;
    
    public function __construct(MiddlewareInterface $app)
    {
        $this->app = $app;
    }
    
    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler) : ResponseInterface
    {
        $response = $this->app->process($request, $handler);
        // do something with the response...
        return $response;
    }
}
```

The above scenario is similar to how StackPHP (and its predecessor, Rack) operates. It requires that you create your "application" from the inside out, with the innermost middleware being passed to the next layer up, and so on and so forth. The innermost would then either use the request handler to create a response, or create one itself and return it. In your bootstrap, you then would execute the outermost layer:

```php
$requestHandler = new class implements RequestHandlerInterface
{
    public function handle(ServerRequestInterface $request) : ResponseInterface
    {
        return new Response();
    }
}

$response = $outermost->process(ServerRequestFactory::fromGlobals(), $requestHandler);
```

The above two scenarios are the most common. The first is the direction most are going due to the fact that it makes composition easier; you do not need to compose middleware to which you delegate directly within other middleware, just the dependencies on which it operates.

Are these the sort of examples you are looking for?

Crell: Yes!  My understanding (from previous drafts, most likely) was that you'd somehow pass the $next middleware directly to each, or you're passing along a stack of them, yet neither seemed to be the case.  I see now where that ended up: Externalizing the list of middlewares out of the spec entirely to the Queue class above.  Which... is going to be calling itself doubly-recursively, so it's not really a queue but still a stack.  The call stack in which every other level is the Queue object is not at all obvious and I'd not have stumbled across that on my own.  (It also makes me a little concerned about stack depth and backtrace depth, but that is perhaps because I am used to Drupal 8's super-deep backtraces.)

However, and this is why I wanted to drill down on this front, doesn't it mean a concrete class implemented for the StackPHP-style wouldn't work with the Queue-ish version?  It's $next is provided by the constructor, so it's not going to use what's passed to it in `process()`.  Are they really mix-and-matchable?

(This is the sort of "implementer guidance" I feel needs to be covered, either in the spec or the metadoc.)

proposed/http-handlers/request-handlers.md
```diff
@@ -110,20 +110,20 @@ use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 
 /**
+ * Participant in processing a server request and response
```
Crell: Period please. :smile:

proposed/http-handlers/request-handlers.md
```diff
@@ -83,17 +83,17 @@ use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 
 /**
- * An HTTP request handler process a HTTP request and produces an HTTP response.
- * This interface defines the methods required to use the request handler.
+ * Handles a server request and produces a response
```
Crell: Period please. :smile:

proposed/http-handlers/request-handlers.md
```diff
@@ -110,20 +110,20 @@ use Psr\Http\Message\ResponseInterface;
...
  */
 interface MiddlewareInterface
 {
     /**
-     * Process an incoming server request and return a response, optionally delegating
-     * response creation to a handler.
-     *
-     * @param ServerRequestInterface $request
-     * @param RequestHandlerInterface $handler
+     * Process an incoming server request
```
Crell: Period please. :smile:

proposed/http-handlers/request-handlers.md
shadowhand: As I stated in a previous comment (hidden because PR was updated):

The [phpdoc standard disagrees](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#51-summary):

> A Summary MUST end with either
> 
> * a full stop (.) followed by a line break
> * or two sequential line breaks.

Only a single-line summary ends with a period. If a description is included, it does not.

proposed/http-handlers/request-handlers-meta.md
weierophinney: > doesn't it mean a concrete class implemented for the StackPHP-style wouldn't work with the Queue-ish version? It's $next is provided by the constructor, so it's not going to use what's passed to it in process(). Are they really mix-and-matchable?

They'll still work. If you pass a stack-style middleware into a queue, then:

- you will need to pass some middleware implementation to its constructor
- any items added to the queue after it... will never be executed UNLESS the stack-style middleware calls on its request handler.
- if the stack-style middleware calls on the middleware implementation it composes, it will be passing it the request handler as well, and everything continues merrily on its way.

If you pass a queue-style middleware to a stack-style workflow, if it calls its request handler, then it will get a response back. Considering such middleware will not accept other middleware to the constructor, this means you would likely be using it as an inner-most layer.

So, interoperability is still available. However, stack-style middleware will have some limitations to how it composes non-stack-style middleware, due to the convention it uses for composing the next middleware. Which is why most middleware runners are going the route of using the request handler as a queue/stack/what-have-you for the purpose of composing layers, vs. the convention-based approach of nesting middleware within one another.

If you want, I can write up the above for the metadoc, if you feel it provides the implementer guidance you feel is missing.

proposed/http-handlers/request-handlers.md
```diff
@@ -83,17 +83,17 @@ use Psr\Http\Message\ResponseInterface;
...
+ * An HTTP request handler process an HTTP request in order to produce an
+ * HTTP response.
  */
 interface RequestHandlerInterface
 {
     /**
-     * Handle the request and return a response.
-     *
-     * @param ServerRequestInterface $request
+     * Handles a request and produces a response
```
mnapoli: > Handles a server request and produces a response

? (this would be completely consistent with the class docblock)

proposed/http-handlers/request-handlers.md
mnapoli: > Processes a server request and produces a response

? That makes it consistent with the docblock of the `HandlerInterface`, the only difference being handles/processes.

proposed/http-handlers/request-handlers.md
```diff
@@ -110,20 +110,20 @@ use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 
 /**
+ * Participant in processing a server request and response
+ *
  * An HTTP middleware component participates in processing an HTTP message,
- * either by acting on the request or the response. This interface defines the
- * methods required to use the middleware.
+ * either by acting on the request, generating the response, or forwarding the
+ * request to a subsequent middleware and possibly acting on its response.
```
mnapoli: s/either by/for example by/ ?

That avoids any potential misinterpretation as a limitation for implementations. For example some implementations would not act on the request or response at all, they would just do side-effects (e.g. a "logger middleware"). It *may* be covered by what's already described above but why take the risk of confusion.

The strict definition is the method signature (which can be summed up as `Processes a server request and produces a response`), the rest of the docblock is just to help understand, not limit the implementations.

shadowhand:
@weierophinney changes have been made here and rebased to get rid of conflicts. ✅ ?

PR #992, opened 2017-12-21
`declare` must follow the rule for assignment operator

PowerKiKi:
`declare` statements must follow the rule for assignment operator which dictates to have exactly 1 space around the operator.

There was no reason to make it an exception. It made the rule harder to explain, and harder to remember. By keeping a single rule for all cases everything is made simpler.

### DIFF
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -39,7 +39,7 @@ This example encompasses some of the rules below as a quick overview:
 ~~~php
 <?php
 
-declare(strict_types=1);
+declare(strict_types = 1);
 
 namespace Vendor\Package;
 
@@ -152,7 +152,7 @@ The following example illustrates a complete list of all blocks:
  * This file contains an example of coding styles.
  */
 
-declare(strict_types=1);
+declare(strict_types = 1);
 
 namespace Vendor\Package;
 
@@ -206,7 +206,7 @@ the strict types declaration and closing tag.
 
 For example:
 ~~~php
-<?php declare(strict_types=1) ?>
+<?php declare(strict_types = 1) ?>
 <html>
 <body>
     <?php
@@ -216,13 +216,13 @@ For example:
 </html>
 ~~~
 
-Declare statements MUST contain no spaces and MUST be exactly `declare(strict_types=1)`
+Declare statements MUST be exactly `declare(strict_types = 1)`
 (with an optional semi-colon terminator).
 
 Block declare statements are allowed and MUST be formatted as below. Note position of
 braces and spacing:
 ~~~php
-declare(ticks=1) {
+declare(ticks = 1) {
     //some code
 }
 ~~~
@@ -507,7 +507,7 @@ the two characters. The declaration keyword (e.g. string) MUST be lowercase.
 ~~~php
 <?php
 
-declare(strict_types=1);
+declare(strict_types = 1);
 
 namespace Vendor\Package;
 
@@ -534,7 +534,7 @@ and the type.
 ~~~php
 <?php
 
-declare(strict_types=1);
+declare(strict_types = 1);
 
 namespace Vendor\Package;
 


proposed/extended-coding-style-guide.md
```diff
@@ -216,13 +216,13 @@ For example:
 </html>
 ~~~
 
-Declare statements MUST contain no spaces and MUST be exactly `declare(strict_types=1)`
+Declare statements MUST be exactly `declare(strict_types = 1)`
```
Jean85: Shouldn't we make the semi-colon mandatory, so it follows the common rule about that too? At that point, this row can be changed into something like

> `declade` follows the same rules as the rest of the code: MUST have one space around the `=` operator and it MUST terminate with a semicolon

Jean85:
This also relates to #984

proposed/extended-coding-style-guide.md
PowerKiKi: I believe the mention of semicolon is only for cases like `<?php declare(strict_types = 1) ?>`. As shown in the example starting with "When wishing to declare...". Because otherwise it would be a parse error AFAIK. I don't think the case of semicolon before `?>` end tag is mentioned anywhere else. Maybe it should be removed from here and addressed separately ?

nicwortel:
Does this mean that PSR-12 is going to ignore the [~90% votes for not using spaces](https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide-meta.md#441-fig-representative-results)?

michaelcullum:
We already agreed not to do this and it's in review now so it cannot be modified without a push back to draft.

alcaeus:
With PSR-12 being moved back to DRAFT status (see https://groups.google.com/forum/#!topic/php-fig/1TBEvCIHFv0), can this be reopened and reconsidered, or should this be brought up in the mailing list first?

PowerKiKi:
I cannot re-open my PR, but I'll gladly submit a new one with the same patch if the PSR guys agree to reconsider the situation.

rentalhost:
@Jean85 so I guess that the "common usage" is counterintuitive here, once that it is definitively an [assignment](https://en.wikipedia.org/wiki/Assignment_(computer_science)). 

The RFC code is not a code style reference, it just should exemplify how a new feature should works. In the same page it put brackets on same line from class definition, so it even respect PSR.

I really don't see any reason to avoid spaces in `declare(X = 1)` but put spaces in `const X = 1`, for instance.

rob006:
@rentalhost `strict_types` it is neither a variable nor a constant, you can't use the result of this "assignment" and you can't "assign" anything to it. Compiler directive *is* a different thing than code, and it makes perfect sense to use different formatting rules for it. The same situation as with `echo` - you're not using rules for functions formatting (`echo('something')`) because it is not a function and different rules help to notice the difference.

rentalhost:
@rob006 Actually `echo` doesn't requires `()` as function does. The `()` here works like an [expression group](https://github.com/php/php-src/blob/524f5245c55223d36d1166554ca894fa77ea2dd1/Zend/zend_language_parser.y#L842), as in `(1 + 2)`. So in that case, the correct usage should be `echo ('something')` or even better, just `echo 'something'`, that is compatible with other language constructor like `return 5` (that must be writed as `return (5)` and not as `return(5)` once that it is not a function). That comes from [PSR-2 1.8](https://www.php-fig.org/psr/psr-2/#1-overview) (_"Control structure keywords MUST have one space after them; method and function calls MUST NOT"_) -- and I hope that _control structure_ here could includes too the `echo` (although that it is a _language constructor_). But I think that in that point we are according.

About `strict_types` not be neither a variable nor a constant, I should disagree with you. It acts like a "write-once" constants, but without the _read_ possibility after initial assignment. But the question here is not "what it does", but "what it seems like", and it seems like a constant assignment, where the constant name could be `ticks`, `encoding` or `strict_types` (it just doesn't respects PSR constant uppercase, but PHP itself doesn't care about that).

I just do not understand "why" everything should be spaced except that one. I don't think that "the common usage" should be a valid argument here, I think that it deserves a logic definition that everyone could understand the reason without come to here to understand "the why".

rob006:
> It acts like a "write-once" constants, but without the read possibility after initial assignment.

Sorry, but this is ridiculous. Then we can say that variable is a constant, but with possibility to overwrite after initialization. And `true` is also a variable, but read-only...

> But the question here is not "what it does", but "what it seems like"

I could not disagree more. This is only about "what it is". And your arguments like "it acts like constant" are very good reason to use different rules for `declare()` to distinguish the compiler directives from constant assignments.

rentalhost:
> Sorry, but this is ridiculous. Then we can say that variable is a constant, but with possibility to overwrite after initialization. And true is also a variable, but read-only...

No, it should sounds like: assignments could/must occur in constants, variables or declarations.

* Constants: assignment must occur once, and turn read-only after;
* Variables: assignment could occur anytime and are keeped as RW;
* Declarations: assignment must occur once (_similar to contants_), and could not be recovered anymore;

> I could not disagree more. This is only about "what it is". 

Okay, it could be about "what it is", and it is an assignment (I mean the internal part like `strict_types = 1`), once that you assign a value to an internal compiler property via userland code.

> And your arguments like "it acts like constant" are very good reason to use different rules for declare() to distinguish the compiler directives from constant assignments.

It doesn't make any sense. `declare()` could acts as a _language constructor_ or a _flow control constructor_ (when it is `declare(...) { ... }`). Then if we use like the second case, it should be `declare (...) { ... }` (like `for (...) { ... }`), but if first case should be `declare(...)`, but ops... It seems a function! We must distinguish it to not confuses the code developer to make sure that `declare<space?>()` is not a function.

PR #993, opened 2017-12-22
Removed statement on method of enforcement of line length limits

chrisemerson:
First of all, I know that it is specifically stated as a non-goal of this PSR to change anything in PSR-2. However, I think this statement is a mistake, and since PSR-2 is accepted and cannot be modified, this is the place to do it.

I and my colleagues don't think it should be the place of this PSR to dictate how the standards within it are enforced - only to define those standards. We believe that outputting warnings for items which aren't fully enforced is harmful, in that it creates noise amongst the real errors and warnings, and trains people to ignore warnings when they occur.

I don't know if this is the right way to approach this issue, or to start a discussion on it, but feel free to move or close this down if it's the wrong place.

### DIFF
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -90,8 +90,7 @@ The closing `?>` tag MUST be omitted from files containing only PHP.
 
 There MUST NOT be a hard limit on line length.
 
-The soft limit on line length MUST be 120 characters; automated style checkers
-MUST warn but MUST NOT error at the soft limit.
+The soft limit on line length MUST be 120 characters.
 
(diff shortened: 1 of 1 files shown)


PowerKiKi:
What about keeping the statement about errors? like so:

> The soft limit on line length MUST be 120 characters; automated style checkers MUST NOT error at the soft limit.

chrisemerson:
It would cause no problem for me personally, but I removed that as well to be consistent with the statement that the guide shouldn't be telling people how to achieve that style, only defining it.

dragonmantank:
+1 on this. We should define the limits, but not how individual tools might react. For personal projects I'd be less strict that work projects.

PR #996, opened 2018-01-04
[PSR-15] Provide middleware dispatch examples

weierophinney:
Per discussions on #987, on the mailing list, and in the PSR-15 working group, this patch builds on #987 to add a new section 6.3 to the meta document, detailing "Example Interface Interactions".

The section covers the following:

- A queue-based request handler that aggregates and dispatches middleware.
- A decorator-based request handler for building a layered dispatch mechanism.
- Two examples of re-usable middleware to demonstrate de-coupling from the application and the request handler.
  
This patch incorporates feedback made on #987, but not incorporated there.

### DIFF
--- a/proposed/http-handlers/request-handlers-meta.md
+++ b/proposed/http-handlers/request-handlers-meta.md
@@ -276,6 +276,13 @@ and a request handler and must return a response. The middleware may:
 - Create and return a response without passing the request to the request handler,
   thereby handling the request itself.
 
+When delegating from one middleware to another in a sequence, one approach for
+dispatching systems is to use an intermediary request handler composing the
+middleware sequence as a way to link middleware together. The final or innermost
+middleware will act as a gateway to application code and generate a response
+from its results; alternately, the middleware MAY delegate this responsibility
+to a dedicated request handler.
+
 #### Why doesn't middleware use `__invoke`?
 
 Doing so would conflict with existing middleware that implements the double-pass
--- a/proposed/http-handlers/request-handlers-meta.md
+++ b/proposed/http-handlers/request-handlers-meta.md
@@ -276,9 +276,11 @@ and a request handler and must return a response. The middleware may:
 - Create and return a response without passing the request to the request handler,
   thereby handling the request itself.
 
-It is expected that a middleware dispatching system will use the request handler
-to delegate response creation to the next available middleware. The last middleware
-would then act as a domain gateway to execute application code.
+When delegating from one middleware to the next in a sequence, a dispatching
+system should use an intermediary request handler as a way to link middleware
+together. The final or innermost middleware will act as a gateway to application
+code and generate a response from its results; alternately, the middleware may
+delegate this responsibility to a _request handler_.
 
 #### Why doesn't middleware use `__invoke`?
 
--- a/proposed/http-handlers/request-handlers-meta.md
+++ b/proposed/http-handlers/request-handlers-meta.md
@@ -429,7 +429,7 @@ $fallbackHandler = new NotFoundHandler();
 $app = new QueueResponseHandler($fallbackHandler);
 
 // Add one or more middleware:
-$app->add(new CorsMiddleware());
+$app->add(new AuthorizationMiddleware());
 $app->add(new RoutingMiddleware());
 
 // execute it:
@@ -494,7 +494,7 @@ $innerHandler = new class ($responsePrototype) implements RequestHandlerInterfac
 };
 
 $layer1 = new DecoratedRequestHandler(new RoutingMiddleware(), $innerHandler);
-$layer2 = new DecoratedRequestHandler(new CorsMiddleware(), $layer1);
+$layer2 = new DecoratedRequestHandler(new AuthorizationMiddleware(), $layer1);
 
 $response = $layer2->handle(ServerRequestFactory::fromGlobals());
 ```
@@ -526,42 +526,41 @@ Generally speaking, implementors of middleware should follow these guidelines:
   a new response by manipulating the one returned (e.g., `return
   $response->withHeader('X-Foo-Bar', 'baz')`).
 
-The `CorsMiddleware` is one that will exercise all three of these guidelines: if
-certain request conditions are not met, it needs to return an error response. In
-the case of a "pre-flight" request, it will return a response immediately with
-details for the client. In a case where the pre-flight request has already been
-made and expected headers are present, it delegates to the request handler, and
-then injects headers into the returned response.
+The `AuthorizationMiddleware` is one that will exercise all three of these guidelines:
+
+- If authorization is required, but the request is not authorized, it will use a
+  composed prototype response to produce an "unauthorized" response.
+- If authorization is not required, it will delegate the request to the handler
+  without changes.
+- If authorization is required and the request is authorized, it will delegate
+  the request to the handler, and sign the response returned based on the request.
 
 ```php
-class CorsMiddleware implements MiddlewareInterface
+class AuthorizationMiddleware implements MiddlewareInterface
 {
-    private $analyzer;
+    private $authorizationMap;
     private $responsePrototype;
 
-    public function __construct(CorsAnalyzer $analyzer, ResponseInterface $responsePrototype)
-    {
-        $this->analyzer = $analyzer;
+    public function __construct(
+        AuthorizationMap $authorizationMap,
+        ResponseInterface $responsePrototype
+    ) {
+        $this->authorizationMap = $authorizationMap;
         $this->responsePrototype = $responsePrototype;
     }
 
     public function process(ServerRequestInterface $request, RequestHandlerInterface $handler) : ResponseInterface
     {
-        $corsResult = $this->analyzer->analyze($request);
-
-        if ($corsResult->isInvalid()) {
-            return $corsResult->prepareForbiddenResponse($this->responsePrototype);
+        if (! $authorizationMap->requestRequiresAuthorization($request)) {
+            return $handler->handle($request);
         }
 
-        if ($corsResult->isPreFlight()) {
-            return $corsResult->preparePreFlightResponse($this->responsePrototype);
+        if (! $authorizationMap->isAuthorized($request)) {
+            return $authorizationMap->prepareUnauthorizedResponse($this->responsePrototype);
         }
 
         $response = $handler->handle($request);
-
-        return $corsResult->isCorsScoped()
-            ? $corsResult->prepareResponse($this->responsePrototype)
-            : $response;
+        return $authorizationMap->signResponse($response, $request);
     }
 }
 ```
--- a/proposed/http-handlers/request-handlers-meta.md
+++ b/proposed/http-handlers/request-handlers-meta.md
@@ -455,7 +455,7 @@ instance and a fallback request handler to pass to it. The application is built
 from the outside-in, passing each request handler "layer" to the next outer one.
 
 ```php
-class DecoratedRequestHandler implements RequestHandlerInterface
+class DecoratingRequestHandler implements RequestHandlerInterface
 {
     private $middleware;
     private $nextHandler;
@@ -489,8 +489,8 @@ $innerHandler = new class ($responsePrototype) implements RequestHandlerInterfac
     }
 };
 
-$layer1 = new DecoratedRequestHandler(new RoutingMiddleware(), $innerHandler);
-$layer2 = new DecoratedRequestHandler(new AuthorizationMiddleware(), $layer1);
+$layer1 = new DecoratingRequestHandler(new RoutingMiddleware(), $innerHandler);
+$layer2 = new DecoratingRequestHandler(new AuthorizationMiddleware(), $layer1);
 
 $response = $layer2->handle(ServerRequestFactory::fromGlobals());
 ```
--- a/proposed/http-handlers/request-handlers-meta.md
+++ b/proposed/http-handlers/request-handlers-meta.md
@@ -535,24 +535,20 @@ The `AuthorizationMiddleware` is one that will exercise all three of these guide
 class AuthorizationMiddleware implements MiddlewareInterface
 {
     private $authorizationMap;
-    private $responsePrototype;
 
-    public function __construct(
-        AuthorizationMap $authorizationMap,
-        ResponseInterface $responsePrototype
-    ) {
+    public function __construct(AuthorizationMap $authorizationMap)
+    {
         $this->authorizationMap = $authorizationMap;
-        $this->responsePrototype = $responsePrototype;
     }
 
     public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
     {
-        if (! $authorizationMap->requestRequiresAuthorization($request)) {
+        if (! $authorizationMap->needsAuthorization($request)) {
             return $handler->handle($request);
         }
 
         if (! $authorizationMap->isAuthorized($request)) {
-            return $authorizationMap->prepareUnauthorizedResponse($this->responsePrototype);
+            return $authorizationMap->prepareUnauthorizedResponse();
         }
 
         $response = $handler->handle($request);
@@ -561,12 +557,9 @@ class AuthorizationMiddleware implements MiddlewareInterface
 }
 ```
 
-Note the use of a "response prototype" in the above example; this approach
-allows the middleware to be agnostic of which PSR-7 implementation is in use,
-and yet still return a response in cases where it determines it should not
-delegate to the handler. Similarly, it is not concerned with how the request
-handler is implemented; it merely uses it to produce a response when
-pre-conditions have been met.
+Note that the middleware is not concerned with how the request handler is
+implemented; it merely uses it to produce a response when pre-conditions have
+been met.
 
 The `RoutingMiddleware` implementation described below follows a similar
 process: it analyzes the request to see if it matches known routes. In this
--- a/proposed/http-handlers/request-handlers-meta.md
+++ b/proposed/http-handlers/request-handlers-meta.md
@@ -396,11 +396,11 @@ request handler to the middleware.
 class QueueRequestHandler implements RequestHandlerInterface
 {
     private $middleware = [];
-    private $fallbackResponse;
+    private $fallbackHandler;
     
-    public function __construct(ResponseInterface $fallbackResponse)
+    public function __construct(RequestHandlerInterface $fallbackHandler)
     {
-        $this->fallbackResponse = $fallbackResponse;
+        $this->fallbackHandler = $fallbackHandler;
     }
     
     public function add(MiddlewareInterface $middleware)
@@ -412,7 +412,7 @@ class QueueRequestHandler implements RequestHandlerInterface
     {
         // Last middleware in the queue has called on the request handler.
         if (0 === count($this->middleware)) {
-            return $this->fallbackResponse;
+            return $this->fallbackHandler->handle($request);
         }
         
         $middleware = array_shift($this->middleware);
@@ -424,11 +424,11 @@ class QueueRequestHandler implements RequestHandlerInterface
 An application bootstrap might then look like this:
 
 ```php
-// Fallback response:
-$response = (new Response())->withStatus(500);
+// Fallback handler:
+$fallbackHandler = new NotFoundHandler();
 
 // Create request handler instance:
-$app = new QueueResponseHandler($response);
+$app = new QueueResponseHandler($fallbackHandler);
 
 // Add one or more middleware:
 $app->add(new CorsMiddleware());
@@ -438,10 +438,10 @@ $app->add(new RoutingMiddleware());
 $response = $app->handle(ServerRequestFactory::fromGlobals());
 ```
 
-This system has essentially two request handlers, one that will produce a
-response if no middleware generates one, and one for dispatching the middleware
-layers. (In this example, the `RoutingMiddleware` will likely execute composed
-handlers on a successful route match; see more on that below.)
+This system has two request handlers: one that will produce a response if the
+last middleware delegates to the request handler, and one for dispatching the
+middleware layers. (In this example, the `RoutingMiddleware` will likely execute
+composed handlers on a successful route match; see more on that below.)
 
 This approach has the following benefits:
 
@@ -450,6 +450,9 @@ This approach has the following benefits:
 - The `QueueRequestHandler` is agnostic of the PSR-7 implementation in use.
 - Middleware is executed in the order it is added to the application, making the
   code explicit.
+- Generation of the "fallback" response is delegated to the application
+  developer. This allows the developer to determine whether that should be a
+  "404 Not Found" condition, a default page, etc.
 
 #### Decoration-based request handler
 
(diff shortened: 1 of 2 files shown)


weierophinney:
Ping @Crell — I think the new section answers your questions, but would like for you to review and confirm.

Also pinging @shadowhand — do the changes look okay for you? Also, I rebased off of master, which picked up changes merged since you opened #987.

proposed/http-handlers/request-handlers-meta.md
```diff
@@ -363,6 +368,237 @@ executing that handler. If unable to route to a handler, it would instead
...
+This system has essentially two request handlers, one that will produce a
+response if no middleware generates one, and one for dispatching the middleware
+layers. (In this example, the `RoutingMiddleware` will likely execute composed
+handlers on a successful route match; see more on that below.)
+
+This approach has the following benefits:
+
+- Middleware does not need to know anything about any other middleware or how it
+  is composed in the application.
+- The `QueueRequestHandler` is agnostic of the PSR-7 implementation in use
```
Crell: Missing period on this line.

proposed/http-handlers/request-handlers-meta.md
```diff
@@ -363,6 +368,237 @@ executing that handler. If unable to route to a handler, it would instead
...
+$app->add(new RoutingMiddleware());
+
+// execute it:
+$response = $app->handle(ServerRequestFactory::fromGlobals());
+```
+
+This system has essentially two request handlers, one that will produce a
+response if no middleware generates one, and one for dispatching the middleware
+layers. (In this example, the `RoutingMiddleware` will likely execute composed
+handlers on a successful route match; see more on that below.)
```
Crell: I don't really see 2 handlers here... I see one handler ($app) that will end up invoked 2-3 times.  Am I missing something or is this worded awkwardly?

proposed/http-handlers/request-handlers-meta.md
```diff
@@ -363,6 +368,237 @@ executing that handler. If unable to route to a handler, it would instead
...
+        $this->nextHandler = $nextHandler;
+    }
+
+    public function handle(ServerRequestInterface $request) : ResponseInterface
+    {
+        return $this->middleware->process($request, $this->nextHandler);
+    }
+}
+
+$innerHandler = new class ($responsePrototype) implements RequestHandlerInterface {
```
Crell: Nitpick: $responsePrototype isn't defined yet.  Is this supposed to be a `new Response())->withStatus(500)`?

proposed/http-handlers/request-handlers-meta.md
```diff
@@ -363,6 +368,237 @@ executing that handler. If unable to route to a handler, it would instead
...
+            ? $corsResult->prepareResponse($this->responsePrototype)
+            : $response;
+    }
+}
+```
+
+Note the use of a "response prototype" in the above example; this approach
+allows the middleware to be agnostic of which PSR-7 implementation is in use,
+and yet still return a response in cases where it determines it should not
+delegate to the handler.Similarly, it is not concerned with how the request
```
Crell: Missing space on this line.

proposed/http-handlers/request-handlers-meta.md
```diff
@@ -363,6 +368,237 @@ executing that handler. If unable to route to a handler, it would instead
...
+- Producing a fallback response if no other layer does.
+- Dispatching middleware.
+
+#### Reusable Middleware Examples
+
+In the examples above, we have two middleware composed in each. In order for
+these to work in either situation, we need to write them such that they interact
+appropriately.
+
+Generally speaking, implementors of middleware should follow these guidelines:
```
Crell: I feel like these guidelines should be moved to the spec proper, in some form, as they directly inform how middlewares will interoperate.  Eg, when to call the next handler cannot be captured in an interface but is still relevant, and if done "wrong" will lead to a non-interoperable implementation.

proposed/http-handlers/request-handlers-meta.md
```diff
@@ -363,6 +368,237 @@ executing that handler. If unable to route to a handler, it would instead
...
+
+        if ($corsResult->isPreFlight()) {
+            return $corsResult->preparePreFlightResponse($this->responsePrototype);
+        }
+
+        $response = $handler->handle($request);
+
+        return $corsResult->isCorsScoped()
+            ? $corsResult->prepareResponse($this->responsePrototype)
+            : $response;
```
Crell: It's cool to see someone besides me that uses/likes that particular formatting idiom. :smile:

proposed/http-handlers/request-handlers-meta.md
```diff
@@ -275,6 +276,10 @@ and a request handler and must return a response. The middleware may:
 - Create and return a response without passing the request to the request handler,
   thereby handling the request itself.
 
+It is expected that a middleware dispatching system will use the request handler
+to delegate response creation to the next available middleware. The last middleware
+would then act as a domain gateway to execute application code.
```
Crell: This paragraph... doesn't quite make sense.  I can't wrap my head around what it's saying.  Let me try to rephrase and tell me if I'm in the right ballpark:

```
When delegating from one middleware to the next in a sequence, a dispatching system should use an intermediary request handler as a way to link two middlewares together.  The final or innermost middleware acts as a gateway to application code that will, in most cases, be responsible for response generation.
```

Is that accurate?  (Feel free to steal liberally if so.)  Should that be in the main spec in some form?

Crell:
This is a huge improvement, but still some bits aren't as direct as they could be.  See commentary above.

proposed/http-handlers/request-handlers.md
```diff
@@ -110,20 +110,20 @@ use Psr\Http\Message\ResponseInterface;
...
 
 /**
- * An HTTP middleware component participates in processing an HTTP message,
- * either by acting on the request or the response. This interface defines the
- * methods required to use the middleware.
+ * Participant in processing a server request and response
+ *
+ * An HTTP middleware component participates in processing an HTTP message;
+ * for example, by acting on the request, generating the response, or forwarding the
+ * request to a subsequent middleware and possibly acting on its response.
```
mnapoli: 👍

proposed/http-handlers/request-handlers-meta.md
```diff
@@ -363,6 +368,237 @@ executing that handler. If unable to route to a handler, it would instead
 execute the handler provided to the middleware. (This sort of mechanism can even
 be used in conjunction with middleware queues and stacks.)
 
+### 6.3 Example Interface Interactions
+
+The two interfaces, `RequestHandlerInterface` and `MiddlewareInterface`, were
+designed to work in conjunction with one another. While a
+`RequestHandlerInterface` can feasibly be used in isolation, it gains more power
```
shadowhand: This sentence isn't necessary, imo. RequestHandler is perfectly viable *without* middleware, not the other way around.

proposed/http-handlers/request-handlers-meta.md
```diff
@@ -363,6 +368,237 @@ executing that handler. If unable to route to a handler, it would instead
...
+- If pre-conditions are met, delegate creation of the response to the provided
+  request handler, optionally providing a "new" request by manipulating the
+  provided request (e.g., `$handler->handle($request->withAttribute('foo',
+  'bar')`).
+
+- Either pass the response returned by the request handler unaltered, or provide
+  a new response by manipulating the one returned (e.g., `return
+  $response->withHeader('X-Foo-Bar', 'baz')`).
+
+The `CorsMiddleware` is one that will exercise all three of these guidelines: if
```
shadowhand: I am mildly uncomfortable using a highly specific example like this. This assumes that I know what CORS is and what the failure conditions mean. If we're going to use a specific example perhaps something like routing would be more generally understood?

proposed/http-handlers/request-handlers-meta.md
```diff
@@ -363,6 +368,237 @@ executing that handler. If unable to route to a handler, it would instead
...
+```
+
+Note the use of a "response prototype" in the above example; this approach
+allows the middleware to be agnostic of which PSR-7 implementation is in use,
+and yet still return a response in cases where it determines it should not
+delegate to the handler.Similarly, it is not concerned with how the request
+handler is implemented; it merely uses it to produce a response when
+pre-conditions have been met.
+
+The `RoutingMiddleware` implementation described below follows a similar
```
shadowhand: Yes this is better, but I think the process should be:

- If routing matches, continue processing.
- Else return a `HTTP Not Found` response.

proposed/http-handlers/request-handlers.md
```diff
@@ -112,8 +112,8 @@ use Psr\Http\Message\ServerRequestInterface;
 /**
  * Participant in processing a server request and response
  *
- * An HTTP middleware component participates in processing an HTTP message,
- * either by acting on the request, generating the response, or forwarding the
+ * An HTTP middleware component participates in processing an HTTP message;
+ * for example, by acting on the request, generating the response, or forwarding the
```
shadowhand: Why the change of phrasing? "Either" seems like the appropriate wording.

proposed/http-handlers/request-handlers-meta.md
weierophinney: No, you're correct. The original example I created composed another request handler instead of a request prototype, and delegated to that handler if the last middleware in the queue calls on the handler instead of returning a response. I'm going to refactor the example to to do that, as it also then answers the "request prototype" question you have elsewhere.

proposed/http-handlers/request-handlers-meta.md
weierophinney: @shadowhand What do you think about moving these into the spec proper? If we do, where should they live?

proposed/http-handlers/request-handlers-meta.md
weierophinney: It could be, or a 404 status. I'll update to make this more clear.

proposed/http-handlers/request-handlers-meta.md
weierophinney: I'll see if I can come up with another example. This was one I liked because it hits all the various ways middleware might produce a response:

- on its own, due to a failed pre-condition
- by acting as a pass-through to the request handler
- by modifying the response returned by the request handler

Let me see if I can come up with an example, and I'll ping you if/when I do.

proposed/http-handlers/request-handlers-meta.md
weierophinney: That's actually exactly how we do it in Expressive. :+1:

The only issue I see with that is that it complicates the example: I then need to add dispatch middleware and a response prototype or factory. I felt the "map route to handler" approach was sufficiently simple and didn't require extra explanation.

proposed/http-handlers/request-handlers.md
weierophinney: This was based on feedback from @mnapoli, IIRC.

proposed/http-handlers/request-handlers-meta.md
weierophinney: Came up with an example using authorization instead of CORS; please review and let me know what you think.

proposed/http-handlers/request-handlers-meta.md
weierophinney: It's correct, except that the "application code that will, in most cases, be responsible for response generation" will likely either be the middleware itself, or the a _request handler_. I'll see if I can come up with a re-phrase that makes sense.

proposed/http-handlers/request-handlers-meta.md
```diff
@@ -276,9 +276,11 @@ and a request handler and must return a response. The middleware may:
...
   thereby handling the request itself.
 
-It is expected that a middleware dispatching system will use the request handler
-to delegate response creation to the next available middleware. The last middleware
-would then act as a domain gateway to execute application code.
+When delegating from one middleware to the next in a sequence, a dispatching
+system should use an intermediary request handler as a way to link middleware
+together. The final or innermost middleware will act as a gateway to application
+code and generate a response from its results; alternately, the middleware may
+delegate this responsibility to a _request handler_.
```
weierophinney: @Crell This is the updated text, using your suggestion as a starting point.

proposed/http-handlers/request-handlers-meta.md
```diff
@@ -368,6 +370,240 @@ executing that handler. If unable to route to a handler, it would instead
...
+
+- If authorization is required, but the request is not authorized, it will use a
+  composed prototype response to produce an "unauthorized" response.
+- If authorization is not required, it will delegate the request to the handler
+  without changes.
+- If authorization is required and the request is authorized, it will delegate
+  the request to the handler, and sign the response returned based on the request.
+
+```php
+class AuthorizationMiddleware implements MiddlewareInterface
```
weierophinney: @shadowhand Let me know if this altered example (replacing CorsMiddleware with AuthorizationMiddleware) makes more sense. I definitely want to demonstrate multiple middleware, and particularly different approaches middleware can take both to be re-usable as well as to change the workflow/paths of an application.

proposed/http-handlers/request-handlers-meta.md
```diff
@@ -368,6 +370,240 @@ executing that handler. If unable to route to a handler, it would instead
...
+        AuthorizationMap $authorizationMap,
+        ResponseInterface $responsePrototype
+    ) {
+        $this->authorizationMap = $authorizationMap;
+        $this->responsePrototype = $responsePrototype;
+    }
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler) : ResponseInterface
+    {
+        if (! $authorizationMap->requestRequiresAuthorization($request)) {
```
shadowhand: How about `doesRequireAuthorization` instead of `requestRequiresAuthorization`?

proposed/http-handlers/request-handlers-meta.md
```diff
@@ -368,6 +370,240 @@ executing that handler. If unable to route to a handler, it would instead
...
+
+    public function __construct(
+        AuthorizationMap $authorizationMap,
+        ResponseInterface $responsePrototype
+    ) {
+        $this->authorizationMap = $authorizationMap;
+        $this->responsePrototype = $responsePrototype;
+    }
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler) : ResponseInterface
```
shadowhand: `): ResponseInterface` as per purposed PSR-12 spec, no space.

https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#45-method-and-function-arguments

proposed/http-handlers/request-handlers-meta.md
```diff
@@ -368,6 +370,240 @@ executing that handler. If unable to route to a handler, it would instead
...
+{
+    private $authorizationMap;
+    private $responsePrototype;
+
+    public function __construct(
+        AuthorizationMap $authorizationMap,
+        ResponseInterface $responsePrototype
+    ) {
+        $this->authorizationMap = $authorizationMap;
+        $this->responsePrototype = $responsePrototype;
```
shadowhand: I'm not sure why we need a response prototype here. I think it would make more sense of the `AuthorizationMap` was responsible for generating/manipulating the response. That would simplify the example and keep the focus on the middleware.

proposed/http-handlers/request-handlers-meta.md
shadowhand: I don't think we should but putting code samples in the spec itself. Having the guidelines at the top of the spec would be good though.

proposed/http-handlers/request-handlers-meta.md
Jean85: Maybe `needsAuthorization` to keep it short? (for easiness of readability)

proposed/http-handlers/request-handlers-meta.md
Crell: Yes, I was referring to the guidelines, (probably written a bit differently).  I agree the code samples should remain in the metadoc.

proposed/http-handlers/request-handlers.md
Crell: This seems fine.  If we want to get super pedantic then "either" is technically only correct when there's 2 options, and in this case there are 3.  "For example" seems OK to me.

proposed/http-handlers/request-handlers-meta.md
Crell: Much clearer!  Why is the last "request handler" set off, though?  Perhaps "to a dedicated request handler"?  (Nitpicking here.)

A statement of this sort seems like it belongs in the spec.  It's not a "why and background", it's a "no, really, this is how you're supposed to use it".  That seems like it should be in the spec proper.

proposed/http-handlers/request-handlers.md
Crell: Or, technically, with the semicolon we could omit that word entirely and it would still be a grammatically valid list of the available options.

proposed/http-handlers/request-handlers-meta.md
```diff
@@ -368,6 +370,240 @@ executing that handler. If unable to route to a handler, it would instead
...
+```php
+class AuthorizationMiddleware implements MiddlewareInterface
+{
+    private $authorizationMap;
+    private $responsePrototype;
+
+    public function __construct(
+        AuthorizationMap $authorizationMap,
+        ResponseInterface $responsePrototype
+    ) {
```
Crell: There's only 2 parameters. No need to go with the harder-to-read multi-line approach.  Icky.

proposed/http-handlers/request-handlers-meta.md
mnapoli: I really think the spec should not dictate *how* middlewares should behave (or other implementation details), e.g. when to generate a response or call the delegate.

Maybe that's not what you meant, but the guidelines below are too precise IMO.

stefanotorresi: I agree with @shadowhand and @mnapoli: I don't think the spec document should contain these guidelines.

weierophinney: @mnapoli and @stefanotorresi — thanks; I think that's great feedback, and I tend to agree. The guidelines presented are implementation details and good practices, but _do not need to be followed in order to comply with the specification_. I'll leave them where they are.

proposed/http-handlers/request-handlers-meta.md
shadowhand: When using objects as type, I prefer the multi-line approach. Easier to read and less churn on diffs when parameters change.

proposed/http-handlers/request-handlers-meta.md
weierophinney: Agreed; making that change now.

proposed/http-handlers/request-handlers-meta.md
weierophinney: I was going to disagree, but when I re-formatted, I realized the line is still shorter than the `process()` method declaration line. About to commit that now.

weierophinney: Plus, with the change to remove the response prototype, it's now only one argument anyways...

weierophinney:
Requested changes have been made (mostly):

- I **did not** move the "guidelines" bullet points to the spec; as several folks noted, coding to those guidelines will definitely mean you're coding to the spec, but you do not need to follow them to implement the spec.
- I removed the emphasis around "request handler", and added the word "dedicated" before them, per @Crell.
- I moved that particular paragraph to section 1.2 of the spec, per @Crell.
- I updated the preamble to `MiddlewareInterface` to incorporate the various bits of feedback.
- I updated code samples to follow PSR-12.
- I made changes to the `AuthorizationMiddleware` example per feedback received.

@Crell, @shadowhand, @mnapoli and @stefanotorresi — can you give one more review so we can finalize these? I think we can then prepare for a CC vote.

proposed/http-handlers/request-handlers-meta.md
mnapoli: Could you make that part clear in the text? Because it reads like a specification (with the "should" keyword).

weierophinney: I'm not sure what other word to user here, TBH, and the fact that it's not ALL CAPS means it does not have the same semantic weight per RFC 2119. Maybe the following?

> Implementors of middleware striving for maximum interoperability may want to consider the following guidelines:

Would that work for you, @mnapoli ?

proposed/http-handlers/request-handlers.md
```diff
@@ -52,6 +52,12 @@ the creation of a resulting response, as defined by PSR-7.
 A middleware component MAY create and return a response without delegating to
 a request handler, if sufficient conditions are met.
 
+When delegating from one middleware to the next in a sequence, a dispatching
+system should use an intermediary request handler as a way to link middleware
+together. The final or innermost middleware will act as a gateway to application
+code and generate a response from its results; alternately, the middleware may
+delegate this responsibility to a dedicated request handler.
```
Crell: Nitpick: SHOULD and MAY should be capitalized to mean the defined SHOULD and MAY, I think, since this is in the spec proper.

weierophinney: My understanding is they should only be capitalized when they have the semantic meanings as defined in RFC 2119; the question is whether or not those words here have those semantic meanings. I did not originally write this with those meanings in mind.

The problem, of course, is that there are very few ways to re-word this to have the same intended meaning _without_ using those words. Can you propose an alternative? Or do you feel they have those semantics?

proposed/http-handlers/request-handlers-meta.md
```diff
@@ -368,6 +364,233 @@ executing that handler. If unable to route to a handler, it would instead
...
+  "404 Not Found" condition, a default page, etc.
+
+#### Decoration-based request handler
+
+In this approach, a request handler implementation decorates both a middleware
+instance and a fallback request handler to pass to it. The application is built
+from the outside-in, passing each request handler "layer" to the next outer one.
+
+```php
+class DecoratedRequestHandler implements RequestHandlerInterface
```
Crell: Nitpick: Technically, this isn't a decoratED handler.  It's a decoratING handler.  The name should be more descriptive.  Sorry for not catching that before.

proposed/http-handlers/request-handlers.md
ashnazg: If not RFC2119, maybe `can` could be used for both?

proposed/http-handlers/request-handlers-meta.md
weierophinney: Fixed.

proposed/http-handlers/request-handlers.md
weierophinney: `CAN` also has semantic meaning via RFC 2119. :)

Been discussing this and the idea of moving the section to the spec document itself heavily today; hopefully we can have resolution shortly.

ashnazg: Hmm, I don't see `CAN` mentioned in the RFC... hence me thinking it safe.

Would something wordy, like `could potentially`, do the job without sounding like a 2119 directive?

weierophinney:
At this time, based on discussion with the working group, I've moved the paragraph in question back to the meta document's section 6.2 preamble, with changes as discussed here and elsewhere to remove implications that these are anything more than optional recommendations.

proposed/http-handlers/request-handlers-meta.md
weierophinney: Updated!

PR #1003, opened 2018-01-23
 Minor sentence structure fix in PSR-15 document.

riveraj:
The line breaks makes the diff look awkward, but the only thing actually changed was adding 'make use of' before 'a factory [...]'.


Jean85:
Since this PSR has been approved, and this is not a simple misplaced comma, I would like to have the approval of either @weierophinney or @shadowhand before merging this as a [formatting fix](http://www.php-fig.org/bylaws/psr-amendments/#32-formatting--typos).

riveraj:
Upon reviewing the metadata document, I see I misunderstood the meaning of the sentence. While it is slightly awkward, it is nonetheless correct, so I am withdrawing the pull request.

PR #1007, opened 2018-02-06
Changed order of "use function" and "use const" declarations to reflect PHPStorm behaviour

kukulich:


michaelcullum:
This change is too large for a review phase, if you wish to discuss it as a change please bring it up on the ML but it's unlikely to change just because it differs from PHP storm.

Crell:
JetBrains has expressed in the past that they're happy to adopt whatever standard we define, as long as we actually give them a standard to follow.

kukulich:
Ok :)

artspb:
@Crell One can't say better.
@all PhpStorm 2018.1 will follow the new order.

rob006:
> but it's unlikely to change just because it differs from PHP storm.

And what is the reasoning behind current order? Class body usually follows `classes (traits), constants, properties, methods` order, why use a different order for imports?

PR #1008, opened 2018-02-13
chore(typo): delete reference to a wrong handler

ericlery:
QueueResponseHandler > QueueRequestHandler


michaelcullum:
This is a typo fix so can just be merged in by a @php-fig/secretaries

20uf:
Nice catch (-:

PR #1009, opened 2018-02-18
Change PSR12 sample code namespacing so that it parses

KorvinSzanto:
This is a small change but I think an important one. Previously attempting to evaluate sample code would give you something like
```
PHP Parse error:  syntax error, unexpected 'Namespace' (T_NAMESPACE), expecting identifier (T_STRING) or '{'
```

After this change you no longer get that parse error though you may still get a fatal error depending on the code.


samdark:
Good catch.

michaelcullum:
@dragonmantank It's up to you re: merging this...

dragonmantank:
Makes sense to me.

PR #1011, opened 2018-02-18
Update CONTRIBUTING.md

michaelcullum:


CONTRIBUTING.md
```diff
@@ -20,13 +20,12 @@ given according to the current bylaws of this group.
...
-All Editors, Coordinators and Sponsors of specifications in draft & review stage
+All Editors, Sponsors and WG members of specifications in draft & review stage
 have access to push to this (php-fig/fig-standards) repository; subject to
 secretary discretion
 
-All Editors, Coordinators and Sponsors of specifications have push access to utility
-and interface repositories and retrain this even after acceptance; subject to secretary
-discretion.
+All Editors and Sponsors of specifications have push access to utility and interface
+repositories and retrain this even after acceptance; subject to secretary discretion.
```
dbu: s/retrain/retain?

CONTRIBUTING.md
```diff
@@ -43,8 +42,8 @@ directly (excluding inital commits on new interface/util repositories) to master
 comment) or merged by the PSR Editor, except in the review phase when the coordinator
 should seek commment from the editor, but merging is at the coordinator's discretion
 * You must never merge a pull request that affects any file in the repository
-other than those you are on a working group for; you should request a secretary
-or member of that working group (mention @php-fig/psr-x) do so
+other than those you are on a working group for unless you are a secretary; you should
+request a secretary or member of that working group (mention @php-fig/psr-x) do so
```
dbu: before, it was implicitly obvious, now its more complicated. or is that a change of policy that secretaries are now allowed to merge things they have not been allowed to? maybe split it into "Secretaries are allowed to merge any pull requests" and "Editors and sponsors may only merge pull requests that exclusively affects files of their working group(s); for other pull requests, they should request..."

dbu (commented):
looks good to me. a typo note and one question/proposal to make it simpler to read.

CONTRIBUTING.md
Jean85: IMHO that specification is not needed; it's basically implicit because the secretaries have by definition admin rights on any PHP-FIG repo, so they act as gatekeepers. There are bylaws in place that already define what or what not a secretary may merge without second guessing... Basically just typos and the like, see: https://www.php-fig.org/bylaws/psr-amendments/#32-formatting--typos

CONTRIBUTING.md
```diff
@@ -66,7 +65,7 @@ access. Merge access is a privilege and not a right.
 # Tagging
 
 Tagging on utility and interface repository should be done regularly, ideally after
-every merge, or every batch of merges after PSR approval.
+every merge, or every batch of merges after PSR approval; and on PSR approval.
```
Jean85: I would reverse the two pieces, to make more sense (temporally):

> [..] ideally on PSR approval and after every subsequent merge (or batch of merges).

CONTRIBUTING.md
Jean85: Yep, that's a typo.

Jean85 (changes requested):
Overall LGTM, but we should refine a few edits.

Jean85:
@michaelcullum do you want to push this further? Do you want to open a discussion somewhere else about this?

Jean85:
I've applied all requested fixes as per review.

vdelau:
@Jean85, do we want to continue with this?

Jean85:
It would need a rework due to the newer bylaws updates. I'll try.

CONTRIBUTING.md
```diff
@@ -20,13 +20,12 @@ given according to the current bylaws of this group.
 
 # Merge & Access Policy
 
-All Editors, Coordinators and Sponsors of specifications in draft & review stage
+All Editors, Sponsors and WG members of specifications in draft & review stage
```
ashnazg, line 23: Is this the only piece that's not accurate based on the bylaw changes since the PR was submitted?  Do we just need to drop `WG members` here?

I don't immediately see any issues with the remaining changes... or does @Jean85 's last comment from Jan mean the whole doc itself needs reworked?

ashnazg: Is this the only piece that's not accurate based on the bylaw changes since the PR was submitted? Do we just need to drop `WG members` here?

I don't immediately see any issues with the remaining changes... or does @Jean85 's last comment from Jan mean the whole doc itself needs reworked?

gm122921980-create (commented):
accepted/PSR-2-coding-style-guide.md$ composer testuse NotificationChannels\Webhook\WebhookChannel;
use NotificationChannels\Webhook\WebhookMessage;
use Illuminate\Notifications\Notification;

class ProjectCreated extends Notification
{
    public function via($notifiable)
    {
        return [WebhookChannel::class];
    }

    public function toWebhook($notifiable)
    {
        return WebhookMessage::create()
            ->data([
               'payload' => [
                   'webhook' => 'data'
               ]
            ])
            ->userAgent("Custom-User-Agent")
            ->header('X-Custom', 'Custom-Header');
    }
}public function routeNotificationForWebhook()
{
    return 'http://requestb.in/1234x';
}$ composer test $ composer testLICENSE.mdpublic function toWebhook($notifiable)
{
    return WebhookMessage::create()
        ->data([
           'payload' => [
               'webhook' => 'data'
           ]
        ])
        ->userAgent("Custom-User-Agent")
        ->header('X-Custom', 'Custom-Header');
}

chairping:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

gm122921980-create (commented):
👍

PR #921, opened 2017-07-07
fix_wrong_word_spelling Update PSR-6-cache.md

phpseasa:
i think the psr-6 has a wrong word spelling ,  the word 'it' should be 'is' , pls check my pr and thx

PR #1015, opened 2018-03-22
Fix non-ASCII characters

szepeviktor:
From #921 

Was two `0xA0` characters, also causing indentation failure.


szepeviktor:
@Jean85 Some must have inserted that line without reformatting the paragraph.

PR #1019, opened 2018-04-03
consistency of order abstract final and static

tipounet:
Always having the same order of keywords makes reading easier and avoids changes along the way.

The order is that indicated for the variables, and exists in other languages.


Majkl578:
:-1: 

Should be:
* `abstract protected function`
* `final public static function`

Jean85:
As discussed previously with the rest of the WG, and following the feedback on @Majkl578 response, we are going to decline this, sorry.

PR #1022, opened 2018-04-16
PSR-17: Incorporate interface changes proposed within working group

weierophinney:
This patch provides changes based on discussion within the working group following a review I performed.

### Interface changes

First, it makes the following substance changes to the interfaces:

- All interface methods now use PHP 7 scalar type hints where appropriate, as well as return type hints.
- `ResponseFactoryInterface::createResponse()` adds an optional string argument, `$reasonPhrase`. This is done to mimic `ResponseInterface::withStatus()`, as the only way to set a reason phrase is to pass it with the associated status code to that method. If we want to provide a way to generate a complete initial response, we need to allow passing the reason phrase.
- `ServerRequestFactoryInterface` removes the `createServerRequestFromArray()` method, but adds an optional array `$serverParams` argument to `createServerRequest()`. The `ServerRequestInterface` does not have a mutator for the server params, so the only way to provide them is at initial creation. Additionally, the structure of server params varies based on SAPI, and may or may not contain the data necessary to generate a URI and/or HTTP method for the request. As such, as a working group, we feel it makes sense to have a single factory method that includes the three required data (HTTP method, URI, and server params) for a complete and valid server request.
- Any factory generating an instance that has collaborators has been updated to extend the factory interfaces for those collaborators. These include:
  - Each of `RequestFactory`, `ResponseFactory`, `ServerRequestFactoryInterface`, and `UploadedFileFactoryInterface` extends `StreamFactoryInterface`, as the factory may need to generate a stream and/or the consumer may need to do so in order to provide it to the factory.
  - Each of `RequestFactoryInterface` and `ServerRequestFactoryInterface` extend `UriFactoryInterface`, as the factory may need to generate a `UriInterface` instance and/or the consumer may need to do so in order to provide it to the factory.
  - `ServerRequestFactoryInterface` extends `UploadedFileFactoryInterface` as the factory may need to generate `UploadedFileInterface` instances and/or the consumer may need to do so in order to provide them to the factory.

The rationale for extending collaborator factories is as follows: these factories in essence already require the other factories in order to produce the dependencies. Having them explicitly extend those factories ensures the required methods are present, and those artifacts may be produced. Doing otherwise would be encouraging a lot of code duplication (and thus potential for bugs), and would also require that middleware and request handlers require multiple factories to produce individual instances.

### Content changes

- Updated the "People" section of the metadocument to detail the working group.
- Updated the "Votes" section to detail the WG formation, and add a placeholder for the review period.
- Added a "Design Decisions" section, with six initial items:
  - Why multiple interfaces? and why createServerRequestFromArray()?

### Editorial changes

On the editorial side of the spectrum, I've made the following changes:

- Use SSL-enable links where possible.
- Added more links.
- Removed a few code examples that were confusing. In particular, two of those that talked about stream creation had the line `fwrite($resource, $body)`, which made no sense given the context of each.
- Use consistent markup for bullet points (`-` instead of `*` everywhere).

### DIFF
--- a/proposed/http-factory/http-factory-meta.md
+++ b/proposed/http-factory/http-factory-meta.md
@@ -128,7 +128,7 @@ that create a resource from a file path.
 - [`Request`][slim-request] Requires `string $method`, `UriInterface $uri`,
   `HeadersInterface $headers`, `array $cookies`, `array $serverParams`, and
   `StreamInterface $body`. Contains a factory method `createFromEnvironment(Environment $environment)`
-  that is framework specific but analogous to the proposed `createServerRequestFromGlobals`.
+  that is framework specific but analogous to the proposed `createServerRequestFromArray`.
 - [`Response`][slim-response] No required parameters, status code defaults to `200`.
 - [`Stream`][slim-stream] Requires `resource $stream` for the body.
 - [`UploadedFile`][slim-uploaded-file] Requires `string $file` for the source file.
--- a/proposed/http-factory/http-factory-meta.md
+++ b/proposed/http-factory/http-factory-meta.md
@@ -122,7 +122,7 @@ that create a resource from a file path.
 
 #### 4.2.3 Slim
 
-Slim is a popular micro-framework that makes use of HTTP Messages from version
+[Slim][slim] is a popular micro-framework that makes use of HTTP Messages from version
 3.0 forward.
 
 - [`Request`][slim-request] Requires `string $method`, `UriInterface $uri`,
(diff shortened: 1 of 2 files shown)


proposed/http-factory/http-factory-meta.md
```diff
@@ -149,27 +161,65 @@ As there is no clear declaration in PSR-7 as to what values are explicitly
...
+(e.g., async application runners that run via the CLI environment), the
+`ServerRequestFactoryInterface::createServerRequest()` method exists, allowing
+you to create a minimal server request instance with the request method and URI.
 
+## 6. People
+
+This PSR was produced by a FIG Working Group with the following members:
+
+* Woody Gilk (editor), <woody.gilk@gmail.com>
+* Matthew Weier O'Phinney (sponsor), <mweierophinney@gmil.com>
```
webimpress: `s/gmil/gmail/` ?

weierophinney: Fixed - thanks!

weierophinney:
@shadowhand Ready for your review (as editor!).

PR #1023, opened 2018-04-19
[PSR-12] Add missing blank line between use blocks

teohhanhui:


KorvinSzanto:
Thanks @teohhanhui! That was a good one to fix!

PR #1027, opened 2018-04-26
Add Alessandro Lai to the PSR-12 WG

Jean85:
As discussed with others WG members, this is to officially add myself to the PSR-12 working group.


KorvinSzanto:
👍 👍

PR #1029, opened 2018-05-05
Index: Update numerical index to use status directly

Krinkle:
The indirection of these one-letter cryptic abbreviations does not
seem to add value to the reader. The cost of having a ever so slightly
wider column and having to type a word rather than a character seems
to outweigh the benefit of readability and reduced confusion.


Krinkle:
@stefanotorresi I've applied the changes.

PR #1030, opened 2018-05-09
Clarity on unary operator spacing in PSR-1 example

navnavnav:
Added comment to PSR-1 example making it clear that the white-space proceeding the unary operator is not part of the standard.

### DIFF
--- a/accepted/PSR-1-basic-coding-standard.md
+++ b/accepted/PSR-1-basic-coding-standard.md
@@ -90,6 +90,7 @@ function foo()
 
 // conditional declaration is *not* a side effect
 if (! function_exists('bar')) {
+    // note: the space proceeding the unary operator above is *not* part of the standard.
     function bar()
     {
         // function body


Jean85:
Hello, and thanks for the contribution!

This is a not trivial fix of the standard; I'm not sure we can merge this. Anyway, we are currently in the way of addressing this issue in the next, superseeding PSR, PSR-12.

navnavnav:
Hi @Jean85 

That's too bad. I've noticed some confusion from some developers, where they were interpreting the practice in that example as a standard.

As you've already mentioned, this issue does seem to be addressed in section 6 of PSR-12. I just worry others may stumble across this example and misinterpret it. Is there any other way we could make it more clear for them?

Jean85:
Sorry for the late response, I missed this one!

Nope, we cannot change PSR-2 for this non-trivial part. BTW PSR-12 will superseed that, so the first will be probably deprecated in favor of the latter, solving this issue.

Closing this, feel free to open a new PR targeting PSR-12 o start a discussion on the mailing list if you wish.

PR #1032, opened 2018-05-15
[PSR-17] Remove unnecessary returns from docblocks

shadowhand:
Strict typing already defines the return type.

Plus a couple of minor corrections.

### DIFF
--- a/proposed/http-factory/http-factory.md
+++ b/proposed/http-factory/http-factory.md
@@ -52,8 +52,6 @@ interface RequestFactoryInterface extends
      * @param UriInterface|string $uri The URI associated with the request. If
      *     the value is a string, the factory MUST create a UriInterface
      *     instance based on it.
-     *
-     * @return RequestInterface
      */
     public function createRequest(string $method, $uri): RequestInterface;
 }
@@ -78,8 +76,6 @@ interface ResponseFactoryInterface extends StreamFactoryInterface
      * @param string $reasonPhrase Reason phrase to associate with status code
      *     in generated response; if none is provided implementations MAY use
      *     the defaults as suggested in the HTTP specification.
-     *
-     * @return ResponseInterface
      */
     public function createResponse(int $code = 200, string $reasonPhrase = ''): ResponseInterface;
 }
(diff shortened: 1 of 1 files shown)


ashnazg:
Normally I agree with omitting tags when the code signature can provide the info...

However, in a *spec*, maybe the redundancy is good, in that it is some *reinforcing* documentation alongside the spec code.  If the tag and code are out of sync, it is a visible warning that the spec is somehow amiss and in need of deeper review.  Without it, an incorrect value might be on the code signature, and not be spotted as incorrect.

slavcodev:
> Having the @return in the docblock could lead to conflicts between the type defined by the code and the type defined in the docblock. There is no value to duplicating the return type.

Sounds incorrect. Which conflicts are you talking about? DocBlock describes interface, interface is explicit by the specific return type. No matter what implementation exactly the method returns.

shadowhand:
```
/**
  * @return FooInterface
  */
function getFoo(): BarInterface;
```

Which one is invalid? Return type should be defined in code, not docblock.

slavcodev:
DocBlock is not the code, it is docs source and should describes the code.

> Which one is invalid? Return type should be defined in code, not docblock.
Of course in this case the documentation is broken, this is mistake and it is responsibility of devs to keep it correct.

ashnazg:
Let's say the true, correct spec is this:
```
function getFoo(): BarInterface;
```

Let's say the spec in draft looks like this:
```
function getFoo(): FooInterface;
```

When reviewing the draft, that could easily be missed as incorrect.

But, if the spec insisted on using all tags to back up the code signatures too, then it's easy to spot the mismatch:
```
/**
 * @return BarInterface
 */
function getFoo(): FooInterface;
```

That's my only point.  In the case of a *spec*, perhaps the duplication can be a check-and-balance type of reinforcement-of-correctness, where the _narrative_ has to match the _signature_.

It's just a suggestion...

shadowhand:
Docblocks are used to generate documentation *when the code does not provide it*. With return typing, there is no need for `@return` documentation. Suggesting that docblocks somehow provide verification of intent is, at best, misguided.

PR #1037, opened 2018-05-23
Clarify Scope

ashnazg:
This limits PSR scope to specifically not make guesses at future doc patterns.


ashnazg:
Will split into two PSRs.

PR #1039, opened 2018-06-04
[PSR-17] Removed comment about UriInterface

Nyholm:
I removed a section because I think it is not needed. Of course it is good to be extra clear but IMHO this is too much. 

A valid request object will always have an `UriOjbect` and of course it should use the input parameter to create that `UriObject`...

--------

There are (at least) two ways to create a `RequestFactory`.

### Method 1

```php
class RequestFactory implements RequestFactoryInterface 
{
    public function createRequest(
        string $method,
        $uri,
        array $headers = [],
        $body = null,
        $protocolVersion = '1.1'
    ): RequestInterface {
        // We let the request implementation create a URI object. 
        return new Request($method, $uri, $headers, $body, $protocolVersion);
    }
}
```

### Method 2

```php
class RequestFactory implements RequestFactoryInterface 
{
    private $uriFactory;
    
    // .. Constructor

    public function createRequest(
        string $method,
        $uri,
        array $headers = [],
        $body = null,
        $protocolVersion = '1.1'
    ): RequestInterface {
        if (is_string($uri)) {
            $uri = $this->uriFactory->createUri($uri);
        }
        return new Request($method, $uri, $headers, $body, $protocolVersion);
    }
}
```

You may read the comment that I'm removing as it is trying to control if you implement a request factory with method 1 or 2. I do not think the interface should try to control this as it is an implementation detail.

### DIFF
--- a/proposed/http-factory/http-factory-meta.md
+++ b/proposed/http-factory/http-factory-meta.md
@@ -244,6 +244,30 @@ implementation-specfic.
 [swoole]: https://www.swoole.co.uk/
 [reactphp]: https://reactphp.org/
 
+### 5.6 Why does RequestFactoryInterface::createRequest allow a string URI?
+
+The primary use case of `RequestFactoryInterface` is to create a request, and
+the only required values for any request are the request method and a URI. While
+`RequestFactoryInterface::createRequest()` can accept a `UriInterface` instance,
+it also allows a string.
+
+The rationale is two-fold. First, the majority use case is to create a request
+instance; creation of the URI instance is secondary. Requiring a `UriInterface`
+means users would either need to also have access to a `UriFactoryInterface`, or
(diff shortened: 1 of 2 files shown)


Isinlor:
@frederikbosch 

Relevant discussion: https://github.com/http-interop/http-factory/pull/39/files#r184989265

Maybe it should be added to meta? `5. Design Decisions`

frederikbosch:
Thanks for the reference. I would add it there, yes. It generates questions why everything is type-hinted except the uri.

weierophinney:
@frederikbosch I've added section 5.6 to the meta document to clarify this decision, and pushed it to the branch @Nyholm has created; can you review, please?

frederikbosch:
@weierophinney Looks good!

PR #1040, opened 2018-06-08
Tiny clarification about a request factory

Isinlor:
We do not want to depend on a concrete implementation, but on the interface.

Prompted by Robert Lu second question: https://groups.google.com/forum/#!msg/php-fig/OpUnkrnFhe0/y2dT7CakAQAJ

### DIFF
--- a/proposed/http-factory/http-factory-meta.md
+++ b/proposed/http-factory/http-factory-meta.md
@@ -30,8 +30,8 @@ a specific vendor implementation in order to create these objects.
 Another pain point is when writing re-usable middleware or request handlers. In
 such cases, package authors may need to create and return a response. However,
 creating discrete instances then ties the package to a specific PSR-7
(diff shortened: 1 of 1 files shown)


proposed/http-factory/http-factory-meta.md
```diff
@@ -30,8 +30,8 @@ a specific vendor implementation in order to create these objects.
 Another pain point is when writing re-usable middleware or request handlers. In
 such cases, package authors may need to create and return a response. However,
 creating discrete instances then ties the package to a specific PSR-7
-implementation. If these packages rely on a request factory instead, they can
-remain agnostic of the PSR-7 implementation.
+implementation. If these packages rely on a request factory interface instead,
```
dbu: if we talk about the factory interface, i think it should be "rely on the request factory interface" (*the* instead of *a*). there is only one factory interface, which is the point. so the library can depend on the psr thing.

PR #1045, opened 2018-06-24
Update PSR-1-basic-coding-standard.md

voyula:


Jean85:
I'm sorry, but that is an approved PSR, it shouldn't be altered. Also, this modification is not justified.

PR #1046, opened 2018-06-25
Use symfony.com canonical URL for security

fabpot:
And remove link at the bottom which does not seem to be used anywhere in the document.

### DIFF
--- a/proposed/security-reporting-process-meta.md
+++ b/proposed/security-reporting-process-meta.md
@@ -45,7 +45,7 @@ for how security vulnerabilities go from discovery to fixing to public
 disclosure. Inspiration could be drawn from this list of security disclosure
 processes in various PHP and non-PHP projects:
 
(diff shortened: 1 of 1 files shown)


proposed/security-reporting-process-meta.md
```diff
@@ -45,7 +45,7 @@ for how security vulnerabilities go from discovery to fixing to public
 disclosure. Inspiration could be drawn from this list of security disclosure
 processes in various PHP and non-PHP projects:
 
-* http://symfony.com/doc/current/contributing/code/security.html
+* http://symfony.com/security
```
xabbuh: What about using HTTPS here instead?

Jean85: We should use HTTPS on all those links, so we should address it with a separate PR

fabpot: see #1048

PR #1058, opened 2018-07-31
Remove reference to interface inheritance

ajgarlag:
I think this statement should had been removed in cd3ffe3f3ea17c2f0874a3b184651f839edefaea.

The `UploadedFileFactoryInterface` does not extend `StreamFactoryInterface`.


Jean85:
Since it's a fix (the code obviously states something clear), we can merge it!

PR #1060, opened 2018-08-13
Provides edits for PSR-18

weierophinney:
This patch reviews primarily the language of the specification, updating it with the following:

- Consistent usage of RFC 2119 verbiage.
- Code marks around all words representing code, HTTP artifacts, etc.
- Minor rephrasing to remove ambiguity, passive phrasing, etc.

PR #1063, opened 2018-08-15
Removes `Exception` subnamespace

Arnch:
This patch removes `Exception` subnamespace (according to [#1060](https://github.com/php-fig/fig-standards/pull/1060))


Nyholm:
Thank you

PR #1064, opened 2018-08-15
Fix incorrect word usage

jsmitty12:
_parenthesis_ was incorrectly used (instead of _brace_) in the _Anonymous Classes_ section


samdark:
Valid bug. Thanks!

PR #1070, opened 2018-09-06
Provide guidance on when to use Messages vs Tasks

Crell:


proposed/event-dispatcher-meta.md
```diff
@@ -90,6 +90,21 @@ The Working Group therefore chose to encapsulate the registration of listeners b
...
+
+Despite their similarities, the Message and Task pipelines have two distinct workflows and use cases.
+
+| Messages                      | Tasks
+|-------------------------------|-------------------------------
+| Must be immutable             | May me mutable
+| Must be serializable          | May be serializable
+| May be delayed                | Must be processed immediately
+| Listener order not guaranteed | Listener order is guaranteed
+| All listeners will fire       | Listeners may short-circuit the pipeline
```
weierophinney: May short-circuit the pipeline **IF AND ONLY IF** the stoppable task interface was used for the task provided.

proposed/event-dispatcher-meta.md
```diff
@@ -44,9 +44,36 @@ On further review, it was determined that Collection was a special case of Objec
...
+
+As a general guideline, a Message Notifier is appropriate when:
+
+* The Emitter does not care what responses are taken to an event.
+* The common case would be that only a single listener is registered, that is, when the Message Notifier is being used as a Command Bus.
+* The Emitter anticipates listeners to be expensive and therefore likely to be deferred for later processing.
+
+A Task Processor is appropriate when:
+
+* The event is being used as a "hook" or "pointcut" to extent or modify the Emitter's behavior.
```
weaverryan: extend

proposed/event-dispatcher-meta.md
```diff
@@ -44,9 +44,36 @@ On further review, it was determined that Collection was a special case of Objec
...
+* The Emitter wishes to allow listeners to interact with each other (through mutating the event).
+* NEED A 3RD EXAMPLE HERE.
```
weaverryan: Maybe... those first two (especially the first) were great.

Crell: Sure, but IMO we should have 3 for each, to make it clear that they're bot legit. :smile:

proposed/event-dispatcher-meta.md
```diff
@@ -44,9 +44,36 @@ On further review, it was determined that Collection was a special case of Objec
...
+
+* The Emitter does not care what responses are taken to an event.
+* The common case would be that only a single listener is registered, that is, when the Message Notifier is being used as a Command Bus.
+* The Emitter anticipates listeners to be expensive and therefore likely to be deferred for later processing.
+
+A Task Processor is appropriate when:
+
+* The event is being used as a "hook" or "pointcut" to extend or modify the Emitter's behavior.
+* The Emitter wishes to allow listeners to interact with each other (through mutating the event).
+* NEED A 3RD EXAMPLE HERE.
```
weierophinney: How about: "The emitter chooses to delegate when processing is complete to the listeners."

proposed/event-dispatcher-meta.md
```diff
@@ -44,9 +44,36 @@ On further review, it was determined that Collection was a special case of Objec
...
 
-### 4.2 Immutable events
+### 4.2 Intended use cases
 
-Initially the Working Group wished to define all Events as immutable message objects, similar to PSR-7.  However, that proved problematic in the Processor case.  In both of those cases Listeners needed a way to return data to the caller.  In concept there were three possible avenues:
+Despite their similarities, the Message and Task pipelines have two distinct workflows and use cases.
+
+| Messages                      | Tasks
+|-------------------------------|-------------------------------
+| Must be immutable             | May me mutable
```
weierophinney: s/me/be/

proposed/event-dispatcher-meta.md
```diff
@@ -44,9 +44,36 @@ On further review, it was determined that Collection was a special case of Objec
...
+| Must be serializable          | May be serializable
+| May be delayed                | Must be processed immediately
```
weierophinney: I'm not entirely convinced that tasks must be processed immediately. As @WyriHaximus demonstrated last week, a `Task` could compose one or more promises, which the listeners either create or react to. As such, processing could be delayed in them as well.

proposed/event-dispatcher-meta.md
```diff
@@ -44,9 +44,36 @@ On further review, it was determined that Collection was a special case of Objec
...
+| Listener order not guaranteed | Listener order is guaranteed
```
weierophinney: Listener order is a product of the listener provider, not the notifier or the task processor; they simply iterate the listeners. As such, I'm not sure we should bring it up here.

proposed/event-dispatcher-meta.md
```diff
@@ -44,9 +44,36 @@ On further review, it was determined that Collection was a special case of Objec
...
+| All listeners will fire       | Listeners may short-circuit the pipeline (if the Stoppable interface is implemented)
+| One-way communication only    | May be one-way or two-way communication
+
+As a general guideline, a Message Notifier is appropriate when:
+
+* The Emitter does not care what responses are taken to an event.
```
weierophinney: I think we should rephrase this to:

> The Emitter does not care what actions listeners take in response to an event.

Unless this is implying the emitter does not check the event state on each iteration. In which case:

> The Emitter will not check event state after each listener.

proposed/event-dispatcher-meta.md
Crell: True; although in that case I would argue that the listener is processing the event by enqueuing the async action.

Perhaps "listeners must be called immediately" (even if the listener then does something async)?

proposed/event-dispatcher-meta.md
Crell: True; although in that case I would argue that the listener is processing the event by enqueuing the async action.

Perhaps "listeners must be called immediately" (even if the listener then does something async)?

proposed/event-dispatcher-meta.md
Crell: A notifier may do something like fork or use pthreads to handle different listeners, which means the order is non-deterministic.  Task order must be determinstic.

proposed/event-dispatcher-meta.md
Crell: It means both; it doesn't care and it won't check.  I've changed it to your first suggestion.

PR #1071, opened 2018-09-09
[PSR-7] Adding throw exeception on `getParsedBody()`

jjsaunier:
It's an open discussion about how rely on deserialization failure when body is malformed. 

Today `null` is returned if body is incorrect (due to signature method) so the error is muted and the description say: 

> A null value indicates absence of body content so it's not the purpose. 

For example with json content, when `json_decode` fail because of invalid content, we should explicitly throw an exception. 

From controller side, `null` look's like empty or missing body rather than invalid content and need more check to indicate the correct issue to the end user. As developer in first hand, you didn't think of that case, you need experience to know that. 

Other point, this encourages developer to mute possible exception from deserializer to match method signature. 

What do you think ?

### DIFF
--- a/accepted/PSR-7-http-message.md
+++ b/accepted/PSR-7-http-message.md
@@ -1082,6 +1082,7 @@ interface ServerRequestInterface extends RequestInterface
      *
      * @return null|array|object The deserialized body parameters, if any.
      *     These will typically be an array or object.
+     * @throws \InvalidArgumentException if body cannot be deserialize.                           
      */
     public function getParsedBody();
 


stefanotorresi:
> Today null is returned if body is incorrect

This is an incorrect assumption.
The docblock states this clearly, and you cited this yourself: "A null value indicates absence of body content".
I think it is a responsibility of the implementation to not allow creating an instance with an invalid body in the first place, so this method should not throw under any circumstance.
A common way to implement this is parsing the content when the body content itself is mutated via `Message::withBody()`, and in that case you can throw an `InvalidArgumentException` for invalid values, as per docblock.

weierophinney:
@ProPheT777 If you're in that situation, it means that whatever process that was responsible for parsing the body content chose to silence errors. This is bad behavior; any such code should raise an exception or return an error response in such situations, and not pass on handling of the request to lower layers.

ProPheT777:
The context  : https://github.com/slimphp/Slim/blob/1ca78596de8f1b0e2389b812a6ce64d1ccd9e49f/Slim/Http/Request.php#L1041

> This is an incorrect assumption.

This is not assumption, it's a fact in my context, but I didn't put it before, so it's my fault.

> The docblock states this clearly, and you cited this yourself: "A null value indicates absence of body content".
Decoding is performed in that method, and the signature do not allow exception, how to deal with it?

> docblock states this clearly, and you cited this yourself: "A null value indicates absence of body content".

Yep, i'm ok with that, but how to know if the body is invalid regarding the format until trying to decode it ?

> A common way to implement this is parsing the content when the body content itself is mutated via Message::withBody() 

I'm not in that case, request is coming from a client, without mutation.

 @weierophinney 

> it means that whatever process that was responsible for parsing the body content chose to silence errors. Yes we are agree on that. My point is, decoding is performed in `getParsedBody` and the signature do not throw exception. how deal with it ? 

If you have solution, without changing signature i'm fine. May be the initial implementation should be improved is some way.

weierophinney:
> My point is, decoding is performed in getParsedBody and the signature do not throw exception. how deal with it ?

You won't like it, but the solution is to not parse the body contents within `getParsedBody()`. Do it in middleware, and have the middleware inject the contents for the next layer to use. As an example, take a look at the [Expressive BodyParamsMiddleware](https://github.com/zendframework/zend-expressive-helpers/blob/c95761fe60cfa5e20cf3264c54ddc1e9f0fc6c7c/src/BodyParams/BodyParamsMiddleware.php). You add this middleware to your application, and have it parse the body content based on the content-type of the request. If the strategy selected cannot parse it, it raises an exception, and your error handling middleware catches it.

weierophinney:
@ProPheT777 I'm not sure if Slim supports it, but Expressive supports per-route pipelines. As such, we typically recommend piping that middleware only where it's needed. This solves the "don't parse it if you don't need it" problem.

PR #1072, opened 2018-09-11
PSR 18 wordsmithing

Crell:
As discussed in Slack, here's my offered rewording/formalizing/clarifying of PSR-18.  I am fairly confident that none of the changes here have any impact on the spec itself; that is, there's no API change here.  It's just moving things around and rewording to be more formal and precise.

I have other comments not related to just rewording in-place where things are not quite clear, but I'll save those for a second PR and/or discussion to make this simpler.

I've broken it up into a bunch of specific commits to make it easier to review.

Also, ugh, hard line breaks in specs makes this kind of editing so much harder. :cry:

### DIFF
--- a/proposed/http-client/http-client.md
+++ b/proposed/http-client/http-client.md
@@ -38,21 +38,24 @@ If a Client chooses to alter either the HTTP request or HTTP response, it MUST e
 response remains internally consistent.  For example, if a Client chooses to decompress the message
 body then it MUST also remove the `Content-Encoding` header and adjust the `Content-Length` header.
 
-### Exceptions
+A Client MUST:
 
-All exceptions thrown by the client MUST implement `Psr\Http\Client\ClientExceptionInterface`.
+* Reassemble a multi-step HTTP 1xx response itself so that what is returned to the Caller is an valid HTTP response
+of status code 200 or higher.
 
-When the HTTP client is called with a request that is invalid and cannot be sent, the client
-MUST throw a `Psr\Http\Client\RequestExceptionInterface`. If there is an error
-with the network or the remote server cannot be reached, the HTTP client MUST throw
-a `Psr\Http\Client\NetworkExceptionInterface`.
+## Error handling
 
-Smaller issues that do not block the client from sending the request (such as
-invalid HTTP versions) MUST NOT result in exceptions.
+A Client MUST NOT treat a well-formed HTTP request or HTTP response as an error condition.  For example, response
+status codes in the 400 and 500 range MUST NOT cause an exception and MUST be returned to the Caller as normal.
 
-If the remote server answers with a response that can be parsed into a PSR-7 response,
-the client MUST NOT throw an exception. For example, response status codes in the
-400 and 500 range MUST NOT cause an exception.
+A Client MUST throw an instance of `Psr\Http\Client\ClientExceptionInterface` if and only if it is unable to send
+the HTTP request at all or if no HTTP response is ever returned (such as a timeout).
+
+If a request cannot be sent because the request message is not a valid HTTP request, the Client MUST throw an instance
+of `Psr\Http\Client\RequestExceptionInterface`.
+
+If the request cannot be sent due to a network failure of any kind the Client MUST throw an instance of
+`Psr\Http\Client\NetworkExceptionInterface`.
 
 ## Interfaces
 
--- a/proposed/http-client/http-client.md
+++ b/proposed/http-client/http-client.md
@@ -80,9 +80,6 @@ interface ClientInterface
     /**
      * Sends a PSR-7 request and returns a PSR-7 response.
      *
-     * Every technically correct HTTP response MUST be returned as-is, even if it represents an HTTP
-     * error response or a redirect instruction.
-     *
      * @param RequestInterface $request
      *
      * @return ResponseInterface
--- a/proposed/http-client/http-client.md
+++ b/proposed/http-client/http-client.md
@@ -62,6 +62,9 @@ of `Psr\Http\Client\RequestExceptionInterface`.
 If the request cannot be sent due to a network failure of any kind the Client MUST throw an instance of
 `Psr\Http\Client\NetworkExceptionInterface`.
 
+Clients MAY throw more specific exceptions than those defined here (a `TimeOutException` or `HostNotFoundException` for
+example), provided they implement the appropriate interface defined above.
+
 ## Interfaces
 
 ### ClientInterface
--- a/proposed/http-client/http-client.md
+++ b/proposed/http-client/http-client.md
@@ -20,8 +20,9 @@ same way when sending a request.
 ## Definitions
 
 * **Client** - A Client is a library that implements this specification for the purposes of
-sending PSR-7-compatible HTTP Request messages and returning a PSR-7-compatible HTTP Response message to a Caller.
-* **Caller** - A Caller is any code that makes use of a Client.
+sending PSR-7-compatible HTTP Request messages and returning a PSR-7-compatible HTTP Response message to a Calling library.
+* **Calling Library** - A Calling Library is any code that makes use of a Client.  It does not implement
+this specification's interfaces but consumes an object that implements them (a Client).
 
 ## Client
 
@@ -31,7 +32,7 @@ A Client MAY:
 
 * Elect to send an altered HTTP request from the one it was provided.  For example, it could
 compress an outgoing message body.
-* Elect to alter a received HTTP response before returning it to the caller. For example, it could
+* Elect to alter a received HTTP response before returning it to the calling library. For example, it could
 decompress an incoming message body.
 
 If a Client chooses to alter either the HTTP request or HTTP response, it MUST ensure that the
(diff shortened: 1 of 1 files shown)


proposed/http-client/http-client.md
```diff
@@ -7,43 +7,63 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
+[Liskov substitutions principle][Liskov]. This means that all clients MUST behave in the
+same way when sending a request.
 
-### Exceptions
+## Definitions
 
-All exceptions thrown by the client MUST implement `Psr\Http\Client\ClientExceptionInterface`.
+* **Client** - A Client is a library that implements this specification for the purposes of
+sending PSR-7-compatible HTTP Request messages and returning a PSR-7-compatible HTTP Response message to a Caller.
+* **Caller** - A Caller is any code that makes use of a Client.
```
Crell: I think in PSR-6 we called this the "calling library", so maybe we should use that term here as well?

Nyholm: Dont we use "consuming library" or something like that?

proposed/http-client/http-client.md
```diff
@@ -38,21 +38,24 @@ If a Client chooses to alter either the HTTP request or HTTP response, it MUST e
...
-Smaller issues that do not block the client from sending the request (such as
-invalid HTTP versions) MUST NOT result in exceptions.
+A Client MUST NOT treat a well-formed HTTP request or HTTP response as an error condition.  For example, response
+status codes in the 400 and 500 range MUST NOT cause an exception and MUST be returned to the Caller as normal.
 
-If the remote server answers with a response that can be parsed into a PSR-7 response,
-the client MUST NOT throw an exception. For example, response status codes in the
-400 and 500 range MUST NOT cause an exception.
+A Client MUST throw an instance of `Psr\Http\Client\ClientExceptionInterface` if and only if it is unable to send
+the HTTP request at all or if no HTTP response is ever returned (such as a timeout).
```
Nyholm: Timeout is covered by the `NetworkExceptionInterface`. Say something line: 

"or if the HTTP response cannot be parsed into a PSR-7 response.".

proposed/http-client/http-client.md
```diff
@@ -60,17 +80,6 @@ interface ClientInterface
...
      * Sends a PSR-7 request and returns a PSR-7 response.
      *
-     * Every technically correct HTTP response MUST be returned as-is, even if it represents an HTTP
-     * error response or a redirect instruction. The only special case is 1xx responses, which MUST
-     * be assembled in the HTTP client.
-     *
-     * The client MAY do modifications to the Request before sending it. Because PSR-7 objects are
-     * immutable, one cannot assume that the object passed to ClientInterface::sendRequest() will be the same
-     * object that is actually sent. For example, the Request object that is returned by an exception MAY
-     * be a different object than the one passed to sendRequest, so comparison by reference (===) is not possible.
```
dbu: i am not on the slack where you discussed, but i would see value in repeating (in condensed form) the most important properties here. the interface is much more likely to be seen by users than the full spec. what would you think of keeping an abbreviated version of this information on the interface?

proposed/http-client/http-client.md
dbu: but its not restricted to libraries, it could also be an application that is using a PSR HTTP client. seems fine to me as you put it now.

Crell: This is why a definitions section is useful, to define precisely what "library" means. ;-)

Here's what we used in PSR-6, for reference:

```
Calling Library - The library or code that actually needs the cache
services. This library will utilize caching services that implement this
standard’s interfaces, but will otherwise have no knowledge of the
implementation of those caching services.
```

proposed/http-client/http-client.md
Crell: In concept I can see the value in doing so, but in this case I cannot think of any notes worth replicating.  Things like "the object may not be the same" are rather edge-case-y.

Crell:
@dbu "Library" is a very generic term.  In this case, Client is specifically referring to "the thing that has an object that implements ClientInterface".  What it is packaged as is irrelevant.  "Caller" is a code that calls `sendRequest()` and expects a thing to happen.

proposed/http-client/http-client.md
dbu: cool if you can add this, adapted for sending requests instead of caching!

proposed/http-client/http-client.md
dbu: i was mainly thinking about the first part that error responses are not exceptions. but if you think its unnecessary, i am ok with this PR.

proposed/http-client/http-client.md
Crell: Pushed a new commit that changes `Caller` to `Calling Library` and tweaks the definition a bit.

proposed/http-client/http-client.md
```diff
@@ -7,43 +7,64 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
+* Reassemble a multi-step HTTP 1xx response itself so that what is returned to the Calling Library is an valid HTTP response
+of status code 200 or higher.
+
+## Error handling
+
+A Client MUST NOT treat a well-formed HTTP request or HTTP response as an error condition.  For example, response
+status codes in the 400 and 500 range MUST NOT cause an exception and MUST be returned to the Calling Library as normal.
+
+A Client MUST throw an instance of `Psr\Http\Client\ClientExceptionInterface` if and only if it is unable to send
+the HTTP request at all or if no HTTP response is ever returned (such as a timeout).
```
dbu: @Nyholm i think your comment here about timeout as an example got hidden by github. do you still think it should be addressed?

dbu:
thanks for the update @Crell, i am now happy with this.

proposed/http-client/http-client.md
```diff
@@ -7,43 +7,64 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
-invalid HTTP versions) MUST NOT result in exceptions.
+A Client is an object implementing `ClientInterface`.
 
-If the remote server answers with a response that can be parsed into a PSR-7 response,
-the client MUST NOT throw an exception. For example, response status codes in the
-400 and 500 range MUST NOT cause an exception.
+A Client MAY:
 
-## Goal
+* Elect to send an altered HTTP request from the one it was provided.  For example, it could
```
Nyholm: Is "Elect" the correct word? It probably is but this word is only in my vocabulary if you talk about politicians. But that might only be me.

proposed/http-client/http-client.md
```diff
@@ -7,43 +7,64 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
+compress an outgoing message body.
+* Elect to alter a received HTTP response before returning it to the calling library. For example, it could
+decompress an incoming message body.
 
-The goal of this PSR is to allow developers to create libraries decoupled from HTTP client
-implementations. This will make libraries more reusable as it reduces the number of
-dependencies and lowers the likelihood of version conflicts.
+If a Client chooses to alter either the HTTP request or HTTP response, it MUST ensure that the
+response remains internally consistent.  For example, if a Client chooses to decompress the message
```
Nyholm: Double space before "For example,"

proposed/http-client/http-client.md
```diff
@@ -7,43 +7,64 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
+the one passed to `sendRequest()`, so comparison by reference (===) is not possible.
+
+A Client MUST:
+
+* Reassemble a multi-step HTTP 1xx response itself so that what is returned to the Calling Library is an valid HTTP response
+of status code 200 or higher.
+
+## Error handling
+
+A Client MUST NOT treat a well-formed HTTP request or HTTP response as an error condition.  For example, response
```
Nyholm: Double space before "For example,"

proposed/http-client/http-client.md
```diff
@@ -7,43 +7,64 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
+status codes in the 400 and 500 range MUST NOT cause an exception and MUST be returned to the Calling Library as normal.
+
+A Client MUST throw an instance of `Psr\Http\Client\ClientExceptionInterface` if and only if it is unable to send
+the HTTP request at all or if no HTTP response is ever returned (such as a timeout).
+
+If a request cannot be sent because the request message is not a valid HTTP request, the Client MUST throw an instance
+of `Psr\Http\Client\RequestExceptionInterface`.
```
Nyholm: Just note. Sending a Request with HTTP version 5.0 is not invalid. Should we mention something like that here?

proposed/http-client/http-client.md
dbu: not a native speaker, but my intuition says it is not wrong to say "elect" in this context. but "decide" may be easier to understand and would be correct too. wdyt @Crell ?

Crell: "Elect" is technically valid English here, but it is rather formal.  I've changed it to "Choose".

proposed/http-client/http-client.md
Crell: That's a can of worms because then you get into defining what constitutes "valid".  Perhaps the slightly less restrictive "message is malformed"?  (I informally get the intent here, but specifying it formally is tricky.)

proposed/http-client/http-client.md
Crell: Updated, and also added a mention of timeouts to `NetworkExceptionInterface`.

proposed/http-client/http-client.md
Nyholm: Exactly. Messages that are not malformed is a good expression. 

Like messsges without a URL or host is impossible to send.

PR #1075, opened 2018-09-12
Edited PSR-15 files for small typos/grammar fixes.

DocFX:
Only non-structural changes, just a few punctuation/wording/grammar fixes.

### DIFF
--- a/accepted/PSR-15-request-handlers.md
+++ b/accepted/PSR-15-request-handlers.md
@@ -6,7 +6,7 @@ This document describes common interfaces for HTTP server request handlers
 that use HTTP messages as described by [PSR-7][psr7] or subsequent
 replacement PSRs.
 
-HTTP request handlers are a fundamental part of any web application. Server side
+HTTP request handlers are a fundamental part of any web application. Server-side
 code receives a request message, processes it, and produces a response message.
 HTTP middleware is a way to move common request and response processing away from
 the application layer.
@@ -65,7 +65,7 @@ specific HTTP message implementation.
 
 ### 1.4 Handling Exceptions
 
-It is RECOMMENDED that any application using middleware include a component
+It is RECOMMENDED that any application using middleware includes a component
 that catches exceptions and converts them into responses. This middleware SHOULD
 be the first component executed and wrap all further processing to ensure that
 a response is always generated.


accepted/PSR-15-request-handlers.md
```diff
@@ -6,7 +6,7 @@ This document describes common interfaces for HTTP server request handlers
 that use HTTP messages as described by [PSR-7][psr7] or subsequent
 replacement PSRs.
 
-HTTP request handlers are a fundamental part of any web application. Server side
+HTTP request handlers are a fundamental part of any web application. Server-side
```
shadowhand: Not sure I agree with this change. While hyphenating might be more precise, I have seen it with and without the hyphen. Would this modification really make the document better?

accepted/PSR-15-request-handlers.md
```diff
@@ -65,7 +65,7 @@ specific HTTP message implementation.
 
 ### 1.4 Handling Exceptions
 
-It is RECOMMENDED that any application using middleware include a component
+It is RECOMMENDED that any application using middleware includes a component
```
shadowhand: This doesn't seem grammatically correct me.

Crell: English being the hodge podge of weirdness that it is, "application includes" or "applications include".  One or the other but not both gets an S.  (It's a XOR operation.)

Don't ask me, I don't understand this wonky language we use...

ashnazg: This one is straightforward... subject+verb is application+includes... both are singular.  A plural-ish prepositional phrase between the two is the usual blame for getting the verb wrong.

accepted/PSR-15-request-handlers.md
DocFX: Just something I saw along the way by going quite far into typos.
Answer is here, I believe: https://en.wikipedia.org/wiki/Server-side

accepted/PSR-15-request-handlers.md
DocFX: The real reason behind this is context. The next sentence uses singular and not plural. Was just for cohesion.

accepted/PSR-15-request-handlers.md
DocFX: It's compound when used as a concept: "server-side".
It's separated when used as two words: "It's a concern that happens to be on the server('s) side.".

PR #1081, opened 2018-09-26
[PSR-19] @throws - also include subtypes of Throwable

ondrejmirtes:
Previous wording excluded interfaces that extend Throwable and also instances of TypError etc.

We could leave in only the expression of being "subtype of Throwable" but I think that might be less clear for someone.

### DIFF
--- a/proposed/phpdoc-tags.md
+++ b/proposed/phpdoc-tags.md
@@ -1061,7 +1061,7 @@ DocBlock.
 ### 5.19. @throws
 
 The @throws tag is used to indicate whether "Structural Elements" throw a
-specific type of exception.
+specific type of Throwable (exception or error).
 
 #### Syntax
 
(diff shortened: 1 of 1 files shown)


proposed/phpdoc-tags.md
```diff
@@ -1072,8 +1072,7 @@ specific type of exception.
 The @throws tag MAY be used to indicate that "Structural Elements" throw a
 specific type of error.
 
-The type provided with this tag MUST represent an object of the class Exception
-or any subclass thereof.
+The type provided with this tag MUST represent an object of the class Exception, or any subclass thereof, or an object that is a subtype of Throwable.
```
ashnazg: Maybe we should just go with a blanket `Throwable`, and leave it at that?  I think Throwable might have come along some time after Mike originally wrote this section.

ashnazg:
@ondrejmirtes can you bring this up for discussion on the ML?

ashnazg:
@ondrejmirtes rather than a new ML thread, if you want to update this to just show `Throwable` rather than `Exception and stuff`, then we can get it approved and merged.  This seems simple enough a change to fit in with our Simple Baseline First Draft approach.

PR #1082, opened 2018-09-26
PSR-5: Fix formatting and indentation in FQSEN definition

robbieaverill:


proposed/phpdoc.md
```diff
@@ -173,27 +173,27 @@ interpreted as described in [RFC 2119][RFC2119].
...
-  *Namespace*:      `\My\Space`
-  *Function*:       `\My\Space\myFunction()`
-  *Constant*:       `\My\Space\MY_CONSTANT`
-  *Class*:          `\My\Space\MyClass`
-  *Interface*:      `\My\Space\MyInterface`
-  *Trait*:          `\My\Space\MyTrait`
-  *Method*:         `\My\Space\MyClass::myMethod()`
-  *Property*:       `\My\Space\MyClass::$my_property`
-  *Class Constant*: `\My\Space\MyClass::MY_CONSTANT`
+  * *Namespace*:      `\My\Space`
```
ashnazg: Are the leading asterisks here necessary?  I realize they're trying to be bullets, but I've learned to avoid using asterisks and underscores in Markdown unless I want italics, because that's typically how they've always been parsed.  Since no bullets were required beforehand, maybe they can continue to be left off.

robbieaverill: Yes, but that part is indented so renders in a pre-formatted block

ashnazg: Ah, ok... if the bullets are rendering good for you in this top section, then maybe that's fine to keep.

robbieaverill: I could switch to `-` for bullets if you'd prefer, to avoid confusion with using `*` for that and italics?

ashnazg:
:+1:

PR #1085, opened 2018-09-28
Finalize PSR-14

Crell:
This brings us up to ready-for-readiness vote, I think.

### DIFF
--- a/proposed/event-dispatcher.md
+++ b/proposed/event-dispatcher.md
@@ -255,7 +255,7 @@ interface StoppableTaskInterface extends TaskInterface
 interface TaskProcessorInterface
 {
     /**
-     * Provide all listeners with a task event to process.
+     * Dispatches a Task to all registered listeners.
      *
      * @param TaskInterface $event
      *  The task to process.
--- a/proposed/event-dispatcher.md
+++ b/proposed/event-dispatcher.md
@@ -226,7 +226,7 @@ interface TaskInterface extends EventInterface
 namespace Psr\EventDispatcher;
 
 /**
- * A Task whose processing may be interrupted when the task is complete.
+ * A Task event whose processing my be interrupted when a listener has completed processing this event.
  *
  * A Processor implementation MUST check to determine if a Task
  * is marked as stopped after each listener is called.  If it is then it should
--- a/proposed/event-dispatcher.md
+++ b/proposed/event-dispatcher.md
@@ -137,4 +137,132 @@ It is RECOMMENDED that a Notifier and Processor be implemented as distinct objec
 
 ## Interfaces
 
-See [the repository](https://github.com/php-fig/event-dispatcher).  The final interfaces will be copied back here before it goes up for a vote.  For now it's easier to just single-source them.
+### General event interfaces
+
+```php
+namespace Psr\EventDispatcher;
+
+/**
+ * Marker interface indicating an event instance.
+ *
+ * Event instances may contain zero methods, or as many methods as they
+ * want. The interface MUST be implemented, however, to provide type-safety
+ * to both listeners as well as listener providers.
+ */
+interface EventInterface
+{
+}
+```
+
+```php
+namespace Psr\EventDispatcher;
+
+/**
+ * Mapper from an event to the listeners that are applicable to that event.
+ */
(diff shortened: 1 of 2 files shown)


proposed/event-dispatcher.md
```diff
@@ -137,4 +137,132 @@ It is RECOMMENDED that a Notifier and Processor be implemented as distinct objec
...
+interface TaskInterface extends EventInterface
+{
+}
+```
+
+```php
+namespace Psr\EventDispatcher;
+
+/**
+ * A Task whose processing may be interrupted when the task is complete.
```
bmack: This seems a bit out of context (looking at the other docblock comments):

How about
"This interface defines a task event whose processing my be interrupted when a listener has completed this task event."

proposed/event-dispatcher.md
```diff
@@ -137,4 +137,132 @@ It is RECOMMENDED that a Notifier and Processor be implemented as distinct objec
...
+{
+    /**
+     * Provide all listeners with a task event to process.
+     *
+     * @param TaskInterface $event
+     *  The task to process.
+     *
+     * @return TaskInterface
+     *  The task that was passed, now modified by callers.
+     */
```
bmack: Should we write down the exception handling?

Crell: We don't elsewhere.  I don't want to get into too much detail in the docblocks.

proposed/event-dispatcher.md
Crell: Meet in the middle: "A Task event whose processing my be interrupted when a listener has completed processing this event."

Crell: Remind me to update the code package when this is merged, too. :smile:

proposed/event-dispatcher.md
bmack: K

proposed/event-dispatcher.md
```diff
@@ -137,4 +137,132 @@ It is RECOMMENDED that a Notifier and Processor be implemented as distinct objec
...
+```
+
+```php
+/**
+ * Defines a processor for task events.
+ */
+interface TaskProcessorInterface
+{
+    /**
+     * Provide all listeners with a task event to process.
```
bmack: That sounds strange.

In the message events we write "Notify listeners of a message event." here it's  "Provide all listeners with a task event to process." I'd go with "Dispatches a task to all registered listeners"?

proposed/event-dispatcher.md
Jean85: Other PSRs do it BTW:
 * https://github.com/php-fig/fig-standards/blame/master/accepted/PSR-11-container.md#L85-L86
 * https://github.com/php-fig/fig-standards/blame/master/accepted/PSR-16-simple-cache.md#L149
 * https://github.com/php-fig/fig-standards/blame/master/accepted/PSR-17-http-factory.md#L136-L137
 * https://github.com/php-fig/fig-standards/blame/master/accepted/PSR-6-cache.md#L287
 * https://github.com/php-fig/fig-standards/blame/master/accepted/PSR-7-http-message.md#L724
 * etc..

Crell: If there's a known exception that the object will throw, yes.  But in this case it could literally throw anything, or rather, it's not allowed to prevent anything that gets thrown from getting caught.  `@throws \Throwable Could be anything.` doesn't seem particularly useful.

proposed/event-dispatcher.md
Crell: Works for me.  Changed.

proposed/event-dispatcher.md
Jean85: Oh yeah, so then that's the right way to proceed IMHO. I think PSR-15 does the same.

Crell: That seems... silly, honestly.  Any function that calls any function could conceivably throw anything, or rather not catch it.  Yet nobody (AFAIK) puts `@throws \Throwable` on every function in their API. Only those exceptions that are directly relevant to and generated by the library.

Jean85: Sorry, I meant that PSR-15 did as you are doing, so all fine! :+1:

PR #1087, opened 2018-10-02
Downstream changes from phpDocumentor/fig-standards fork

ashnazg:


ashnazg:
There were simple PRs at the old fork that mvriel used previously.  I now have it updated, so we'll use it as the "interop"-ish sandbox repo for PSRs 5 & 19.

ashnazg:
Ping @michaelcullum @Jean85 ... I think @neuro159 might be missing from the WG members list in the PullApprove config.

PR #1088, opened 2018-10-06
PSR "ID" renaming for sort-order

dan-lugg:
# Overview

Currently the accepted PSRs use the file-naming scheme (exemplified with good old `printf`) of `PSR-%d`.

This PR aims to change and further propose enforcement of the scheme `PSR-%03d`. This ensures that even the most naive file-sorting algorithms will order the listing by "ID" naturally in order of acceptance.

# Scope

The objective of this PR, and related discussion, is to change the file-naming scheme.

Changes to notation in other cases (such as simply referencing `PSR-{n}` in text) are beyond the intended scope here. Frankly, I don't feel those cases should have this change imposed.

Example:

```markdown
... [something regarding PSR-7](/accepted/PSR-007-http-message-meta.md)
```

# `PSR-%03d`

The proposed naming scheme is `PSR-%03d` (right-aligned, 0-padded, 3-digit). This scheme should be subsequently followed for accepted PSRs to ensure future sort-order adherence.

Of course, 4-digit, 42-digit, etc., schemes are an open point of discussion; I personally do not favor any width >3 over another. 3-digit was used for what I believe to be reasonable visual width, future-proofability, and triviality of search-replace.


dan-lugg:
Identifying non-compliant markdown links can be achieved with the PCRE `PSR-\d{1,2}[a-zA-Z_-]*\.md`.

jkufner:
This will break all links around the Internet pointing to the renamed PSRs.

ondrejmirtes:
I don’t think the renaming is necessary. “PSR-7” reads much better than “PSR-007”. Also, you can implement custom sorting algorithm in places where you list all the PSRs...

Jean85:
I'm also :-1: on this, for the same reasons as the others; also, this could broke a lot of stuff on the website.

Since I see that those reply are endorsed too, I will decline this, I'm sorry.

PR #1090, opened 2018-10-10
Add return type rule for closures

Jean85:
As [suggested on Twitter](https://twitter.com/BackEndTea/status/1049992124463292416), we should explicitly state how the return type of a closure should be formatted. I've followed (and cited) the same rule as for normal functions/methods.


BackEndTea:
How about adding an example with a return type as well?

PR #1097, opened 2018-10-15
Downstream changes from phpDocumentor/fig-standards fork

ashnazg:


ashnazg:
These will be the last changes sent from the phpDocumentor fork... see https://groups.google.com/d/msg/php-fig/oYo0cZ-BBUk/87j32TXBBQAJ

proposed/phpdoc-tags.md
```diff
@@ -824,15 +824,15 @@ function count(array $items)
...
+an option array with two elements, representing 'required' and 'label'.
 
 ```php
 /**
  * Initializes this class with the given options.
  *
  * @param array $options {
- *     @var bool   $required Whether this element is required
- *     @var string $label    The display name for this element
+ *     @var bool   Whether this element is required
```
neuro159: This part is useless - what is bool and what is string in this array? Key/Value? some particular named keys?

ashnazg: I think that represents a problem with the Inline PHPDoc syntax overall, so we should probably start that conversation anew separately.  In here, my main thing was recognizing that the `@var` syntax was showing variable names that actually do not exist in the code... they were simply given names to hint at their intent, while ordering of the tags is what's supposed to align with ordering of the array members.

My inclination is to consider dropping the Inline PHPDocs, and either let them go or start the discussion anew... like we're doing with generics.

ashnazg:
@neuro159 I think that represents a problem with the Inline PHPDoc syntax overall, so we should probably start that conversation anew separately. In here, my main thing was recognizing that the @var syntax was showing variable names that actually do not exist in the code... they were simply given names to hint at their intent, while ordering of the tags is what's supposed to align with ordering of the array members.

My inclination is to consider dropping the Inline PHPDocs, and either let them go or start the discussion anew... like we're doing with generics.

With this PR, I really just hope to get the other fixes into the drafts.  I do have plans to open the floor on Inline PHPDocs as a whole topic of its own, later on.

muglug:
> My inclination is to consider dropping the Inline PHPDocs, and either let them go or start the discussion anew... like we're doing with generics.

I agree with that, but I'm still unsure how the change in this PR makes sense? I suppose this idea is for an array like `[true, "hello"]` but the diff complicates matters. Why not just scrap it?

PR #1099, opened 2018-10-15
revert deprecation of link tag

ashnazg:
Deprecating the `@link` tag (in deference to relying on `@see` for the same behavior) was a proposal in the early days of PSR-5 that received a negative response.  A decision *was* made back then (https://github.com/phpDocumentor/fig-standards/issues/104#issuecomment-138111143) to revert the deprecation, but the old draft was never updated.  This PR handles that.

This PR is solely to baseline the status of the `@link` tag back to normal.  A new discussion to restore the deprecation can be made if parties wish to bring it before the Working Group.

### DIFF
--- a/proposed/phpdoc-tags.md
+++ b/proposed/phpdoc-tags.md
@@ -23,7 +23,7 @@ PSR-19: PHPDoc tags
   - [5.7.  @global](#57-global)
   - [5.8.  @internal](#58-internal)
   - [5.9.  @license](#59-license)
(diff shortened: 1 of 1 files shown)


ashnazg:
Ping WG... I'd like to get this merged before more people review the drafts.

ashnazg:
@Jean85 bump

Jean85:
Merging anyway because Pullapprove doesn't seem to consider the previous review from @neuro159

PR #1100, opened 2018-10-15
deprecate double closing brace on inline internal tag

ashnazg:


ashnazg:
Discussion should go on the mailing list thread (https://groups.google.com/forum/#!topic/php-fig/KckIDHss-QI).

PR #1106, opened 2018-10-26
[PSR-5] Update Summary grammar

shadowhand:
As per [discussion on the mailing list][1], requires that a summary be
separated from other content by two line breaks.

[1]: https://groups.google.com/d/msg/php-fig/59qEeX-VkZU/ZpErzi1fBgAJ

### DIFF
--- a/proposed/phpdoc.md
+++ b/proposed/phpdoc.md
@@ -245,7 +245,7 @@ definition:
 
     PHPDoc             = [summary] [description] [tags]
     inline-phpdoc      = "{" *SP PHPDoc *SP "}"
-    summary            = *CHAR ("." 1*CRLF / 2*CRLF)
+    summary            = *CHAR (2*CRLF)
     description        = 1*(CHAR / inline-tag) 1*CRLF ; any amount of characters
                                                      ; with inline tags inside
     tags               = *(tag 1*CRLF)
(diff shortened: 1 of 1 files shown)


Crell:
I completely disagree here, actually.  A summary should be a complete sentence, and a complete sentence ends with a period.  Always.  

I didn't understand the discussion thread on the list going in this direction, but explicitly saying no period is a massive BC break and a very bad one to do.

I would go the other direction entirely.  Always a period, and 2 carriage returns if there's anything after it.

shadowhand:
@Crell dang, I just reviewed the existing [phpdoc demo](http://demo.phpdoc.org/Responsive/classes/Cilex.Provider.JmsSerializerServiceProvider.html) and [Symfony API docs](https://api.symfony.com/4.1/Symfony/Component/DependencyInjection/Alias.html) and it looks like the period is the current standard. I could have sworn it was the other way around. 😭 

I will update this PR.

proposed/phpdoc.md
```diff
@@ -266,10 +266,8 @@ A Summary MUST contain an abstract of the "Structural Element" defining the
 purpose. It is RECOMMENDED for Summaries to span a single line or at most two,
 but not more than that.
 
-A Summary MUST end with either
-
-* a full stop (.) followed by a line break
-* or two sequential line breaks.
+A Summary MUST end with a full stop (.) and if there is additional content it
+MUST be separated from the Summary by two sequential line breaks.
```
Crell: I had to read this sentence twice to understand it.  It sounds like "A summary must... and it must be separated from the summary..."  Which makes me think you're separating the summary from the summary. :smile: (I think it's because Summary is the only proper noun in the sentence.)

Suggested alternative: 

```
A Summary MUST end with a full stop (.).  

If there is any additional content, the Summary MUST be followed by two sequential line breaks to separate it from the additional content.
```

shadowhand:
@miguelrs I tend to agree but I don't want to broaden the scope of this PR.

shadowhand:
@ashnazg @Crell updated.

ashnazg:
@miguelrs that's probably a good conversation to have... care to open a new PR for that?

PR #1107, opened 2018-10-26
[PSR-15] Correct summary format

shadowhand:
Periods are included in all existing PSR specifications except PSR-15. This is also consistent with current PSR-5 proposal, which has an explicit requirement for summary format.

Refs #1106

### DIFF
--- a/accepted/PSR-15-request-handlers.md
+++ b/accepted/PSR-15-request-handlers.md
@@ -83,7 +83,7 @@ use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 
 /**
- * Handles a server request and produces a response
+ * Handles a server request and produces a response.
  *
  * An HTTP request handler process an HTTP request in order to produce an
  * HTTP response.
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
👍

JanTvrdik:
There is no PSR-5 standard. It's a draft. Please wait for PSR-5 to stabilize before refactoring  stuff based on it.

iansltx:
Looking through the PSRs in this repo, I'm indeed seeing that this one is one of very few exceptions to the _de facto_ standard of throwing a full stop at the end of summaries...*for this repo*. There are a few other exceptions in PSR-3:

https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-3-logger-interface.md

It looks like the cases where the period is missing in the above were oversights, but I'd like a PR to fix those if we're cleaning this up in the name of consistency as well.

For actual interfaces in accepted PSRs, the interface companion to this also doesn't follow the two-CRLFs-after-one-line-and-a-full-stop rule, though there's a PR to resolve that as well:

https://github.com/php-fig/http-server-middleware/pull/3

I've requested a few reviewers to the above so someone who's worked on the PSR sees what's going on there.

With the above in mind, I'll merge this if:

1. I get another +1 here
2. I get a +1 on the middleware interface PR
3. I get a PR to fix the summary on PSR-3

Weigh in if I'm not being reasonable here :)

shadowhand:
@iansltx #1112 has been added to make PSR-3 consistent.

PR #1111, opened 2018-10-30
Update PSR-18-http-client.md

tomtomsen:
title matches now classname


Jean85:
I would like to have a :+1: from @Nyholm too before merging.

PR #1116, opened 2018-11-15
PullApprove v3 test, take two

Jean85:


Jean85:
Great! This seems to work! :tada:

PR #1117, opened 2018-11-16
Removals of Old-Deprecated and Too-New-To-Be-In-Use things

ashnazg:
Removed these tags altogether, rather than specify them and immediately deprecate them:  category, subpackage, global.

Removed specification of File-Level DocBlock.  Without these, the license tag doesn't have much use, so removed it as well.

Removed specification of Inline PHPDoc, since it's not something that was ever implemented.

Removed Ending Version portion of deprecated tag, since it was a new addition that was also never implemented.

### DIFF
--- a/proposed/phpdoc-tags.md
+++ b/proposed/phpdoc-tags.md
@@ -16,27 +16,26 @@ PSR-19: PHPDoc tags
 - [5. Tags](#5-tags)
   - [5.1.  @api](#51-api)
   - [5.2.  @author](#52-author)
-  - [5.3.  @category [deprecated]](#53-category-deprecated)
-  - [5.4.  @copyright](#54-copyright)
-  - [5.5.  @deprecated](#55-deprecated)
-  - [5.6.  @example](#56-example)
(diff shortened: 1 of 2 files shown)


robbieaverill:
Hi @ashnazg,

Re: 2b8faf6 "Removed Ending Version portion of deprecated tag, since it was a new addition that was also never implemented" - is it accepted that the standard will only provide for the version deprecated _from_ now, and not the version it will be removed? 

I must've missed this from the mailing list, which is fine - just checking to ensure that this is what the PHP-FIG is aiming for going forwards, so I can update our libraries to remove the _to_ version.

Thanks!

ashnazg:
@robbieaverill , I'd say sit tight for the moment on anticipating changes in your projects due to PSR-5/19.  Much of what I've done lately has been to pare down the scope that's in the tag catalog, so that the "standard" won't be painfully expansive.  Anything that I've taken out to do that is fair game to be restored by the WG / PRs if there is a strong interest / case presented for those additions.

PR #1119, opened 2018-11-16
More updates from my notes

ashnazg:


proposed/phpdoc-tags.md
```diff
@@ -955,7 +933,7 @@ Another example is to document the variable in a foreach explicitly; many IDEs
 use this information to help you with auto-completion:
 
 ```php
-/** @var \Sqlite3 $sqlite */
+/* @var \Sqlite3 $sqlite */
```
jaapio: I just checked with phpstorm, `/*` and `/**` do work both. But `/**` has some improved auto-completion. Where `/*` requires some more typing. 

I don't know if this document states anything about inline comments like

use `//` for single line inline comments 
use `/*` for multi line inline comments
use `/**` for docblocks, above `fqsen`

at this moment `//` doesn't support tags, at least in phpstorm it doesn't. And I have never seen anythink like that in the wild. But this change triggers me that there should be some statement about inline docs.

proposed/phpdoc-tags.md
```diff
@@ -917,14 +897,10 @@ Each Constant or Property definition or Variable where the type is ambiguous
 or unknown SHOULD be preceded by a DocBlock containing the @var tag. Any
 other variable MAY be preceded with a DocBlock containing the @var tag.
 
-The @var tag MUST contain the name of the element it documents. An exception
+The @var tag SHOULD contain the name of the element it documents. An exception
```
jaapio: This confuses me. I think the `MUST` was better in this case. 
`@var` has to contain the element's name, otherwise the interpreter will never know which element is described. There is just one exception. In case it is used above a single property. 

Maybe add an example of multi property definition?

proposed/phpdoc-tags.md
ashnazg: I went to the `phpdoc.md` format PSR to look at having these three syntaxes mentioned.  I found original verbiage that says the `/** @var */` syntax for `foreach` variables is explicitly **not** defined in the standard.  So, this begs the question of whether or not we *do* want it defined in the standard.  If yes, then the verbiage I mention gets removed... if no, then this `@var` example above gets removed.

ashnazg: My intuition is that everyone *does* want the `foreach @var` usage to be included in the var tag definition, so I've removed the "not in standard" verbiage.

ashnazg:
@jaapio ready to review again.

proposed/phpdoc-tags.md
```diff
@@ -644,13 +644,9 @@ RECOMMENDED in case of complicated return structures, such as associative arrays
...
 delimiting.
 
-It is RECOMMENDED to use this tag with every function and method. An exception to
-this recommendation, as defined by the Coding Standard of any individual project,
-MAY be:
-
-   **functions and methods without a `return` value**: the @return tag MAY be
-   omitted here, in which case an interpreter MUST interpret this as if
-   `@return void` is provided.
+It is RECOMMENDED to use this tag with every function and method. The @return
```
JanTvrdik: I would suggest to keep the recommendation in the specification to minimum. For example here lot of people may argue that it's pointless to add `@return int` to methods with native `int` return type.

proposed/phpdoc-tags.md
```diff
@@ -644,13 +644,9 @@ RECOMMENDED in case of complicated return structures, such as associative arrays
...
+tag MAY be omitted by functions and methods without a `return` value, in which
+case an interpreter MUST interpret this as if `@return void` is provided.
```
muglug: I don't agree with this (allowing tag to be emitted in the case of `void` return and then treating everything without it as returning `void`).

- As above, a user should not have to add a `@return int` to a function with an `int` return type. This rule would mean treating that function as if it had returned `void`. That makes no sense to me.
- Code with `return` statements but no `@return` is far more common, in my experience, than code with incorrect `@return` types. If implementers of static analysis tools abided by this rule we'd break a lot of code that uses third-party libraries:
    ```php
    $a = third_party_function_without_docblock(); // AssignmentToVoid issue in Psalm
    ```

I propose instead always requiring an `@return` type of some description if there's no typehint given.

ashnazg: Yep, was going to point this out after Jan's comment reminded me of return type in code. This verbiage predates that being added to PHP.

muglug: I'd suggest
```
It is RECOMMENDED to use this tag with every function and method.
If no `@return` type is given, and no return type declaration is provided,
an interpreter MUST interpret this as if `@return mixed` is provided.
```

ashnazg:
This discussion on the return tag has me rethinking something I keep forgetting to bring up... should the PSRs only define **how** to make a given Tag, and **where* it _can_ be used... or should they go further and specify that certain Tags **MUST/MUST NOT** be used in certain scenarios?  I think there is lingering old verbiage from early drafts that do the latter... and I'm pondering if we should drop those bits entirely.

proposed/phpdoc-tags.md
ashnazg: I don't think any implementation can be required to parse the method body itself in search of `return` statements, in order to derive types when they are not already specified by tag or code signature.  In those cases, @muglug 's suggestion of defaulting to `mixed` seems good enough to me.

jaapio:
I think there are no scenarios where `@return` or `@param` must be defined, when we are defining a PSR for php 7.1+. But since this PRS should be compatible with the real world where type hints are omitted and only deep code analysis could determine the real type of return values and params I would like to state that when no native types are defined we should `RECOMMEND` the usage of tags. 

I think the `@param`, `@var` and `@return` are the only tags that need a recommendation or stronger guide lines about there required usage.

Jean85:
Would you reccomend `@throws` too?

Crell:
Offering RECOMMENDED/NOT RECOMMENDED guidelines seems wise, if for no other reason than consistency since people are, presumably, going to look to the spec for guidance on "so what good is this and when should I use it?"  MUST/MUST NOT for tag usage seems like it's probably an over-reach that will get a lot of pushback, even if I personally would favor "must always have full param and return tags". :smile:

Jean85:
But  "must always have full param and return tags" may not endure the test of time.. Since the advent of PHP 7, a lot of people started dropping those in favor of native arguments/return types.

ashnazg:
Right... I think it's PHPStan that I remember in the phpDocumentor builds that will complain if you include a tag with "redundant" info that's provided by typehints in the code signature.  So I think that's one WG implementation that already would not want to force all tags all the time.

jaapio:
It is true that people do drop non documented tags in favor of native arguments and return types. However as I stated before php developers tend to use mixed return types which cannot be defined native. Although it is a bad practice even php itself does this.

And yes I would recommend the usage of @throws when a method throws an exception. Even more than @return, since there is no native way to do this.

Crell:
@Jean85 Right, which is why I think RECOMMENDED is better to include than MUST. :smile: But including various RECOMMENDEDs is fine.

JanTvrdik:
Bad recommendation is a lot worse than not recommending anything.

proposed/phpdoc-tags.md
glen-84: Doesn't this recommendation also contradict the "This document SHALL NOT ... Describe best practices or recommendations for Coding Standards on the application of the PHPDoc standard."?

proposed/phpdoc-tags.md
Isinlor: I just want to mention that `MUST` seems to _exclude_ option of using more precise type, while implementation should be allowed to do so.

ashnazg:
@muglug ping for review... can we get this PR merged, and continue its open discussions for changes in my next PR?

proposed/phpdoc-tags.md
```diff
@@ -641,16 +638,10 @@ function or method. When provided, it MUST contain a "Type"
...
-this recommendation, as defined by the Coding Standard of any individual project,
-MAY be:
-
-   **functions and methods without a `return` value**: the @return tag MAY be
-   omitted here, in which case an interpreter MUST interpret this as if
-   `@return void` is provided.
+It is RECOMMENDED to use this tag with every function and method.
+If no `@return` type is given, and no return type declaration is provided in
+the code signature, an interpreter MUST interpret this as if `@return mixed`
+is provided.
```
JanTvrdik, line 644: So is the interpreter not allowed to look at native return type?

proposed/phpdoc-tags.md
```diff
@@ -917,9 +892,9 @@ Each Constant or Property definition or Variable where the type is ambiguous
 or unknown SHOULD be preceded by a DocBlock containing the @var tag. Any
 other variable MAY be preceded with a DocBlock containing the @var tag.
 
-The @var tag MUST contain the name of the element it documents. An exception
-to this is when property declarations only refer to a single property. In this
-case the name of the property MAY be omitted.
+The @var tag MUST contain the name of the element it documents, unless this
+property declaration only refers to a single property. In this case the name of
+the property MAY be omitted.
```
JanTvrdik, line 897: Maybe

```suggestion
the property SHOULD be omitted.
```

Writing the property name is quite strange to be honest.

~~~php
/** @var string $name */
private $name;
~~~

JanTvrdik: Also do we intentionally disallow ommiting element name when using it with variable assignment?

```php
/** @var string */
$name = 'John';
```

proposed/phpdoc.md
```diff
@@ -109,6 +118,7 @@ interpreted as described in [RFC 2119][RFC2119].
   }
   ```
 
+<<<<<<< HEAD
```
JanTvrdik, line 121: .

proposed/phpdoc-tags.md
jaapio: Please re-read this sentence. It states also "when no return type is defined in the code signature" 😉

proposed/phpdoc-tags.md
jaapio: Regarding the property definition, this is written this way because of multi properties defenition in a single line.
When you add the property name the @var will only be applied to that property. Otherwise it will be applied to all properties.

But I agree it needs some clarification.

ashnazg:
Closing this, as it really has too many things on it to have targeted discussions.

PR #1120, opened 2018-11-20
Remove Symfony

fabpot:


drupol:
Why this change?

Jan0707:
What's the reasoning here?

javiereguiluz:
@drupol @Jan0707 Fabien explained the reasons in this Twitter thread: https://twitter.com/fabpot/status/1064946698089365505

mickaelandrieu:
I don't understand you here @fabpot: in which case(s) removing Symfony project from PHP Fig will help the interoperability and PHP standards in the future? 

Do we want to come back 6 years ago with custom autoloading, custom coding styles and custom ways to do everything ?

Pierstoval:
> Do we want to come back 6 years ago with custom autoloading, custom coding styles and custom ways to do everything ?

Is this really about autoloading and coding style? Because Fabien's thread is clear about this: they both are great PSRs, it's mostly the "new wave of incoming PSRs" that's not interop and kinda breaks the initial project into a more opinionated set of rules instead of an interop one

gmponos:
> Is this really about autoloading

Go back when the PSR-0 came up... Can you imagine back then how many projects could not follow that? Wasn't the PSR-0 not interop back then?

Bilge:
@fabpot is completely correct, on all points. And to his point about PSR-7 specifically, there's actually an inherent security flaw in its design pertaining to file uploads. I tried to alert the FIG to it some years ago, but given that they [deliberately hide the GitHub issues feature from the repository](https://github.com/php-fig/http-message), I took it as a strong indicator that they don't want to know.

michaelcullum:
This isn't a place for this discussion. Please use the mailing list.

PR #1128, opened 2018-12-19
PSR 14 Updates

Crell:
See: https://github.com/php-fig/event-dispatcher/pull/19
See: https://github.com/php-fig/event-dispatcher-util/pull/3

### DIFF
--- a/proposed/event-dispatcher-meta.md
+++ b/proposed/event-dispatcher-meta.md
@@ -37,7 +37,7 @@ The Working Group identified four possible workflows for event passing, based on
 * Collection.  ("Give me all your things, that I may do something with that list.")
 * Alternative chain.  ("Here's a thing; the first one of you that can handle it do so, then stop.")
 
-On further review, it was determined that:
+On further review, the Working Goup determined that:
 
 * Collection was a special case of Object enhancement (the collection being the object that is enhanced)
 * Alternative chain is similarly a special case of Object enhancement, as the signature is identical and the dispatch workflow is nearly identical, albeit with an extra check included.
@@ -52,7 +52,7 @@ Although in concept one-way notification can be done asynchronously (including d
 * Passing a collection to a series of Listeners to allow them to register values with it so that the Emitter may act on all of the collected information.
 * Passing a collection to a series of Listeners to allow them to modify the collection in some way before the Emitter takes action.
 * Passing some contextual information to a series of Listeners so that all of them may "vote" on what action to take, with the Emitter deciding based on the aggregate information provided.
-* Passing an object to a series of listeners and allowing one of them to set a value and then prevent further listeners from running.
+* Passing an object to a series of listeners and allowing any listener to terminate the process early before other listeners have completed.
 
 ### 4.3 Immutable events
 
@@ -68,9 +68,11 @@ However, Stoppable Events (the alternative chain case) also needed to have a cha
 * Returning a sentinel value from the listener (`true` or `false`) to indicate that propagation should terminate.
 * Evolving the Event to be stopped (`withPropagationStopped()`)
 
-Of those, the first would mandate a mutable Event in at least some cases.  The second would mandate a mutable Event as the return value was already in use.  And the third seemed unnecessarily ceremonial and pedantic.
+Of those alternatives, therefore, the first would mandate a mutable Event in at least some cases.  The second would mandate a mutable Event as the return value was already in use.  And the third seemed unnecessarily ceremonial and pedantic as it would entail additional syntax requirements for developers with little added value.
 
-Having listeners return evolvable events also posed a challenge.  That pattern is not used by any known implementations in PHP or elsewhere.  It also relies on the listener to remember to return the Event (extra work for the listener author) and to not return some other, new object that might not be fully compatible with later listeners (such as a subclass or superclass of the Event).
+Additionally, a desired feature was the ability to derive whether or not to stop propagation based on some values collected from the listeners.  (For example, to stop when one of them has provided a certain value, or after at least three of them have indicated a "reject this request" flag, or similar.)  While technically possible to implement as an evolvable object, such behavior is intrinsically stateful so would be highly cumbersome for both implementers and users.
+
+Having listeners return evolvable events also posed a challenge.  That pattern is not used by any known implementations in PHP or elsewhere.  It also relies on the listener to remember to return the Event (additional work for the listener author) and to not return some other, new object that might not be fully compatible with later listeners (such as a subclass or superclass of the Event).
 
 Immutable events also rely on the event definer to respect the admonition to be immutable.  Events are, by nature, very loosely designed and the potential for implementers to ignore that part of the spec, even inadvertently, is high.
 
@@ -81,20 +83,20 @@ That left two possible options:
 
 Given those options the Working Group felt mutable events were the safer alternative.
 
-That said, there is no requirement that an Event be mutable.  Implementers should provide mutator methods on an Event object if and only if it is necessary and appropriate to the use case at hand.
+That said, *there is no requirement that an Event be mutable*.  Implementers should provide mutator methods on an Event object *if and only if it is necessary* and appropriate to the use case at hand.
 
 ### 4.4 Listener registration
 
 Experimentation during development of the specification determined that there were a wide range of viable, legitimate means by which a Dispatcher could be informed of a listener.  A listener:
 
 * could be registered explicitly;
-* it could be the registered explicitly based on reflection of its signature;
-* it could be registered with a numeric priority order;
-* it could be registered using a before/after mechanism to control ordering more precisely;
-* it could be registered from a service container;
-* it could use a pre-compile step to generate code;
-* it could be based on method names on objects in the event itself;
-* it could be limited to certain situations or contexts based on arbitrarily complex logic (only for certain users, only on certain days, only if certain system settings are present, etc).
+* could be the registered explicitly based on reflection of its signature;
+* could be registered with a numeric priority order;
+* could be registered using a before/after mechanism to control ordering more precisely;
+* could be registered from a service container;
+* could use a pre-compile step to generate code;
+* could be based on method names on objects in the event itself;
+* could be limited to certain situations or contexts based on arbitrarily complex logic (only for certain users, only on certain days, only if certain system settings are present, etc).
 
 These and other mechanisms all exist in the wild today in PHP, all are valid use cases worth supporting, and few if any can be conveniently represented as a special case of another.  That is, standardizing one way, or even a small set of ways, to inform the system of a listener turned out to be impractical if not impossible without cutting off many use cases that should be supported.
 
@@ -102,7 +104,7 @@ The Working Group therefore chose to encapsulate the registration of listeners b
 
 It is even possible, and potentially advisable, to allow libraries to include their own Providers that get aggregated into a common provider that aggregates their listeners to return to the Dispatcher.  That is one possible way to handle arbitrary listener registration within an arbitrary framework, although the Working Group is clear that is not the only option.
 
-While combining the Dispatcher and Provider into a single object is a valid and permissible degenerate case, it is NOT RECOMMENDED as it reduces the flexibility of system integrators.  Instead, the provider should be composed as a dependent object.
+While combining the Dispatcher and Provider into a single object is a valid and permissible degenerate case, it is NOT RECOMMENDED as it reduces the flexibility of system integrators.  Instead, the provider SHOULD be composed as a dependent object.
 
 ### 4.5 Return values
 
--- a/proposed/event-dispatcher-meta.md
+++ b/proposed/event-dispatcher-meta.md
@@ -7,7 +7,7 @@ The purpose of this document is to describe the rationale and logic behind the E
 
 ## 2. Why Bother?
 
-Many libraries, components, and frameworks have long supported mechanisms for allowing arbitrary third party code to interact with them.  Most are variations on the classic Observer pattern, often mediated through an intermediary object or service.  Others take a more Aspect-Oriented Programming (AOP) approach.  Nonetheless all have the same basic concept: interupt program flow at a fixed point to provide information to arbitrary third party libraries with information about the action being performed and allow them to either react or influence the program behavior.
+Many libraries, components, and frameworks have long supported mechanisms for allowing arbitrary third party code to interact with them.  Most are variations on the classic Observer pattern, often mediated through an intermediary object or service.  Others take a more Aspect-Oriented Programming (AOP) approach.  Nonetheless all have the same basic concept: interrupt program flow at a fixed point to provide information to arbitrary third party libraries with information about the action being performed and allow them to either react or influence the program behavior.
 
 This is a well-established model, but a standard mechanism by which libraries do so will allow them to interoperate with more and more varied third party libraries with less effort by both the original developer and extension developers.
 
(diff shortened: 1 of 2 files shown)


proposed/event-dispatcher-meta.md
```diff
@@ -37,85 +37,72 @@ The Working Group identified four possible workflows for event passing, based on
...
+* One-way notification is a degenerate case of the others, or can be represented as such.
 
-Notification can safely be done asynchronously (including delaying it through a queue) but Modification by nature involve passing data back to the caller and thus must be synchronous.  Despite that difference the Working Group determined that the use cases were close enough to be considered in a single PSR.  The two different workflows however are represented by two different interfaces, `MessageNotifierInterface` (for Messages) and `TaskProcessorInterface` (for Tasks).
+Although in concept one-way notification can be done asynchronously (including delaying it through a queue), in practice few explicit implementations of that model exist.  That provides fewer places from which to draw guidance on details such as proper error handling.  After much consideration the Working Group elected to not provide an explicitly separate workflow for one-way notification as it could be adequately represented as a degenerate case of the others.
 
-### 4.2 Intended use cases
+### 4.2 Example applications
```
Crell: I'm not entirely happy with this list where it is, but it was in the spec previously to differentiate tasks from messages and I didn't want to just lose it.  I'm open to suggestions on where to move it or how to modify it.

proposed/event-dispatcher-meta.md
```diff
@@ -7,7 +7,7 @@ The purpose of this document is to describe the rationale and logic behind the E
 
 ## 2. Why Bother?
 
-Many libraries, components, and frameworks have long supported mechanisms for allowing arbitrary third party code to interact with them.  Most are variations on the classic Observer pattern, often mediated through an intermediary object or service.  Others take a more Aspect-Oriented Programming (AOP) approach.  Nonetheless all have the same basic concept: interupt program flow at a fixed point to provide information to arbitrary third party libraries with information about the action being performed and allow them to either react or influence the program behavior.
+Many libraries, components, and frameworks have long supported mechanisms for allowing arbitrary third party code to interact with them.  Most are variations on the classic Observer pattern, often mediated through an intermediary object or service.  Others take a more Aspect-Oriented Programming (AOP) approach.  Nonetheless all have the same basic concept: interrupt program flow at a fixed point to provide information to arbitrary third party libraries with information about the action being performed and allow them to either react or influence the program behavior.
```
weierophinney: What changed in this paragraph? I've read through it twice, and I cannot for the life of me figure out what it might have been.

Suggestion: can we wrap lines at 80 characters, please? Doing so makes identifying changes in paragraphs simpler, as only the lines with changes will be marked in the diff, not the entire paragraph.

proposed/event-dispatcher-meta.md
```diff
@@ -37,85 +37,84 @@ The Working Group identified four possible workflows for event passing, based on
 * Collection.  ("Give me all your things, that I may do something with that list.")
 * Alternative chain.  ("Here's a thing; the first one of you that can handle it do so, then stop.")
 
-On further review, it was determined that Collection was a special case of Object enhancement (the collection being the object that is enhanced).  Alternative chain is similarly a special case of Object enhancement, as the signature is identical and the dispatch workflow is nearly identical.  That leaves two relevant workflows:
+On further review, it was determined that:
```
weierophinney: This is a passive sentence. I suggest rephrasing it to: "On further review, the working group determined that:"

proposed/event-dispatcher-meta.md
```diff
@@ -37,85 +37,84 @@ The Working Group identified four possible workflows for event passing, based on
...
-* The Emitter wishes to allow a listener to terminate the pipeline early before other listeners have completed.
-
-If uncertain which is appropriate, the Task Processor offers more functionality and is therefore the safer default.
+* Indicating that some change in system configuration or some user action has occurred and allowing other systems to react in ways that do not affect program flow (such as sending an email or logging the action).
+* Passing an object to a series of Listeners to allow it to be modified before it is saved to a persistence system.
+* Passing a collection to a series of Listeners to allow them to register values with it so that the Emitter may act on all of the collected information.
+* Passing a collection to a series of Listeners to allow them to modify the collection in some way before the Emitter takes action.
+* Passing some contextual information to a series of Listeners so that all of them may "vote" on what action to take, with the Emitter deciding based on the aggregate information provided.
+* Passing an object to a series of listeners and allowing one of them to set a value and then prevent further listeners from running.
```
weierophinney: The phrasing in the original was better here. Might I suggest: "Passing an object to a series of listeners, and allowing any listener to terminate early before other listeners have completed."

proposed/event-dispatcher-meta.md
```diff
@@ -37,85 +37,84 @@ The Working Group identified four possible workflows for event passing, based on
...
+However, Stoppable Events (the alternative chain case) also needed to have a channel by which to indicate that further listeners should not be called.  That could be done either by:
 
-* Modifying the Task (`stopPropagation()`)
+* Modifying the Event (e.g., calling a `stopPropagation()` method)
 * Returning a sentinel value from the listener (`true` or `false`) to indicate that propagation should terminate.
-* Evolving the Task to be stopped (`withPropagationStopped()`)
+* Evolving the Event to be stopped (`withPropagationStopped()`)
 
-Of those, the first would mandate a mutable Task in at least some cases.  The second would mandate a mutable Task as the return value was already in use.  And the third seemed unnecessarily ceremonial and pedantic.
+Of those, the first would mandate a mutable Event in at least some cases.  The second would mandate a mutable Event as the return value was already in use.  And the third seemed unnecessarily ceremonial and pedantic.
```
weierophinney: This paragraph and the above points seem to be missing a case: Allowing the `isPropagationStopped()` to return a value based on other state in the event. That particular case essentially requires either mutability or evolvability to provide state changes.

weierophinney: I think you need to clarify "unnecessarily ceremonial and pedantic": _how_ and _why_ does the WG come to that conclusion? (The phrase is quite subjective, so we need to provide our context.)

proposed/event-dispatcher-meta.md
```diff
@@ -37,85 +37,84 @@ The Working Group identified four possible workflows for event passing, based on
...
+Having listeners return evolvable events also posed a challenge.  That pattern is not used by any known implementations in PHP or elsewhere.  It also relies on the listener to remember to return the Event (extra work for the listener author) and to not return some other, new object that might not be fully compatible with later listeners (such as a subclass or superclass of the Event).
 
 Immutable events also rely on the event definer to respect the admonition to be immutable.  Events are, by nature, very loosely designed and the potential for implementers to ignore that part of the spec, even inadvertently, is high.
 
 That left two possible options:
 
-* Allow tasks to be mutable.
+* Allow events to be mutable.
 * Require, but be unable to enforce, immutable objects with a high-ceremony interface, more work for listener authors, and a higher potential for breakage that may not be detectable at compile time.
```
weierophinney: Again, "high-ceremony" is a subjective term, and needs qualification.

proposed/event-dispatcher-meta.md
```diff
@@ -37,85 +37,84 @@ The Working Group identified four possible workflows for event passing, based on
...
 
-Given those options the Working Group felt mutable tasks were the safer alternative.
+Given those options the Working Group felt mutable events were the safer alternative.
 
-The Message Notification use case, however, has no need for return communication so those issues are moot.  Those objects therefore MUST be treated as immutable, regardless of what the object's methods are.
+That said, there is no requirement that an Event be mutable.  Implementers should provide mutator methods on an Event object if and only if it is necessary and appropriate to the use case at hand.
```
weierophinney: I would make this paragraph bold, as I think it needs emphasis. There is a vocal contingent of evaluators who feel immutability is required; we can point them to this section to note that immutability is not _against the specification_, but simply _not required_.

proposed/event-dispatcher-meta.md
```diff
@@ -37,85 +37,84 @@ The Working Group identified four possible workflows for event passing, based on
...
-* A listener could be registered explicitly; it could be the registered explicitly based on reflection of its signature;
+* could be registered explicitly;
+* it could be the registered explicitly based on reflection of its signature;
 * it could be registered with a numeric priority order;
 * it could be registered using a before/after mechanism to control ordering more precisely;
 * it could be registered from a service container;
 * it could use a pre-compile step to generate code;
-* it could be based on method names on objects in the event itself.
+* it could be based on method names on objects in the event itself;
+* it could be limited to certain situations or contexts based on arbitrarily complex logic (only for certain users, only on certain days, only if certain system settings are present, etc).
```
weierophinney: The subject of each of these bullet points is "A listener"; remove the "it" pre-amble to each, as is done correctly in the first point.

proposed/event-dispatcher-meta.md
```diff
@@ -37,85 +37,84 @@ The Working Group identified four possible workflows for event passing, based on
...
 
-While combining the Notifier, Processor, and Provider into a single object is a valid and permissible degenerate case, it is NOT RECOMMENDED as it reduces the flexibility of system integrators.  Instead, the provider should be composed as a dependent object.
+It is even possible, and potentially advisable, to allow libraries to include their own Providers that get aggregated into a common provider that aggregates their listeners to return to the Dispatcher.  That is one possible way to handle arbitrary listener registration within an arbitrary framework, although the Working Group is clear that is not the only option.
+
+While combining the Dispatcher and Provider into a single object is a valid and permissible degenerate case, it is NOT RECOMMENDED as it reduces the flexibility of system integrators.  Instead, the provider should be composed as a dependent object.
```
weierophinney: `s/should/SHOULD/` here.

proposed/event-dispatcher.md
```diff
@@ -66,54 +54,28 @@ A Listener MUST have a `void` return, and SHOULD type hint that return explicitl
...
-* Passing an object to a series of listeners and allowing one of them to set a value and then prevent further listeners from running.
-
-Tasks passed to a Processor SHOULD have some sort of mutator methods on them to allow Listeners to modify the object.  The nature of those methods is up to each implementation to determine.
-
-A Processor
+A Dispatcher
 
 * MUST call Listeners synchronously in the order they are returned from a ListenerProvider.
-* MUST return the Task it was passed after it is done invoking Listeners.
+* MUST return the same Event object it was passed after it is done invoking Listeners.
```
weierophinney: Question: should this be the same event instance passed to the dispatcher, or the same instance the dispatcher passed to listeners? The latter would free up a scenario such as seen in symfony/messenger whereby the messenger may wrap the message in an envelope prior to dispatching it. If it then returned the envelope, it would be in violation of the spec per the above, but if we changed the wording as I suggest, it would still be compliant.

proposed/event-dispatcher.md
```diff
@@ -66,54 +54,28 @@ A Listener MUST have a `void` return, and SHOULD type hint that return explicitl
...
 * MUST NOT return to the Emitter until all Listeners have executed.
-* As an exception to the previous point, if the Task is a Promise then the the Dispatcher MAY return that Promise before all Listeners have executed.  However, the Promise MUST NOT be treated as fulfilled until all Listeners have executed.
+* As an exception to the previous point, if the Event is a Promise object then the the Dispatcher MAY return that Promise before all Listeners have executed.  However, the Promise MUST NOT be treated as fulfilled until all Listeners have executed.
```
weierophinney: Why is this even in the spec? `Promise` is not in the list of definitions, and we've not spoken about anything async up to this point.

I tend to agree that this is how the system should operate in such a scenario, but I don't recall having this conversation at any point.

proposed/event-dispatcher.md
```diff
@@ -124,145 +86,94 @@ A Listener Provider is a service object responsible for determining what Listene
...
 /**
- * Defines a notifier for message events.
+ * Defines a dispatcher for events.
  */
-interface MessageNotifierInterface
+interface EventDispatcherInterface
 {
     /**
-     * Notify listeners of a message event.
+     * Provide all listeners with an event to process.
```
weierophinney: `s/Provide all/Provide all relevant/`, as not _every_ listener will be notified, only those capable of processing the event.

proposed/event-dispatcher.md
```diff
@@ -124,145 +86,94 @@ A Listener Provider is a service object responsible for determining what Listene
...
-     * This method MAY act asynchronously.  Callers SHOULD NOT
-     * assume that any action has been taken when this method
-     * returns.
+     * @param object $event
+     *  The object to process.
      *
-     * @param MessageInterface $event
-     *   The event to notify listeners of.
+     * @return object
+     *  The Event that was passed, now modified by listeners.
```
weierophinney: Here you indent by one space, below you indent by two. Pick one. (Personally, we use a full indentation - 4 spaces - in such scenarios. Main thing is consistency.)

proposed/event-dispatcher-meta.md
bmack: At this point we put in words like "Emitter" which hasn't been introduced yet. I think going with "Dispatcher" might be easier.

proposed/event-dispatcher-meta.md
```diff
@@ -37,85 +37,84 @@ The Working Group identified four possible workflows for event passing, based on
...
+Of those, the first would mandate a mutable Event in at least some cases.  The second would mandate a mutable Event as the return value was already in use.  And the third seemed unnecessarily ceremonial and pedantic.
 
-Having listeners return evolvable tasks also posed a challenge.  That pattern is not used by any known implementations in PHP or elsewhere.  It also relies on the listener to remember to return the object (extra work for the listener author) and to not return some other, new object that might not be fully compatible with later listeners (such as a subclass or superclass of the event).
+Having listeners return evolvable events also posed a challenge.  That pattern is not used by any known implementations in PHP or elsewhere.  It also relies on the listener to remember to return the Event (extra work for the listener author) and to not return some other, new object that might not be fully compatible with later listeners (such as a subclass or superclass of the Event).
```
bmack: extra work = additional work, sounds a bit more elegant IMHO.

proposed/event-dispatcher-meta.md
bmack: "That said, there is no requirement that an Event be mutable" => maybe I'm just non-native, but shouldn't we say "that an Event may be mutable"?

proposed/event-dispatcher.md
bmack: That is something we should discuss IMHO, as it is unclear in my skunk-work as well, what is the proper return value.

proposed/event-dispatcher-meta.md
Crell: Um.  Probably whitespace is my guess?  Not sure.  And hard-wrapping the text makes editing and updating it a huge PITA, because every edit requires reflowing every line.  That's why I try to avoid it.

proposed/event-dispatcher-meta.md
Crell: Updated.

proposed/event-dispatcher-meta.md
Crell: "Emitter" is defined in the spec itself to mean "The thing that creates an event and passes it to the dispatcher".  Dispatcher being "The thing that implements DispatcherInterface".

Updated per MWOP, although I included "the process" as the thing being terminated.

proposed/event-dispatcher.md
Crell: I would favor same object, period.  As in:

```php
$e = new Event();
$e2 = $d->dispatch($e);
assert($e === $e2);
```

Returning a wrapping envelope object means I have to either known how to unwrap it, or I need to know what its signature is.  And if that's not consistent between all implementations I now have an implementation dependency.

IMO the Symfony Message component is still "out of scope" as just not a thing we're dealing with, period.

proposed/event-dispatcher.md
Crell: It's something we discussed very early on, last spring.  I'm pretty sure it would be in the notes doc somewhere.  It was included in order to improve compatibility with async systems without addressing async directly.

proposed/event-dispatcher.md
Crell: Dur, I'll have to change that in the package, too.  Need to leave a note to myself somewhere.

proposed/event-dispatcher.md
Crell: Standardized on 2.  Need to propagate to the package as well.  Mussemfussem...

proposed/event-dispatcher-meta.md
Crell: Done.

proposed/event-dispatcher-meta.md
```diff
@@ -37,85 +37,86 @@ The Working Group identified four possible workflows for event passing, based on
...
-However, Stoppable Task (the alternative chain case) also needed to have a channel by which to indicate that further listeners should not be called.  That could be done either by:
+However, Stoppable Events (the alternative chain case) also needed to have a channel by which to indicate that further listeners should not be called.  That could be done either by:
 
-* Modifying the Task (`stopPropagation()`)
+* Modifying the Event (e.g., calling a `stopPropagation()` method)
 * Returning a sentinel value from the listener (`true` or `false`) to indicate that propagation should terminate.
-* Evolving the Task to be stopped (`withPropagationStopped()`)
+* Evolving the Event to be stopped (`withPropagationStopped()`)
+
+Of those alternatives, therefore, the first would mandate a mutable Event in at least some cases.  The second would mandate a mutable Event as the return value was already in use.  And the third seemed unnecessarily ceremonial and pedantic as it would entail additional syntax requirements for developers with little added value.
```
weierophinney: Why does the second mandate a mutable event? And how is the return value already in use - we currently _ignore_ return values from listeners (which is where the sentinel value would be returned from)?

I suggest the following for the above paragraph:

> Each of these alternatives have drawbacks. The first means that, at least for the purposes of indicating propagation status, events must be mutable. The second requires that listeners return a value, at least when they intend to halt event propagation; this could have ramifications with existing libraries, and potential issues in terms of documentation. The third requires that listeners return the event or mutated event in _all_ cases, and would require dispatchers to test to ensure that the returned value is of the same type as the value passed to the listener; it effectively puts an onus both on consumers and implementers, and thus raising more potential integration issues.

proposed/event-dispatcher-meta.md
weierophinney: Let's try this:

> Require, but be unable to enforce, immutable objects. This leads to increased syntax for both event and listener authors, and a higher potential for breakage that may not be detectable at compile time.

proposed/event-dispatcher.md
weierophinney: I tend to agree with @Crell here - he's saying essentially the same thing I mentioned when we discussed envelopes before, which is that if they are needed, they should decorate the message _before_ you pass it to the dispatcher. That way you don't have to test what was returned; you already know what it is. It also means if you call methods on that object after dispatching the event, you know you're calling methods on the same object sent to listeners.

As such, I'll retract my question and suggestion. :smile:

proposed/event-dispatcher.md
weierophinney: When I said, "we've not spoken about anything async up to this point," I'm talking about the specification itself. We definitely discussed it — but my recollection is we decided to just NOT address it AT ALL, as that allows implementers to decide whether or not they can handle async events/listeners/etc. and still comply with the spec.

My argument here is that it's a new addition to the spec, and my recollection is that as a WG, we decided it wasn't necessary to discuss it within the spec.

Crell: It's not a new addition.  It's been in the spec since last spring, as evidenced by the fact that the line is being modified in place, not added. :smile: The only change in this line is s/Task/Event/ and adding the word "object" for linguistic clarity.

Revisting the existence of that line is out of scope for this PR, IMO.

proposed/event-dispatcher-meta.md
```diff
@@ -37,85 +37,86 @@ The Working Group identified four possible workflows for event passing, based on
...
-
-* The event is being used as a "hook" or "pointcut" to extend or modify the Emitter's behavior.
-* The Emitter wishes to allow listeners to interact with each other (through mutating the event).
-* The Emitter wishes to allow a listener to terminate the pipeline early before other listeners have completed.
-
-If uncertain which is appropriate, the Task Processor offers more functionality and is therefore the safer default.
+* Indicating that some change in system configuration or some user action has occurred and allowing other systems to react in ways that do not affect program flow (such as sending an email or logging the action).
+* Passing an object to a series of Listeners to allow it to be modified before it is saved to a persistence system.
+* Passing a collection to a series of Listeners to allow them to register values with it so that the Emitter may act on all of the collected information.
+* Passing a collection to a series of Listeners to allow them to modify the collection in some way before the Emitter takes action.
```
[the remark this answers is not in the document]
Crell: Well, then it becomes essentially the same as the second point, no?  I'll see if I can wordsmith it a bit better.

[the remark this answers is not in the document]
Crell: I ended up just collapsing both lines into one.  It works well enough.

proposed/event-dispatcher.md
weierophinney: I saw the `+` and immediately thought "new addition". My bad!

Let's revisit after this PR is done.

PR #1130, opened 2019-01-01
Formatting fix

Big-Shark:


Crell:
I'm... not sure what the fix is here?  The diff looks like a strange noop unless I'm missing something.

That said, this would conflict with #1128, which is gutting that section anyway.  So thanks, but not applicable at the moment.

PR #1131, opened 2019-01-08
Add statement about compound statements

jaapio:
Compound statements can lead to very unclear behavior in IDE's
and api-doc generators. Thus they should be avoided.

### DIFF
--- a/proposed/phpdoc.md
+++ b/proposed/phpdoc.md
@@ -109,7 +109,7 @@ interpreted as described in [RFC 2119][RFC2119].
   }
   ```
 
-  It is NOT RECOMMENDED to use compound definitions for Constants or Properties. Since the
+  It is NOT RECOMMENDED to use compound definitions for Constants or Properties, since the
   handling of DocBlocks in these situations can lead to unexpected results. If compound statement is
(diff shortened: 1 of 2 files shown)


proposed/phpdoc.md
```diff
@@ -109,6 +109,27 @@ interpreted as described in [RFC 2119][RFC2119].
   }
   ```
 
+  It is NOT RECOMMENDED to use compound definitions for Constants or Properties. Since the
```
Jean85:
```suggestion
  It is NOT RECOMMENDED to use compound definitions for Constants or Properties, since the
```

jaapio:
PSR-12 was not there when this was written. But still this is true, if you don't apply PRS-12 to your code this is applicable. A project can choose which PSR's it follows, so no reason IMO to scratch this.

ashnazg:
I think I'm good with going either direction... add this verbiage, or drop all mentions of compound statements based on a PSR-12 adherence assumption.  In the interest of handling the PR, we can merge it now, and later discussion can drive the "drop all mentions" thoughts.

ashnazg (approved):

PR #1132, opened 2019-01-09
PSR-14: Provide examples of "high-ceremony" usage

weierophinney:
Per discussion within the PSR-14 working group, this patch provides an explanation of what is meant by the phrase "high-ceremony", along with implementation examples and narrative that demonstrate the concept in the specific context of stopping event propagation.

### DIFF
--- a/proposed/event-dispatcher-meta.md
+++ b/proposed/event-dispatcher-meta.md
@@ -98,6 +98,23 @@ foreach ($provider->getListenersForEvent($event) as $listener) {
     
     if (! $returnedEvent instanceof $event) {
         // This is an exceptional case!
+        // 
+        // We now have an event of a different type, or perhaps nothing was
+        // returned by the listener. An event of a different type might mean:
+        // 
+        // - we need to trigger the new event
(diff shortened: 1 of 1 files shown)


proposed/event-dispatcher-meta.md
```diff
@@ -80,7 +80,37 @@ That left two possible options:
...
+
+In the latter case, dispatcher implementations, checks on the return value would be required:
+
+```php
+foreach ($provider->getListenersForEvent($event) as $listener) {
+    $returnedEvent = $listener($event);
+    
+    if (! $returnedEvent instanceof $event) {
+        // This is an exceptional case!
+    }
```
Crell: The comment here should be expanded, as it's a key point.  The dispatcher has to handle this case because a buggy listener may not return an event, or the same event, or anything at all, all of which could break things in exciting ways.  Not relying on the return value of the listener removes a whole array of error conditions that no longer need to be addressed.

weierophinney: I've pushed a change that should address this more fully.

PR #1133, opened 2019-01-09
Psr 14 tweaks

Crell:
Per discussion in the WG:

* Add a link to define what a Promise is.
* Check propagation status before calling listeners.
* Better explain how flexible secondary processing is.

### DIFF
--- a/proposed/event-dispatcher.md
+++ b/proposed/event-dispatcher.md
@@ -69,7 +69,7 @@ A Dispatcher
 
 If passed a Stoppable Event, a Dispatcher
 
-* MUST call `isPropagationStopped()` on the Event after each Listener has been called.  If that method returns `true` it MUST return the Event to the Emitter immediately and MUST NOT call any further Listeners.
+* MUST call `isPropagationStopped()` on the Event before each Listener has been called.  If that method returns `true` it MUST return the Event to the Emitter immediately and MUST NOT call any further Listeners.  That is, if an Event is passed to the Dispatcher that always returns `true` from `isPropagationStopped()`, zero listeners will be called.
 
 ### Error handling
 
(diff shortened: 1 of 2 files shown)


proposed/event-dispatcher.md
```diff
@@ -65,11 +67,11 @@ A Dispatcher
 * MUST call Listeners synchronously in the order they are returned from a ListenerProvider.
 * MUST return the same Event object it was passed after it is done invoking Listeners.
 * MUST NOT return to the Emitter until all Listeners have executed.
-* As an exception to the previous point, if the Event is a Promise object then the the Dispatcher MAY return that Promise before all Listeners have executed.  However, the Promise MUST NOT be treated as fulfilled until all Listeners have executed.
+* As an exception to the previous point, if the Event is a [Promise object](https://promisesaplus.com/) then the the Dispatcher MAY return that Promise before all Listeners have executed.  However, the Promise MUST NOT be treated as fulfilled until all Listeners have executed.
```
weierophinney: `s/then the the/then the/`

proposed/event-dispatcher.md
```diff
@@ -56,6 +56,8 @@ A Listener MAY delegate actions to other code.  That includes a Listener being a
 
 A Listener MAY enqueue information from the Event for later processing by a secondary process, using cron, a queue server, or similar techniques.  It MAY serialize the Event object itself to do so, however, care should be taken that not all Event objects may be safely serializable.
 
+A secondary process MUST assume that any changes it makes to an Event object will NOT propagate to other Listeners.  Additionally, the `isPropagationStopped()` method will not be called in this case and thus the Event is effectively unstoppable.
```
weierophinney: What is implied by a "secondary process" here? I'm missing context, and don't quite understand what this paragraph is specifying. Are we talking a queue runner that has received a deferred event and listener here? If so, this needs to be part of the previous paragraph to ensure that context is retained.

Crell: Per the previous paragraph, "Secondary process" means cron, a queue server, or "similar techniques" (ie, "something like that").  So whatever mechanism is being used for "later", it loses its bidirectional communication mechanisms.

weierophinney: Then, as I noted, I think the two paragraphs need to be combined. Each of the previous paragraphs is detailing a single scenario, so when I read this one, I assumed it was separate from the previous paragraph, and thus I had questions about context. Combining it with the previous paragraph would make this eminently clear.

PR #1136, opened 2019-01-14
Revise wording around event identification.

Crell:


proposed/event-dispatcher.md
```diff
@@ -32,7 +32,7 @@ Some examples:
 
 Events are objects that act as the unit of communication between an Emitter and appropriate Listeners.
 
-All Events are identified primarily by their PHP type: i.e., their class and any interfaces they implement.  Events SHOULD NOT have any other identifier, such as an arbitrary string ID.
+The class name of the Event object SHOULD be the primary identifier of the Event to differentiate it from other Events.  Additionally, implementers SHOULD include an event's parent classes or interfaces as identifying the Event when determining the appropriate way to handle it.
```
weierophinney: Only comment I have is that `event's` should be `Events'`. Otherwise, this is what we have discussed previously.

PR #1137, opened 2019-01-15
Final passthrough on PSR-14

weierophinney:
This is my final passthrough on PSR-14 before we prepare our review readiness vote. In particular, I looked for the following:

- consistent casing of defined terms. In a few cases, they were left lower-case when it was clear they were referring to a broader definition.
- removal of "that is" phrases. These are a weird idiom, and better phrasing exists in almost every case, even if just dropping to the latin `i.e.`.
- rephrasing where the language was actively getting in the way of understanding.

I plan to do a review of the patch to detail some of the decisions before asking for a review by a WG member.

### DIFF
--- a/proposed/event-dispatcher-meta.md
+++ b/proposed/event-dispatcher-meta.md
@@ -3,7 +3,7 @@ Event Dispatcher Meta Document
 
 ## 1. Summary
 
-The purpose of this document is to describe the rationale and logic behind the Event Dispatcher PSR.
+The purpose of this document is to describe the rationale and logic behind the Event Dispatcher specification.
 
 ## 2. Why Bother?
 
@@ -16,15 +16,15 @@ This is a well-established model, but a standard mechanism by which libraries do
 ### 3.1 Goals
 
 * Simplify and standardize the process by which libraries and components may expose themselves to extension via "events" so that they may be more easily incorporated into applications and frameworks.
-* Simplify and standardize the process by which libraries and components may register an interest in responding to an event so that they may be more easily incorporated into arbitrary applications and frameworks.
+* Simplify and standardize the process by which libraries and components may register an interest in responding to an Event so that they may be more easily incorporated into arbitrary applications and frameworks.
 * To the extent feasible, ease the process for existing code bases to transition toward this specification.
 
 ### 3.2 Non-Goals
 
 * Asynchronous systems often have a concept of an "event loop" to manage interleaving coroutines.  That is an unrelated matter and explicitly irrelevant to this specification.
-* Storage systems implementing an "Event Source" pattern also have a concept of an "event".  That is unrelated to the events discussed here and explicitly out of scope.
+* Storage systems implementing an "Event Source" pattern also have a concept of an "event".  That is unrelated to the Events discussed here and explicitly out of scope.
 * Strict backward compatibility with existing event systems is not a priority and is not expected.
-* While this specification will undoubtedly suggest implementation patterns, it does not seek to define One True Event Dispatcher Implementation, only how callers and listeners communicate with that dispatcher.
+* While this specification will undoubtedly suggest implementation patterns, it does not seek to define One True Event Dispatcher Implementation, only how callers and Listeners communicate with that Dispatcher.
 
 ## 4. Approaches
 
@@ -39,11 +39,11 @@ The Working Group identified four possible workflows for event passing, based on
 
 On further review, the Working Goup determined that:
 
-* Collection was a special case of Object enhancement (the collection being the object that is enhanced)
-* Alternative chain is similarly a special case of Object enhancement, as the signature is identical and the dispatch workflow is nearly identical, albeit with an extra check included.
+* Collection was a special case of object enhancement (the collection being the object that is enhanced).
+* Alternative chain is similarly a special case of object enhancement, as the signature is identical and the dispatch workflow is nearly identical, albeit with an extra check included.
 * One-way notification is a degenerate case of the others, or can be represented as such.
 
-Although in concept one-way notification can be done asynchronously (including delaying it through a queue), in practice few explicit implementations of that model exist.  That provides fewer places from which to draw guidance on details such as proper error handling.  After much consideration the Working Group elected to not provide an explicitly separate workflow for one-way notification as it could be adequately represented as a degenerate case of the others.
+Although in concept one-way notification can be done asynchronously (including delaying it through a queue), in practice, few explicit implementations of that model exist, providing fewer places from which to draw guidance on details (such as proper error handling).  After much consideration, the Working Group elected not to provide an explicitly separate workflow for one-way notification as it could be adequately represented as a degenerate case of the others.
 
 ### 4.2 Example applications
 
@@ -51,36 +51,36 @@ Although in concept one-way notification can be done asynchronously (including d
 * Passing an object to a series of Listeners to allow it to be modified before it is saved to a persistence system.
 * Passing a collection to a series of Listeners to allow them to register values with it or modify existing values so that the Emitter may act on all of the collected information.
 * Passing some contextual information to a series of Listeners so that all of them may "vote" on what action to take, with the Emitter deciding based on the aggregate information provided.
-* Passing an object to a series of listeners and allowing any listener to terminate the process early before other listeners have completed.
+* Passing an object to a series of Listeners and allowing any Listener to terminate the process early before other Listeners have completed.
 
 ### 4.3 Immutable events
 
-Initially the Working Group wished to define all Events as immutable message objects, similar to PSR-7.  However, that proved problematic in all but the one-way notification case.  In the other scenarios Listeners needed a way to return data to the caller.  In concept there were three possible avenues:
+Initially the Working Group wished to define all Events as immutable message objects, similar to PSR-7.  However, that proved problematic in all but the one-way notification case.  In the other scenarios, Listeners needed a way to return data to the caller.  In concept, there were three possible avenues:
 
-* Make the event mutable and modify it in place.
-* Require that events be evolvable (immutable but with `with*()` methods like PSR-7 and PSR-13) and that listeners return the event to pass along.
-* Make the Event immutable but aggregate and return the return value from each Listener.
+* Make the Event mutable and modify it in place.
+* Require that Events be evolvable (immutable but with `with*()` methods like PSR-7 and PSR-13) and that Listeners return the Event to pass along.
+* Make the Event immutable, but aggregate and return the return value from each Listener.
 
-However, Stoppable Events (the alternative chain case) also needed to have a channel by which to indicate that further listeners should not be called.  That could be done either by:
+However, Stoppable Events (the alternative chain case) also needed to have a channel by which to indicate that further Listeners should not be called.  That could be done either by:
 
 * Modifying the Event (e.g., calling a `stopPropagation()` method)
-* Returning a sentinel value from the listener (`true` or `false`) to indicate that propagation should terminate.
+* Returning a sentinel value from the Listener (`true` or `false`) to indicate that propagation should terminate.
 * Evolving the Event to be stopped (`withPropagationStopped()`)
 
-Each of these alternatives have drawbacks. The first means that, at least for the purposes of indicating propagation status, events must be mutable. The second requires that listeners return a value, at least when they intend to halt event propagation; this could have ramifications with existing libraries, and potential issues in terms of documentation. The third requires that listeners return the event or mutated event in all cases, and would require dispatchers to test to ensure that the returned value is of the same type as the value passed to the listener; it effectively puts an onus both on consumers and implementers, and thus raising more potential integration issues.
+Each of these alternatives have drawbacks. The first means that, at least for the purposes of indicating propagation status, Events must be mutable. The second requires that Listeners return a value, at least when they intend to halt event propagation; this could have ramifications with existing libraries, and potential issues in terms of documentation. The third requires that Listeners return the Event or mutated Event in all cases, and would require Dispatchers to test to ensure that the returned value is of the same type as the value passed to the Listener; it effectively puts an onus both on consumers and implementers, thus raising more potential integration issues.
 
-Additionally, a desired feature was the ability to derive whether or not to stop propagation based on some values collected from the listeners.  (For example, to stop when one of them has provided a certain value, or after at least three of them have indicated a "reject this request" flag, or similar.)  While technically possible to implement as an evolvable object, such behavior is intrinsically stateful so would be highly cumbersome for both implementers and users.
+Additionally, a desired feature was the ability to derive whether or not to stop propagation based on values collected from the Listeners.  (For example, to stop when one of them has provided a certain value, or after at least three of them have indicated a "reject this request" flag, or similar.)  While technically possible to implement as an evolvable object, such behavior is intrinsically stateful, so would be highly cumbersome for both implementers and users.
 
-Having listeners return evolvable events also posed a challenge.  That pattern is not used by any known implementations in PHP or elsewhere.  It also relies on the listener to remember to return the Event (additional work for the listener author) and to not return some other, new object that might not be fully compatible with later listeners (such as a subclass or superclass of the Event).
+Having Listeners return evolvable Events also posed a challenge.  That pattern is not used by any known implementations in PHP or elsewhere.  It also relies on the Listener to remember to return the Event (additional work for the Listener author) and to not return some other, new object that might not be fully compatible with later Listeners (such as a subclass or superclass of the Event).
 
-Immutable events also rely on the event definer to respect the admonition to be immutable.  Events are, by nature, very loosely designed and the potential for implementers to ignore that part of the spec, even inadvertently, is high.
+Immutable Events also rely on the Event author to respect the admonition to be immutable.  Events are, by nature, very loosely designed, and the potential for implementers to ignore that part of the spec, even inadvertently, is high.
 
 That left two possible options:
 
-* Allow events to be mutable.
-* Require, but be unable to enforce, immutable objects with a high-ceremony interface, more work for listener authors, and a higher potential for breakage that may not be detectable at compile time.
+* Allow Events to be mutable.
+* Require, but be unable to enforce, immutable Events with a high-ceremony interface, more work for Listener authors, and a higher potential for breakage that may not be detectable at compile time.
 
-By "high-ceremony", we imply that verbose syntax and/or implementations would be required.  In the former case, listener authors would need to (a) create a new instance with the propagation flag toggled, and (b) return the new instance so that the dispatcher could examine it:
+By "high-ceremony", we imply that verbose syntax and/or implementations would be required.  In the former case, Listener authors would need to (a) create a new Event instance with the propagation flag toggled, and (b) return the new Event instance so that the Dispatcher could examine it:
 
 ```php
 function (SomeEvent $event) : SomeEvent
@@ -90,7 +90,7 @@ function (SomeEvent $event) : SomeEvent
 }
 ```
 
-In the latter case, dispatcher implementations, checks on the return value would be required:
+The latter case, Dispatcher implementations, would require checks on the return value:
 
 ```php
 foreach ($provider->getListenersForEvent($event) as $listener) {
@@ -127,13 +127,13 @@ foreach ($provider->getListenersForEvent($event) as $listener) {
 
 In both situations, we would be introducing more potential edge cases, with little benefit, and few language-level mechanisms to guide developers to correct implementation.
 
-Given these options the Working Group felt mutable events were the safer alternative.
+Given these options, the Working Group felt mutable Events were the safer alternative.
 
 That said, *there is no requirement that an Event be mutable*.  Implementers should provide mutator methods on an Event object *if and only if it is necessary* and appropriate to the use case at hand.
 
 ### 4.4 Listener registration
 
-Experimentation during development of the specification determined that there were a wide range of viable, legitimate means by which a Dispatcher could be informed of a listener.  A listener:
+Experimentation during development of the specification determined that there were a wide range of viable, legitimate means by which a Dispatcher could be informed of a Listener.  A Listener:
 
 * could be registered explicitly;
 * could be the registered explicitly based on reflection of its signature;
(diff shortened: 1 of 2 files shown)


proposed/event-dispatcher-meta.md
```diff
@@ -3,7 +3,7 @@ Event Dispatcher Meta Document
 
 ## 1. Summary
 
-The purpose of this document is to describe the rationale and logic behind the Event Dispatcher PSR.
+The purpose of this document is to describe the rationale and logic behind the Event Dispatcher specification.
```
weierophinney: While I know we are defining PSRs, they are more generally _specifications_, and using that verbiage here makes the document more approachable for those not familiar with the PSR system in my opinion.

proposed/event-dispatcher-meta.md
```diff
@@ -39,48 +39,48 @@ The Working Group identified four possible workflows for event passing, based on
 
 On further review, the Working Goup determined that:
 
-* Collection was a special case of Object enhancement (the collection being the object that is enhanced)
-* Alternative chain is similarly a special case of Object enhancement, as the signature is identical and the dispatch workflow is nearly identical, albeit with an extra check included.
+* Collection was a special case of object enhancement (the collection being the object that is enhanced).
+* Alternative chain is similarly a special case of object enhancement, as the signature is identical and the dispatch workflow is nearly identical, albeit with an extra check included.
```
weierophinney: The casing of "Object enhancement" was simply a by-product of it being the first word of a bulleted point; it's not a specific vocabulary definition we have provided. As such, I'm recommending here we remove the special casing.

proposed/event-dispatcher-meta.md
```diff
@@ -39,48 +39,48 @@ The Working Group identified four possible workflows for event passing, based on
...
 * One-way notification is a degenerate case of the others, or can be represented as such.
 
-Although in concept one-way notification can be done asynchronously (including delaying it through a queue), in practice few explicit implementations of that model exist.  That provides fewer places from which to draw guidance on details such as proper error handling.  After much consideration the Working Group elected to not provide an explicitly separate workflow for one-way notification as it could be adequately represented as a degenerate case of the others.
+Although in concept one-way notification can be done asynchronously (including delaying it through a queue), in practice, few explicit implementations of that model exist, providing fewer places from which to draw guidance on details (such as proper error handling).  After much consideration, the Working Group elected not to provide an explicitly separate workflow for one-way notification as it could be adequately represented as a degenerate case of the others.
```
weierophinney: Rephrased this to make the sentence more active, as a continuation.

weierophinney: Also, corrected a split infinitive. ;-)

proposed/event-dispatcher-meta.md
```diff
@@ -39,48 +39,48 @@ The Working Group identified four possible workflows for event passing, based on
...
 * Passing some contextual information to a series of Listeners so that all of them may "vote" on what action to take, with the Emitter deciding based on the aggregate information provided.
-* Passing an object to a series of listeners and allowing any listener to terminate the process early before other listeners have completed.
+* Passing an object to a series of Listeners and allowing any Listener to terminate the process early before other Listeners have completed.
 
 ### 4.3 Immutable events
 
-Initially the Working Group wished to define all Events as immutable message objects, similar to PSR-7.  However, that proved problematic in all but the one-way notification case.  In the other scenarios Listeners needed a way to return data to the caller.  In concept there were three possible avenues:
+Initially the Working Group wished to define all Events as immutable message objects, similar to PSR-7.  However, that proved problematic in all but the one-way notification case.  In the other scenarios, Listeners needed a way to return data to the caller.  In concept, there were three possible avenues:
```
weierophinney: Added commas to break the flow more naturally.

proposed/event-dispatcher-meta.md
```diff
@@ -39,48 +39,48 @@ The Working Group identified four possible workflows for event passing, based on
...
+Each of these alternatives have drawbacks. The first means that, at least for the purposes of indicating propagation status, Events must be mutable. The second requires that Listeners return a value, at least when they intend to halt event propagation; this could have ramifications with existing libraries, and potential issues in terms of documentation. The third requires that Listeners return the Event or mutated Event in all cases, and would require Dispatchers to test to ensure that the returned value is of the same type as the value passed to the Listener; it effectively puts an onus both on consumers and implementers, thus raising more potential integration issues.
```
weierophinney: This change is primarily around casing of vocabulary defined terms. The only other change is in the last sentence, where I removed an extraneous "and".

proposed/event-dispatcher-meta.md
```diff
@@ -39,48 +39,48 @@ The Working Group identified four possible workflows for event passing, based on
...
 
-Additionally, a desired feature was the ability to derive whether or not to stop propagation based on some values collected from the listeners.  (For example, to stop when one of them has provided a certain value, or after at least three of them have indicated a "reject this request" flag, or similar.)  While technically possible to implement as an evolvable object, such behavior is intrinsically stateful so would be highly cumbersome for both implementers and users.
+Additionally, a desired feature was the ability to derive whether or not to stop propagation based on values collected from the Listeners.  (For example, to stop when one of them has provided a certain value, or after at least three of them have indicated a "reject this request" flag, or similar.)  While technically possible to implement as an evolvable object, such behavior is intrinsically stateful, so would be highly cumbersome for both implementers and users.
```
weierophinney: In addition to casing of defined terms, I removed the word "some" in the first sentence (it was unnecessary), and added a comma in the last sentence (to break flow more naturally).

proposed/event-dispatcher-meta.md
```diff
@@ -39,48 +39,48 @@ The Working Group identified four possible workflows for event passing, based on
...
+Having Listeners return evolvable Events also posed a challenge.  That pattern is not used by any known implementations in PHP or elsewhere.  It also relies on the Listener to remember to return the Event (additional work for the Listener author) and to not return some other, new object that might not be fully compatible with later Listeners (such as a subclass or superclass of the Event).
 
-Immutable events also rely on the event definer to respect the admonition to be immutable.  Events are, by nature, very loosely designed and the potential for implementers to ignore that part of the spec, even inadvertently, is high.
+Immutable Events also rely on the Event author to respect the admonition to be immutable.  Events are, by nature, very loosely designed, and the potential for implementers to ignore that part of the spec, even inadvertently, is high.
```
weierophinney: "author" read better than "definer" here.

proposed/event-dispatcher-meta.md
```diff
@@ -39,48 +39,48 @@ The Working Group identified four possible workflows for event passing, based on
...
 
 That left two possible options:
 
-* Allow events to be mutable.
-* Require, but be unable to enforce, immutable objects with a high-ceremony interface, more work for listener authors, and a higher potential for breakage that may not be detectable at compile time.
+* Allow Events to be mutable.
+* Require, but be unable to enforce, immutable Events with a high-ceremony interface, more work for Listener authors, and a higher potential for breakage that may not be detectable at compile time.
```
weierophinney: I substituted "Events" for "objects" here, to make it clear that we are discussing Events specifically.

proposed/event-dispatcher-meta.md
```diff
@@ -39,48 +39,48 @@ The Working Group identified four possible workflows for event passing, based on
...
 
-By "high-ceremony", we imply that verbose syntax and/or implementations would be required.  In the former case, listener authors would need to (a) create a new instance with the propagation flag toggled, and (b) return the new instance so that the dispatcher could examine it:
+By "high-ceremony", we imply that verbose syntax and/or implementations would be required.  In the former case, Listener authors would need to (a) create a new Event instance with the propagation flag toggled, and (b) return the new Event instance so that the Dispatcher could examine it:
```
weierophinney: I qualified "instance" with "Event" (i.e., "Event instance") in two places here; again, I did this to clarify that we are discussing Events and not arbitrary objects.

proposed/event-dispatcher-meta.md
```diff
@@ -90,7 +90,7 @@ function (SomeEvent $event) : SomeEvent
 }
 ```
 
-In the latter case, dispatcher implementations, checks on the return value would be required:
+The latter case, Dispatcher implementations, would require checks on the return value:
```
weierophinney: The previous wording was really unwieldy and passive.

proposed/event-dispatcher-meta.md
```diff
@@ -127,38 +127,38 @@ foreach ($provider->getListenersForEvent($event) as $listener) {
...
 
-The Working Group therefore chose to encapsulate the registration of listeners behind the `ListenerProviderInterface`.  A provider object may have an explicit registration mechanism available, or multiple such mechanisms, or none.  It could also be generated code produced by some compile step.  That is up to the implementer.  However, that also splits the responsibility of managing the process of dispatching an event from the process of mapping an event to listeners.  That way different implementations may be mixed-and-matched with different provider mechanisms as needed.
+The Working Group therefore chose to encapsulate the registration of Listeners behind the `ListenerProviderInterface`.  A Provider object may have an explicit registration mechanism available, or multiple such mechanisms, or none.  It could also be generated code produced by some compile step.  However, that also splits the responsibility of managing the process of dispatching an Event from the process of mapping an Event to Listeners.  That way different implementations may be mixed-and-matched with different Provider mechanisms as needed.
```
weierophinney: I removed the sentence "That is up to the implementer." It was redundant; the whole paragraph is about how implementations may vary.

proposed/event-dispatcher-meta.md
```diff
@@ -168,7 +168,7 @@ While technically a side effect of the design, it is essentially the same approa
 
 ### 4.6 Return values
 
-Per the spec, a Dispatcher MUST return the event it was passed to the Emitter.  That is done to provide a more ergonomic experience for users.  That is, it allows short-hands similar to the following:
+Per the spec, a Dispatcher MUST return the Event passed by the Emitter.  This is specified to provide a more ergonomic experience for users, allowing short-hands similar to the following:
```
weierophinney: "That is", "That is"... New version flows more naturally.

proposed/event-dispatcher-meta.md
```diff
@@ -199,12 +199,18 @@ Matthew Weier O'Phinney
 
 ## 6. Votes
 
-* [Entrance vote](https://groups.google.com/d/topic/php-fig/6kQFX-lhuk4/discussion)
+* [Entrance vote][]
```
weierophinney: We should use exactly one style for links everywhere; we had a mix of different styles. I chose for the "annotated" variety, as it's more consistent with other specifications.

proposed/event-dispatcher.md
```diff
@@ -11,75 +11,77 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 ## Goal
 
-Having common interfaces for dispatching and handling events, allows developers to create libraries that can interact with many frameworks and other libraries in a common fashion.
+Having common interfaces for dispatching and handling events allows developers to create libraries that can interact with many frameworks and other libraries in a common fashion.
```
weierophinney: The comma was breaking the sentence unnaturally.

proposed/event-dispatcher.md
```diff
@@ -11,75 +11,77 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
 
 Some examples:
 
 * A security framework that will prevent saving/accessing data when a user doesn't have permission.
 * A common full page caching system.
-* Libraries that extent other specific libraries, regardless of what framework they are both integrated into.
+* Libraries that extend other libraries, regardless of what framework they are both integrated into.
```
weierophinney: - `s/extent/extend/`
- Removes "specific"; it's redundant.

proposed/event-dispatcher.md
```diff
@@ -11,75 +11,77 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
 * A logging package to track all actions taken within the application
 
 ## Definitions
 
 * **Event** - An Event is a message produced by an *Emitter*.  It may be any arbitrary PHP object.
 * **Listener** - A Listener is any PHP callable that expects to be passed an Event.  Zero or more Listeners may be passed the same Event.  A Listener MAY enqueue some other asynchronous behavior if it so chooses.
-* **Emitter** - An Emitter is any arbitrary code that wishes to send an Event.  This is also known as the "calling code".  It is not represented by any particular data structure but refers to the use case.
+* **Emitter** - An Emitter is any arbitrary code that wishes to dispatch an Event.  This is also known as the "calling code".  It is not represented by any particular data structure but refers to the use case.
```
weierophinney: I did a `s/send/dispatch/` here, to have affinity with the `Dispatcher` term we bring up next.

proposed/event-dispatcher.md
```diff
@@ -11,75 +11,77 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
 
-An Event that implements `StoppableEventInterface` MUST return `true` from `isPropagationStopped()` when whatever Event it represents has been completed.  It is up to the implementer of the class to determine when that is.  For example, an Event that is asking for a PSR-7 `RequestInterface` object to be matched with a corresponding `ResponseInterface` object MAY have a `setResponse(ResponseInterface $res)` method for a Listener to call, which sets an internal flag that `isPropagationStopped()` will use to return `true` once the response has been set.
+An Event that implements `StoppableEventInterface` MUST return `true` from `isPropagationStopped()` when whatever Event it represents has been completed.  It is up to the implementer of the class to determine when that is.  For example, an Event that is asking for a PSR-7 `RequestInterface` object to be matched with a corresponding `ResponseInterface` object MAY have a `setResponse(ResponseInterface $res)` method for a Listener to call, which causes `isPropagationStopped()` to return `true`.
```
weierophinney: I rephrased this last sentence significantly. The idea of a flag is an implementation detail of "I have a response; therefor, stop propagation," and trying to shoe-horn that in was making it hard to understand that specific point.

proposed/event-dispatcher.md
```diff
@@ -11,75 +11,77 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
 
 ## Listeners
 
-A Listener may be any PHP callable.  A Listener MUST have one and only one parameter, which is the Event to which it responds.  Listeners SHOULD type hint that parameter as specifically as is relevant for their use case; that is, a Listener MAY type hint against an interface to indicate it is compatible with any Event type that implements that interface.
+A Listener may be any PHP callable.  A Listener MUST have one and only one parameter, which is the Event to which it responds.  Listeners SHOULD type hint that parameter as specifically as is relevant for their use case; that is, a Listener MAY type hint against an interface to indicate it is compatible with any Event type that implements that interface, or to a specific implementation of that interface.
```
weierophinney: I added some verbiage to the end of this sentence, as I think we lost the point that the listener can also respond to a _specific implementation_. I don't want readers to think they MUST create interfaces for their events!

proposed/event-dispatcher.md
```diff
@@ -11,75 +11,77 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
 
-A Listener MUST have a `void` return, and SHOULD type hint that return explicitly.
+A Listener SHOULD have a `void` return, and SHOULD type hint that return explicitly.
```
weierophinney: This is likely the most controversial change I've proposed.

The reason I've changed it from `MUST` to `SHOULD` is because many existing event systems _allow return values_. Some aggregate them, some do not; some ignore them entirely. But if we make this a `MUST`, it may potentially cause some readers to feel that their existing listeners are invalid, forcing them to choose between not implementing the spec, or breaking backwards compatibility.

proposed/event-dispatcher.md
```diff
@@ -11,75 +11,77 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
 
-A Listener MAY delegate actions to other code.  That includes a Listener being a thin wrapper around retrieving an object from a service container that contains the actual business logic to run, or other similar forms of indirection.  In that case the callable containing the actual business logic SHOULD conform to the same rules as if it were called directly as a Listener.
+A Listener MAY delegate actions to other code.  That includes a Listener being a thin wrapper around retrieving an object from a service container that contains the actual business logic to run, or other similar forms of indirection.  In that case, the callable containing the actual business logic SHOULD conform to the same rules as if it were called directly as a Listener.
```
weierophinney: Added a comma to break the flow more naturally.

proposed/event-dispatcher.md
```diff
@@ -11,75 +11,77 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
 
-A Listener MAY enqueue information from the Event for later processing by a secondary process, using cron, a queue server, or similar techniques.  It MAY serialize the Event object itself to do so, however, care should be taken that not all Event objects may be safely serializable. A secondary process MUST assume that any changes it makes to an Event object will NOT propagate to other Listeners.  Additionally, the `isPropagationStopped()` method will not be called in this case and thus the Event is effectively unstoppable.
+A Listener MAY enqueue information from the Event for later processing by a secondary process, using cron, a queue server, or similar techniques.  It MAY serialize the Event object itself to do so; however, care should be taken that not all Event objects may be safely serializable. A secondary process MUST assume that any changes it makes to an Event object will NOT propagate to other Listeners.
```
weierophinney: I dropped the last sentence here.

The reason is: we have zero idea of how _other_ listeners might handle the event, and one of them may stop propagation either before or after a queueing listener operates. All it means is that any modifications made to the event by the secondary process will likely not propagate to other listeners. I do not think that information needs to be spelled out, as it's something that can be inferred.

proposed/event-dispatcher.md
```diff
@@ -11,75 +11,77 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
 
 ## Dispatcher
 
-A Dispatcher is a service object implementing `EventDispatcherInterface`.  It is responsible for invoking listeners provided by a Listener Provider on an Event.
+A Dispatcher is a service object implementing `EventDispatcherInterface`.  It is responsible for invoking listeners supplied by a Listener Provider for the Event dispatched.
```
weierophinney: Previous wording was terrible.

proposed/event-dispatcher.md
```diff
@@ -11,75 +11,77 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
+* As an exception to the previous point, if the Event is a [Promise object][] then the Dispatcher MAY return that Promise before all Listeners have executed.  However, the Promise MUST NOT be treated as fulfilled until all Listeners have executed.
 
 If passed a Stoppable Event, a Dispatcher
 
-* MUST call `isPropagationStopped()` on the Event before each Listener has been called.  If that method returns `true` it MUST return the Event to the Emitter immediately and MUST NOT call any further Listeners.  That is, if an Event is passed to the Dispatcher that always returns `true` from `isPropagationStopped()`, zero listeners will be called.
+* MUST call `isPropagationStopped()` on the Event before each Listener has been called.  If that method returns `true` it MUST return the Event to the Emitter immediately and MUST NOT call any further Listeners.  This implies that if an Event is passed to the Dispatcher that always returns `true` from `isPropagationStopped()`, zero listeners will be called.
```
weierophinney: I rephrased the last sentence to make it more active.

proposed/event-dispatcher.md
```diff
@@ -11,75 +11,77 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
+
+[Promise object]: https://promisesaplus.com/
 
 ### Error handling
 
-An Exception or Error thrown by a Listener MUST block the execution of any further Listeners.  An Error or Exception thrown by a Listener MUST be allowed to propagate back up to the caller.
+An Exception or Error thrown by a Listener MUST block the execution of any further Listeners.  An Error or Exception thrown by a Listener MUST be allowed to propagate back up to the Emitter.
```
weierophinney: The caller is the `Emitter`, so I substituted that term for clarity

proposed/event-dispatcher.md
```diff
@@ -11,75 +11,77 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
 
 A Dispatcher MAY catch a thrown object to log it, allow additional action to be taken, etc., but then MUST rethrow the original throwable.
 
 ## Listener Provider
 
-A Listener Provider is a service object responsible for determining what Listeners are relevant to and should be call for a given Event.  It may determine both what Listeners are relevant and the order in which to return them by whatever means it chooses.  That MAY include
+A Listener Provider is a service object responsible for determining what Listeners are relevant to and should be called for a given Event.  It may determine both what Listeners are relevant and the order in which to return them by whatever means it chooses.  That MAY include:
```
weierophinney: - `s/call/called` (typo)
- Added a `:` before the bulleted list (consistency)

proposed/event-dispatcher.md
```diff
@@ -88,11 +90,11 @@ A Listener Provider is a service object responsible for determining what Listene
 * Extracting some information from an object referenced by the Event, such as an Entity, and calling pre-defined lifecycle methods on that object.
 * Delegating its responsibility to one or more other Listener Providers using some arbitrary logic.
 
-Or some combination of those, or some other mechanism as desired.
+Any combination of the above, or other mechanisms, MAY be used as desired.
```
weierophinney: I think the proposed changes make the intention more clear here: providers have great leeway in how they identify relevant listeners.

proposed/event-dispatcher.md
Crell: Hrm.  My concern here is a PSR-14 "compatible" implementation that does something with the return values.  Then someone relying on that is no longer PSR-14 compatible as soon as they switch to one that doesn't.

proposed/event-dispatcher.md
Crell: Hm, that's confusing then.  My point here was that if you serialize an event, stick it in a queue, and load it later, then trying to stop propagation on it is kinda pointless as nothing is going to be able to check or operate on it.

Do you think that's sufficiently covered by the previous sentence?

proposed/event-dispatcher.md
Crell: Agreed, but strictly speaking the new wording here doesn't say that the listeners are invoked on the event, just that they're invoked.  The listener must be given the event.

Is that sufficiently covered elsewhere that we don't need to cover it here?  If not, needs a bit more wordsmithing.

Crell:
A few comments, nothing dramatic.  I'm :ok_hand: on anything I didn't comment on.

proposed/event-dispatcher.md
weierophinney: Yes.

proposed/event-dispatcher.md
Crell: :thinking: What if we changed it to a SHOULD, but also added a MUST NOT for the dispatcher, that it's not allowed to do anything meaningful with the return?  That would still allow the layering BC you're talking about but would, I think, preclude any incompatible variations.

weierophinney: That would work; shall I propose a wording via a new commit?

proposed/event-dispatcher.md
Crell: Endorse.

PR #1142, opened 2019-01-22
[PSR-12] Added variadic arguments

samdark:


JanTvrdik:
What about references and types?

samdark:
Would you please elaborate?

JanTvrdik:
There should be an example with both reference and type present, e.g. `string &...$parts` to make it clear how they should be combined together.

proposed/extended-coding-style-guide-meta.md
```diff
@@ -160,6 +160,7 @@ specification for a full understanding of its contents.
 * Operators - Section 6
 * Anonymous classes - Section 8
 * Type casting - Section 9
+* Variadic argument three dots MUST NOT have spaces before variable - Section 4.5
```
Jean85:
```suggestion
* Variadic and reference argument operators - Section 4.5
```

proposed/extended-coding-style-guide.md
```diff
@@ -546,6 +546,15 @@ class ReturnTypeVariations
 }
 ~~~
```
Jean85:
```suggestion

When using the reference operator `&` before an argument, there MUST be a space before the operator, and there MUST NOT be a space after it, like in the previous examples.

```

proposed/extended-coding-style-guide.md
```diff
@@ -546,6 +546,15 @@ class ReturnTypeVariations
...
 
+Variadic argument three dots MUST NOT have space before argument name:
+
+```php
+public function process(string $algorithm, ...$parts)
+{
+    // processing
+}
+```
+
```
Jean85:
~~~suggestion

When combining both the reference operator and the variadics three dots, there MUST NOT be any space between the two of them:

```php
public function process(string $algorithm, &...$parts)
{
    // processing
}
```

~~~

proposed/extended-coding-style-guide.md
JanTvrdik: The space should imho be there only when type is present, i.e. `function foo(&$args)`  and not `function foo( &$args)`

samdark: I guess, "there MUST be a space before the operator" should be omitted since that space is covered by "space after comma" rule and, as @JanTvrdik said, is wrong in case argument is first.

Jean85: Yup, right.

samdark:
@JanTvrdik, @Jean85 done. Please re-check.

proposed/extended-coding-style-guide.md
```diff
@@ -539,13 +539,35 @@ namespace Vendor\Package;
...
     {
         return 'foo';
     }
 }
 ~~~
 
+When using the reference operator `&` before an argument, there MUST NOT be
+a space after it, like in the previous example.
+
+Variadic argument three dots MUST NOT have space before argument name:
```
JanTvrdik: the wording of this sentence seems odd, but I'm not a native speaker, so I'm not sure how to write it properly

samdark: @michaelcullum would you please take a look?

ashnazg: "There MUST NOT be a space between the variadic three dots and the argument name."  ?

PR #1149, opened 2019-02-06
PSR-14 edits

ashnazg:
The extraneous 'the' shouldn't require much thought.

The `MAY` usage, though... the sentence sounds to me like it's solely describing an example to help illustrate the point.  But by using `MAY`, it implies a specific _spec-powerful_ meaning.  I didn't think that was the intent here.


bmack:
Agreed on both points as well! Thanks for the feedback!

PR #1150, opened 2019-02-12
Update PSR-1-basic-coding-standard.md

MrFibunacci:
Misspelling of `<?`

### DIFF
--- a/accepted/PSR-1-basic-coding-standard.md
+++ b/accepted/PSR-1-basic-coding-standard.md
@@ -14,7 +14,7 @@ interpreted as described in [RFC 2119].
 
 ## 1. Overview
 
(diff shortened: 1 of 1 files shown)


rob006:
This is not misspelling, this is "short echo" syntax: http://php.net/manual/en/function.echo.php

drupol:
> 
> 
> This is not misspelling, this is "short echo" syntax: http://php.net/manual/en/function.echo.php

Yes I know, but I think that in this context, it's not the right syntax.

rob006:
> If `<?=` is accepted as a tag opening, then it must be fixed in other places in the document.

What places? This document does not have any places where using short-echo would make sense. The whole point of this rule is to disallow `<?`.

drupol:
On line 17 and 37.

PR #1152, opened 2019-02-23
rephrase the delegate thin wrapper example

drealecs:
If the example looks like this, I'm trying to rephrase it for better understanding:
```php
$thinWrapperListener = function(BusinessEvent $event) use ($container) {
    $businessService = $container->get(BusinessService::class);
    $businessService->process($event);
};
```

### DIFF
--- a/proposed/event-dispatcher.md
+++ b/proposed/event-dispatcher.md
@@ -50,7 +50,7 @@ A Listener may be any PHP callable.  A Listener MUST have one and only one param
 
 A Listener SHOULD have a `void` return, and SHOULD type hint that return explicitly.  A Dispatcher MUST ignore return values from Listeners.
 
-A Listener MAY delegate actions to other code.  That includes a Listener being a thin wrapper around retrieving an object from a service container that contains the actual business logic to run, or other similar forms of indirection.  In that case, the callable containing the actual business logic SHOULD conform to the same rules as if it were called directly as a Listener.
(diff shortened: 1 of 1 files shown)


Crell:
Yes, that's what's being described in the spec, although it would typically be handled inside the Provider.  I don't think the proposed language really changes anything, though.  It's still the same thing and doesn't seem any more or less readable than what's there.

ostrolucky:
It's not the same thing. Old spec for some reason highlight service container, while it's not what's important in the sentence. It doesn't matter where was the object pulled from. It also implies wrapper must pull it from container, as if it would not apply to this

```php
$businessService = $container->get(BusinessService::class);
$thinWrapperListener = function(BusinessEvent $event) use ($businessService) {
    $businessService->process($event);
};
```

drealecs:
I agree with the last sentence of the paragraph not expressing what it should. I mentioned this also on an email and I think it should be rephrased.
I guess it's a good idea to drop the whole paragraph altogether. IMO, it just provides an example of what a listener can be or can do, out of a lot of other examples. And this example was one of the most obvious considering "we can solve any problem by introducing an extra level of indirection".
For now I'm going to change the PR proposal to removing most of the paragraph in order to keep it simple and clear.

Crell:
After sleeping on it, I'm on board with the latest version here, too.  :+1:

PR #1161, opened 2019-03-30
[PSR-12] Punctuation and articles

samdark:


proposed/extended-coding-style-guide.md
```diff
@@ -420,7 +420,7 @@ Method names MUST NOT be prefixed with a single underscore to indicate
 protected or private visibility. That is, an underscore prefix explicitly has
 no meaning.
 
-Method and function names MUST NOT be declared with a space after the method name. The
+Method and function names MUST NOT be declared with space after the method name. The
```
Jean85: I'm not sure that removing the article here is correct, we need to say that we require one single space.

samdark: That's the opposite: "MUST NOT be declared with space after the method name" but that should not matter.

Jean85: OH right! I misread.

proposed/extended-coding-style-guide.md
```diff
@@ -896,7 +896,7 @@ if ($a === $b) {
 
 ## 7. Closures
 
-Closures MUST be declared with a space after the `function` keyword, and a
+Closures MUST be declared with space after the `function` keyword, and a
```
jrfnl: IMO, this makes this sentence less clear - `with a space` = precisely one space, `with space` could mean any number of spaces, tabs or even new lines.

Jean85: Same objection that I made: https://github.com/php-fig/fig-standards/pull/1161#discussion_r270668995

But since it's a negation (`MUST NOT`) it doesn't matter how many, no spaces are allowed.

ashnazg: @jean85, I don't follow what you mean...

This line is a `MUST`, not a `MUST NOT`...

Without the `a`, it does read as "MUST have Greater Than Zero spaces"... thus, 20 spaces after `function` is valid.

Jean85: Umh you're right, my comment was on another line!

samdark: This case, indeed, may cause issues. I'll send a pull request.

samdark: Here we are: https://github.com/php-fig/fig-standards/pull/1163

PR #1162, opened 2019-03-30
[PSR-12] restructuring around operators

Theriault:
- removed "excluding `!` which is unary" - seemed unnecessary when `~`, `+` (identity), `-` (negation) are not listed 
- typecasting in PHP is a unary operation - relocating under operators
- removing `insteadof` and `as` from operators as they are not used within expressions.
  Similar to `as` and `=>` in `foreach ($array as $key => $value)`, `=` in `define(strict_types=1)`, etc.
  That is, they don't take one or more operands and yield a result. Additionally, styling around these
  symbols is already explained elsewhere.
- the pipe for multiple exceptions is a symbol that would be better explained under
  the `try-catch-finally` section instead of under operators (same reasoning as above)
- instead of "all binary operators MUST ..." followed by a separate list of which ones that includes, I joined the two statements to be more explicit. This also eliminates the possibility of other operators that may be classified as binary operators from being included, such as the scope resolution operator `::`, object operator `->`, etc. - as these are valid operators within expressions where both sides are operands (`($foo ? $a : $b)->{$bar ? "x" : "y"}`)
- simplified string concatenation operators to string operators. 
- restructuring operators section to be grouped by arity for readability and clarity
- minor fix of a link to be consistent with the others (http://php.net instead of https://secure.php.net)

### DIFF
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -862,7 +862,7 @@ try {
     // try body
 } catch (FirstThrowableType $e) {
     // catch body
-} catch (OtherThrowableType $e) {
+} catch (OtherThrowableType | AnotherThrowableType $e) {
     // catch body
 } finally {
     // finally body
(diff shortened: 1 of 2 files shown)


proposed/extended-coding-style-guide.md
```diff
@@ -871,29 +871,45 @@ try {
...
-    $variable = $foo ? 'foo' : 'bar';
+    $foo = $a + $b * $c;
 }
 ~~~
 
+### 6.3. Ternary operators
+
+The conditional operator, also known simply as the ternary operator, MUST be
+preceded and followed by at least one space around both the `?`
+and `:` characters:
```
Jean85: What about the compressed (`?:`) form?

Theriault: Good catch! At that point, there's 2 operands. I think other languages call it the Elvis operator. PHP has it classified under comparison operators, so in a sense it's already being addressed by requiring space around binary comparison operators. But that is probably not explicit/intuitive enough to leave as is, as PHP also lets you put spaces/comments between the 2 characters `$foo ? /*this operand is optional*/ : $bar` while still behaving as a binary operator. So do we just add to the standard that when the middle operand is omitted, the operator MUST always be written as `?:` and MUST be preceded and followed by at least one space?

Jean85: :+1: for me, but we should check if it should be counted as a clarification or as a new rule. @dragonmantank WDYT?

proposed/extended-coding-style-guide.md
```diff
@@ -862,7 +862,7 @@ try {
     // try body
 } catch (FirstThrowableType $e) {
     // catch body
-} catch (OtherThrowableType $e) {
+} catch (OtherThrowableType | AnotherThrowableType $e) {
```
samdark: Good addition.

proposed/extended-coding-style-guide.md
```diff
@@ -1044,22 +1067,14 @@ $instance = new class extends \Foo implements
 };
 ~~~
 
-## 9. Type Casting
```
samdark: There was a reference to it in meta-document. Should be checked.

proposed/extended-coding-style-guide.md
Theriault: I removed the one part requiring that it MUST be written as `?:` - that could definitely be seen as a new rule that the original document didn't specify, but I left the clarification that when the middle operand is left out, format it like other binary comparison operators. So the current version should contain no new rules, just clarifications and reformatting.

proposed/extended-coding-style-guide.md
Theriault: Good catch! I updated the meta-doc to contain the new structure. Thanks!

Forien:
I only find it weird that some unary operators (like `+` in `+=`, `!` in `!$var` or `++` in `$var++`) are to be without space, but typecasting is shown with space.

How about rule like this:
> There MUST NOT be any whitespace between the unary operator and the operand.

dragonmantank:
I agree, I think this just clears things up, expanding that single paragraph and also providing examples. I don't think this materially changes anything.

PR #1165, opened 2019-04-18
[PSR-12] Proposed alternative to function keyword arrangements

joshbruce:
First time, apologies if over explaining and for the bumps.

Functions *require* three things (to the best of my knowledge) `function` keyword, name of function, and a parenthetical marking argument collection in that order. (Default visibility of `public`.)

PSR 12 states that optional arguments within the function MUST come last; establishing a left-right relationship between required-optional, respectively. Agreed and already a good convention from the standard library.

PSR 12 makes visibility a required keyword (`private`, `protected`, `public`) to make the implicit explicit. 

This PR suggests that visibility remain closest to the center of gravity with the other *required* elements in all instances (break from PSR-1 & 2), which is not the case as of this writing as `static` can be placed *between* visibility and `function` keywords.

Further, this PR suggests that linguistically the hierarchy reads "better" when answering the following questions in turn:

1. Does it have an implementation or not (`abstract`)?
2. Can it be overridden (`final`)?
3. Is it a class function or an instance method (`static`)?
4. How can I access it, if at all?

From left to right.

### DIFF
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -63,7 +63,7 @@ class Foo extends Bar implements FooInterface
         }
     }
 
-    final public static function bar()
+    final static public function bar()
     {
         // method body
     }
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -575,11 +575,11 @@ public function process(string $algorithm, &...$parts)
 
 ### 4.6 `abstract`, `final`, and `static`
 
-When present, the `abstract` and `final` declarations MUST precede the
-visibility declaration.
+When present, optional keywords (`abstract` or `final` and `static`) MUST precede the visibility declaration.
 
-When present, the `static` declaration MUST come after the visibility
-declaration.
+When present, `static` directly precedes visibility.
+
+When `static` is NOT present, `abstract` or `final` directly precedes visibility. If `static` IS present, `abstract` or `final` directly precedes `static`.
 
 ~~~php
 <?php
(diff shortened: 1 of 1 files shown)


samdark:
That's not what common conventions for other languages are: https://stackoverflow.com/questions/10299067/modifier-keyword-order-in-java

joshbruce:
That’s fair to a point and why the reasoning behind the recommendation is not based on convention or personal style; instead, it is self-referential using the language of PHP and PSR-12. 

(Numbered bullets are for referring back not to represent order of operations.)

1. PHP, as a language has a core requirement for declaring a `function`, without which there is a hard stop: function nameStartingWithCharacterNotNumber_may_contain_underscores() {}

2. PSR-12 says arguments with defaults (optional) MUST go on the outside (righthand side), compared to the core: public function foo(int $arg1, &$arg2, $arg3 = []) {}

3. PSR-12 says visibility (an optional keyword from a language perspective) MUST be present (lefthand side); thereby making the keyword required to comply with the recommendation: See 2

4. This recommendation says the logic of 2 should be applied to 3 (and other keywords); therefore, visibility should directly precede the `function` keyword in all cases.

Alternatively, making visibility optional would make the internal logic seem more consistent across the board.

Cheers,
Josh Bruce

dragonmantank:
Thanks for the opinions and the request.

Unfortunately, I think I'm going to close this PR. I don't think the proposed helps make anything cleaner, and the entire point of PSR-2 and PSR-12 is to be an opinionated set of guidelines (thereby forcing the public/protected/private keywords, even if we don't need `public` from a language standpoint).

joshbruce:
I’m sorry, I’m not sure I follow the part in parentheses??

PR #1166, opened 2019-04-21
Support to @deterministic tag

rentalhost:
_Early proposal, to be discussed._

This feature is similar to MySQL `DETERMINISTIC` for `STORED PROCEDURES`. 

It is very useful to help IDE to identify calls that could be avoided securely.

PS. It is my first PR for fig, so I don't know how it works. Please, forgive me. :P

### DIFF
--- a/proposed/phpdoc-tags.md
+++ b/proposed/phpdoc-tags.md
@@ -18,6 +18,7 @@ PSR-19: PHPDoc tags
   - [5.2.  @author](#52-author)
   - [5.3.  @copyright](#53-copyright)
   - [5.4.  @deprecated](#54-deprecated)
+  - [5.x.  @deterministic](#5x-deterministic)
   - [5.5.  @internal](#55-internal)
   - [5.6.  @link](#56-link)
   - [5.7.  @method](#57-method)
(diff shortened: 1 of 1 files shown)


ashnazg:
The toolmakers on the team will have to convince me that this can be properly utilized to "save their resources".  I follow the idea, but can't be sure it's really going to save their parsers that much work.

Crell:
Shouldn't this be called `@Pure` or similar, since "a pure function" is the universal terminology for such a thing?

That said, now that the base language has attributes of its own I strongly suspect markers like this are better implemented as attributes, not as docblocks.  Attributes probably should cause the scope of PSR-5/19/etc. to scale back dramatically.

Jean85:
Sure. IIRC the `#[Pure]` attribute is already present in JetBrains package.

rentalhost:
Actually, I think that "pure" is not needly the same as deterministic.

While "pure" means that the function have no side-effect if return is not used, "deterministic" means that "same argument have same return".

`sum($a, $b)` is deterministic if `$a` and `$b` is same in two different calls.

Crell:
A pure function isn't just lack of side effects, but also that the return value is consistent.  Viz, a pre function is deterministic by definition.

ashnazg:
I'm with @crell on this one... an attribute is probably the way this will denoted in the wild, and thus will likely evolve as the "standard" way to highlight it.

PR #1168, opened 2019-05-06
Update extended-coding-style-guide.md

deleugpn:
Previous text makes it look like `interfaces` references `extends`.

Sorry if PRs like these are not allowed. If that's the case, maybe the group can take the suggestion internally and decide whether it makes sense to change or not.


deleugpn:
I see now what you mean, makes total sense. For a moment there, I forgot that its impossible to extend multiple classes which led me to the confusion.

PR #1171, opened 2019-06-03
*Exception Implements Throwable

sukei:
Hello guys,

Why not implementing the Throwable interface for each exception interfaces defined by the PSRs? As you can see, it works very well: https://3v4l.org/Pf2SA. Is there any reason not doing so (except compatibility with PHP5)?

For now, each PSRs breaks all the different static analysis tools. Each of them state that a non-throwable interface can't be thrown. That's damn true and it breaks the interoperability promise IMHO. Doing so will even enforce a proper implementation of the PSR exceptions at the langage level itself.

The PHP5 compatibility seems to be the only valid argument against this and PHP 5.6 (7.0 as well) version is now [EOL](https://www.php.net/supported-versions.php). Maybe its time to propose a global update? Compatibility with old versions can still be achieved with composer using a specific version constraint. WDYT?

### DIFF
--- a/accepted/PSR-6-cache.md
+++ b/accepted/PSR-6-cache.md
@@ -416,7 +416,7 @@ namespace Psr\Cache;
 /**
  * Exception interface for all exceptions thrown by an Implementing Library.
  */
-interface CacheException implements \Throwable
+interface CacheException extends \Throwable
 {
 }
 ~~~
--- a/accepted/PSR-6-cache.md
+++ b/accepted/PSR-6-cache.md
@@ -416,7 +416,7 @@ namespace Psr\Cache;
 /**
  * Exception interface for all exceptions thrown by an Implementing Library.
  */
-interface CacheException
+interface CacheException implements \Throwable
 {
 }
 ~~~


accepted/PSR-6-cache.md
```diff
@@ -416,7 +416,7 @@ namespace Psr\Cache;
 /**
  * Exception interface for all exceptions thrown by an Implementing Library.
  */
-interface CacheException
+interface CacheException implements \Throwable
```
danopz: extends

sukei: I'm a piece of shit...

Crell:
If we were making this spec today, I'd agree.  However, it was released before PHP 7 was, and changing  the type now would be a spec change, even if a small one.  That requires going through a new PSR process, which I don't think anyone has the interest in doing right now.

shadowhand:
@Crell I would argue that this was implied but not enforceable before PHP 7. Couldn't an errata be created to address this, along with a `2.0.0` of the package to bump the version to `^7.0` ?

Jean85:
@shadowhand the issue is probably a little more complicated than that, because we would like to avoid a rift in impementations between 1.0 and 2.0.

I started a proposal on this topic on the ML just today: https://groups.google.com/d/topic/php-fig/OyC3plRYhqg/discussion

Jean85:
We've been discussing this kind of issues a lot lately, and we've condensed our proposed solution to updating PSR interfaces in a blog post. Please read it and provide your feedback!

https://www.php-fig.org/blog/2019/10/upgrading-psr-interfaces/

vdelau:
@Crell can we close this PR? I think this PR is not allowed as an amendment and indeed would require a new PSR.

Crell:
This was done as part of the PHP 8 update already.

PR #1173, opened 2019-07-08
Clarify php file terminated with single LF

bmewburn:
@MarkusRodler pointed out in bmewburn/vscode-intelephense#540 that if a blank line is a line containing 0 characters then a file cant end in a non-blank line AND be terminated with an LF.

### DIFF
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -83,7 +83,7 @@ each word is capitalized including the very first letter.
 
 All PHP files MUST use the Unix LF (linefeed) line ending only.
 
-All PHP files MUST end with a non-blank line, terminated with a single LF.
+All PHP files MUST be terminated with a single LF.
 
 The closing `?>` tag MUST be omitted from files containing only PHP.
 


proposed/extended-coding-style-guide.md
```diff
@@ -83,7 +83,7 @@ each word is capitalized including the very first letter.
 
 All PHP files MUST use the Unix LF (linefeed) line ending only.
 
-All PHP files MUST end with a non-blank line, terminated with a single LF.
+All PHP files MUST be terminated with a single LF.
```
shadowhand: I think this should read

> All PHP files MUST be terminated with a single blank line.

As "blank line" is used throughout the document, and the line ending was defined in the statement above.

MarkusRodler:
It would also be nice that the example in the beginning shows that there must be a single blank line. I know that GitHub does not display it, but maybe displayed this way:
~~~php
[…]
    final public static function bar()
    {
        // method body
    }
}
 
~~~

michaelcullum:
This PR changes the meaning as it would allow for multiple LFs at the end of the document. The current wording restricts to
```foo{LF}
bar{LF}```

this new wording would allow
```foo{LF}
bar{LF}
{LF}```

PR #1175, opened 2019-07-27
Use non-hypothetical formulation

PowerKiKi:
Using "be" instead of "would be" make it absolutely clear that it is unconditionally not allowed, in any and all situation


PowerKiKi:
implemented in 8319378c8533bc8b3d6ffe56094aa097be4d931d

PR #1176, opened 2019-07-27
Add example for increment/decrement operators spacing

PowerKiKi:


PowerKiKi:
I see your point, but IMHO this is not changing the rules established by the spec at all, it is merely a very trivial clarification that really cannot can be misinterpreted.

KorvinSzanto:
I'm not sure that I agree that this clarifies anything, the wording is already super clear. I definitely agree with @Jean85, no changes during voting phase.

PowerKiKi:
rebased

goetas:
Thank you @PowerKiKi

PR #1177, opened 2019-07-27
Enforce `elseif` instead of  `else if`

PowerKiKi:
This has a very low visual impact and is trivial to apply to existing code. So instead of leaving the specification opened to preferences, it is more beneficial to enforce our choice.

This also remove the only `SHOULD` that is not about line length, making the specification less about possibilities and more about enforced rules.

### DIFF
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -678,7 +678,7 @@ if ($expr1) {
 }
 ~~~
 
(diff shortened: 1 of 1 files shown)


PowerKiKi:
Fair enough, this change has a bit more impact than the others, and it makes sens to be careful about it.

KorvinSzanto:
Yeah I agree, we should keep this as a `SHOULD` and deal with it in a future PSR if we want to. Thanks for thinking about the spec @PowerKiKi!

PR #1178, opened 2019-07-28
Small fix in 'extends' and 'implements' section

d-Metrius:
Fix 'extends' and 'implements' section to make sense with 'in the case of interfaces' part.

### DIFF
--- a/proposed/extended-coding-style-guide.md
+++ b/proposed/extended-coding-style-guide.md
@@ -272,7 +272,7 @@ class ClassName extends ParentClass implements \ArrayAccess, \Countable
 }
 ~~~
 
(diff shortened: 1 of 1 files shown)


goetas:
List of `extends` in the class context makes no sense as there can be only one `extend`.

samdark:
Indeed the original meaning was covering the following:

```php
interface MyInterface extends
    A,
    B,
    C
{
  // ...
}
```

```php
class MyClass implements
    A,
    B,
    C
{
    // ...
}
```

d-Metrius:
> After the change multiple extends for interfaces was lost.

The wording 'in the case of interfaces' was not obvious to me. Now it is, it reads as extending multiple interfaces with an interface.

PR #1180, opened 2019-08-10
Deprecate PSR-2

goetas:
PSR-12 replaces PSR-2


goetas:
duplicate of https://github.com/php-fig/fig-standards/pull/1181

PR #1183, opened 2019-08-12
[PSR-12] Multiple small impact updates:

DocFX:
- wording/grammar
- new lines coherence
- examples enrichment matching given enumerated domains of application
- italics for word reference quoting
- added exceptions for HEREDOC/NOWDOC to trailing spaces
- switched to ordered list for PHP header file order listing to better reflect ordering
- updated wording to better reflect internal references
- added proposed complements for unary operators needing clarification (especially regarding the ambigous "- $a" / "-$a")

### DIFF
--- a/accepted/PSR-12-extended-coding-style-guide.md
+++ b/accepted/PSR-12-extended-coding-style-guide.md
@@ -76,7 +76,7 @@ class Foo extends Bar implements FooInterface
 
 Code MUST follow all rules outlined in [PSR-1].
 
-The term 'StudlyCaps' in PSR-1 MUST be interpreted as PascalCase where the first letter of
+The term '*StudlyCaps*' in PSR-1 MUST be interpreted as *PascalCase* where the first letter of
 each word is capitalized including the very first letter.
 
 ### 2.2 Files
@@ -85,7 +85,7 @@ All PHP files MUST use the Unix LF (linefeed) line ending only.
 
 All PHP files MUST end with a non-blank line, terminated with a single LF.
 
-The closing `?>` tag MUST be omitted from files containing only PHP.
+The closing `?>` tag MUST be omitted from files containing only PHP code.
 
 ### 2.3 Lines
 
@@ -96,7 +96,8 @@ The soft limit on line length MUST be 120 characters.
 Lines SHOULD NOT be longer than 80 characters; lines longer than that SHOULD
 be split into multiple subsequent lines of no more than 80 characters each.
 
-There MUST NOT be trailing whitespace at the end of lines.
+There MUST NOT be trailing whitespace at the end of lines, with the exception 
+of HEREDOC/NOWDOC syntax.
 
 Blank lines MAY be added to improve readability and to indicate related
 blocks of code except where explicitly forbidden.
@@ -114,8 +115,8 @@ All PHP reserved keywords and types [[1]][keywords][[2]][types] MUST be in lower
 
 Any new types and keywords added to future PHP versions MUST be in lower case.
 
-Short form of type keywords MUST be used i.e. `bool` instead of `boolean`,
-`int` instead of `integer` etc.
+Short form of type keywords MUST be used, i.e., `bool` instead of `boolean`,
+`int` instead of `integer`, etc.
 
 ## 3. Declare Statements, Namespace, and Import Statements
 
@@ -124,14 +125,14 @@ each of the blocks below MUST be separated by a single blank line, and MUST NOT 
 a blank line. Each block MUST be in the order listed below, although blocks that are
 not relevant may be omitted.
 
-* Opening `<?php` tag.
-* File-level docblock.
-* One or more declare statements.
-* The namespace declaration of the file.
-* One or more class-based `use` import statements.
-* One or more function-based `use` import statements.
-* One or more constant-based `use` import statements.
-* The remainder of the code in the file.
+1. Opening `<?php` tag.
+2. File-level docblock.
+3. One or more `declare` statements.
+4. The `namespace` declaration of the file.
+5. One or more class-based `use` import statements.
+6. One or more function-based `use` import statements.
+7. One or more constant-based `use` import statements.
+8. The remainder of the code in the file.
 
 When a file contains a mix of HTML and PHP, any of the above sections may still
 be used. If so, they MUST be present at the top of the file, even if the
@@ -180,6 +181,7 @@ class FooBar
 
 Compound namespaces with a depth of more than two MUST NOT be used. Therefore the
 following is the maximum compounding depth allowed:
+
 ~~~php
 <?php
 
@@ -191,7 +193,8 @@ use Vendor\Package\SomeNamespace\{
 };
 ~~~
 
-And the following would not be allowed:
+And the following would *not* be allowed:
+
 ~~~php
 <?php
 
@@ -207,6 +210,7 @@ opening and closing tags, the declaration MUST be on the first line of the file
 and include an opening PHP tag, the strict types declaration and closing tag.
 
 For example:
+
 ~~~php
 <?php declare(strict_types=1) ?>
 <html>
@@ -236,7 +240,7 @@ The term "class" refers to all classes, interfaces, and traits.
 Any closing brace MUST NOT be followed by any comment or statement on the
 same line.
 
-When instantiating a new class, parentheses MUST always be present even when
+When instantiating a new class, parentheses MUST always be present, even when
 there are no arguments passed to the constructor.
 
 ~~~php
@@ -298,7 +302,8 @@ class ClassName extends ParentClass implements
 ### 4.2 Using traits
 
 The `use` keyword used inside the classes to implement traits MUST be
-declared on the next line after the opening brace.
+declared on the next line after the first opening brace of the item 
+implementing it.
 
 ~~~php
 <?php
@@ -333,8 +338,8 @@ class ClassName
 }
 ~~~
 
-When the class has nothing after the `use` import statement, the class
-closing brace MUST be on the next line after the `use` import statement.
+When the class has nothing after the `use` import statements, the class
+closing brace MUST be on the next line after the last `use` import statement.
 
 ~~~php
 <?php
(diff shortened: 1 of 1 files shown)


accepted/PSR-12-extended-coding-style-guide.md
```diff
@@ -191,7 +193,8 @@ use Vendor\Package\SomeNamespace\{
 };
 ~~~
 
-And the following would not be allowed:
+And the following would *not* be allowed:
```
gmponos: Maybe here "SHOULD/MUST NOT" might be more suited?

samdark:
Most of it could be definitely considered errata.

accepted/PSR-12-extended-coding-style-guide.md
DocFX: I just fixed a few doubts I had, the PSR is fresh, not really intending to modify the M/S/C/W.

But maybe we should open another discussion regarding this one, IDK.

gmponos: Not sure what `M/S/C/W` means... 

If it means `must`/`should`/`could`/`would` then according to the first paragraph of the standard there is no `could` or `would`.. for everyone to be on the same page :)

> The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119.

Never the less I do agree that this particular part should be on a separate discussion.
Will try to open a new PR related to this... 

I just happened to notice it now that you applied these changes... Will leave this open until the new PR.

DocFX: Sure. :)

Thanks for the note, though!

accepted/PSR-12-extended-coding-style-guide.md
```diff
@@ -900,6 +913,8 @@ if ($a === $b) {
     $foo = $bar ?? $a ?? $b;
 } elseif ($a > $b) {
     $foo = $a + $b * $c;
+} elseif ($a <=> $c) {
+    $foo += $bar % $a & $b ** $c;
```
zonuexe: This looks like meaningless code(especially `elseif ($a <=> $c)` doesn't make sense), what is it intended for?

goetas: i guess is just an example, to show how to use the `<=>` operator

accepted/PSR-12-extended-coding-style-guide.md
```diff
@@ -882,8 +888,15 @@ All operators not described here are left undefined.
...
-The increment/decrement operators MUST NOT have any space between
-the operator and operand.
+The increment/decrement operators, the logical not operator, the conversion 
+and negation operators MUST NOT have any space between the operator and operand.
+~~~php
+$value++;
+--$value;
+$a = !$b;
+$c = +$d;
+$e = -$f;
```
goetas: Does it make sense the syntax
```php
$c = +$d;
$e = -$f;
```

?

accepted/PSR-12-extended-coding-style-guide.md
DocFX: This exactly has no sense.
But this exactly shows the syntax of less thought operators.

That was the idea behind that. :)

accepted/PSR-12-extended-coding-style-guide.md
DocFX: Those are unary PHP operators.
They can actually be confused with arithmetic operations.

`
$a = -$b; // opposite value in |N or |R sets in maths
$a = +$b; // identity convert the variable
$a = --$b; // convert and substract before using the value
$a = ++$b; // convert and sum before using the value
$a = 1 + $b; // convert and add 1
$a = 1 +$b; // still adds :)

$c = 1.0;
var_dump(+$c); // float(1)
var_dump(-$c); // float(-1)
var_dump(+-$c); // float(-1)
var_dump(-+$c); // float(-1)
`

accepted/PSR-12-extended-coding-style-guide.md
```diff
@@ -882,8 +888,15 @@ All operators not described here are left undefined.
 
 ### 6.1. Unary operators
 
-The increment/decrement operators MUST NOT have any space between
-the operator and operand.
+The increment/decrement operators, the logical not operator, the conversion 
+and negation operators MUST NOT have any space between the operator and operand.
+~~~php
```
PowerKiKi: Missing a newline here:

```suggestion

~~~php
```

PowerKiKi:
If this is merged, then #1176 should probably also be reconsidered.

accepted/PSR-12-extended-coding-style-guide.md
DocFX: Done, there were actually 3 missing new lines! :)

Theriault:
Should the `+` unary operator be called the identity operator and not the conversion operator? (https://www.php.net/manual/en/language.operators.arithmetic.php)

Also, if we're defining spacing around three additional unary operators with this PR, should we consider defining other unary operators at this time?

- bitwise not operator `~`: `~$a`
- error control operator `@`: `@fopen`

DocFX:
> 
> 
> Should the `+` unary operator be called the identity operator and not the conversion operator? (https://www.php.net/manual/en/language.operators.arithmetic.php)
> 
> Also, if we're defining spacing around three additional unary operators with this PR, should we consider defining other unary operators at this time?
> 
>     * bitwise not operator `~`: `~$a`
> 
>     * error control operator `@`: `@fopen`

That's what the manual calls it, actually. 

Regarding the error control operator, it's a very bad practice to use it, just like `goto`. There's no need to frame the use of bad practices. Also, the initial statement says: "All operators not described here are left undefined.".

accepted/PSR-12-extended-coding-style-guide.md
```diff
@@ -461,7 +466,7 @@ In the argument list, there MUST NOT be a space before each comma, and there
 MUST be one space after each comma.
 
 Method and function arguments with default values MUST go at the end of the argument
-list.
+list, before the variable length argument lists.
```
Jean85: I don't know if this can be considered a new rule or a simple clarification, depending on the behavior of the PHP engine. Ping @dragonmantank.

accepted/PSR-12-extended-coding-style-guide.md
```diff
@@ -882,10 +888,19 @@ All operators not described here are left undefined.
 
 ### 6.1. Unary operators
 
-The increment/decrement operators MUST NOT have any space between
-the operator and operand.
+The increment/decrement operators, the logical not operator, the conversion 
+and negation operators MUST NOT have any space between the operator and operand.
```
Jean85: Those are new operators added. It's not a simple fix.

accepted/PSR-12-extended-coding-style-guide.md
DocFX: So far, they need to be placed in the end (which is the only logical place they can be set to ^^).

https://www.php.net/manual/en/functions.arguments.php#functions.variable-arg-list

accepted/PSR-12-extended-coding-style-guide.md
DocFX: Agree, this is NOT a simple cosmetic/semantic rework, this is an extension of what was herebefore written. Do you think it damages the spec?

Jean85: Without this, I could merge this mysels as a secretary, as simple typos/fixes. Since this is a change, the editor (@dragonmantank) has the final say on that.

accepted/PSR-12-extended-coding-style-guide.md
```diff
@@ -1034,9 +1054,9 @@ $foo->bar(
 );
 ~~~
 
-## 8. Anonymous Classes
+## 8. Anonymous classes
```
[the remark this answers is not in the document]
DocFX: English grammar. No other title gets a random cap on that common word.

accepted/PSR-12-extended-coding-style-guide.md
```diff
@@ -564,7 +569,7 @@ public function process(string $algorithm, ...$parts)
 ```
 
 When combining both the reference operator and the variadic three dot operator,
-there MUST NOT be any space between the two of them:
+there MUST NOT be any space between the two of them, like in the examples above:
```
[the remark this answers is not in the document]
DocFX: I hesitated when fixing this, as a fore reference. Maybe it's not meant to be read from top to bottom.

accepted/PSR-12-extended-coding-style-guide.md
```diff
@@ -298,7 +302,8 @@ class ClassName extends ParentClass implements
 ### 4.2 Using traits
 
 The `use` keyword used inside the classes to implement traits MUST be
-declared on the next line after the opening brace.
+declared on the next line after the first opening brace of the item
```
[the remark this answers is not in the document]
DocFX: You can use put some other stuff in between, though no one reallys does it. And it's always better to explicit which brace is the good one.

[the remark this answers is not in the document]
DocFX: I surely can. Will do, along with the update do to the new conflicts. ;)

accepted/PSR-12-extended-coding-style-guide.md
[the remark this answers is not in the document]
DocFX: Done.

accepted/PSR-12-extended-coding-style-guide.md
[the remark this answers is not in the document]
DocFX: Done.

DocFX:
@KorvinSzanto done editing your changes advice. Thanks a lot.
The question from @Jean85 regarding the variable length arguments list is still pending, though I think we all agree on that, despite requiring care.

KorvinSzanto:
Thanks for these changes @DocFX! I think the variable length change makes sense given that the language requires you to do it that way.

PR #1185, opened 2019-08-26
"Would" does not exist in RFC 2119

gmponos:
Follow up from here: https://github.com/php-fig/fig-standards/pull/1183#discussion_r313064864


goetas:
Thanks @gmponos

PR #1187, opened 2019-09-06
Fix position of full stop

u-mulder:
I presume it should be in front of `PSR-12`, otherwise I don't see any sense in these two sentences.


goetas:
Thanks!

PR #1190, opened 2019-10-02
PSR-1: fix example

WinterSilence:
Current example may be confusing because not backward compatible with psr-12:
> requires adherence to PSR-1
> All binary arithmetic, comparison, assignment, bitwise, logical, string, and type operators MUST be preceded and followed by at least one space

### DIFF
--- a/accepted/PSR-1-basic-coding-standard.md
+++ b/accepted/PSR-1-basic-coding-standard.md
@@ -89,7 +89,7 @@ function foo()
 }
 
 // conditional declaration is *not* a side effect
-if (! function_exists('bar')) {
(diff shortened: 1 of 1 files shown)


goetas:
Ping @KorvinSzanto

rob006:
`!` is not binary - it is unary operator.

Jean85:
As @rob006 correctly reported, that's a unary operator, so it's not covered by that rule.

WinterSilence:
@Jean85 @rob006 it,s logic operator - https://www.php.net/manual/en/language.operators.logical.php

rob006:
@WinterSilence It is **unary** logical operator.

PR #1192, opened 2019-10-11
Added EventInterface for correct overloading

NikolasSumrak:
Current realization of EventDispatcherInterface not compatible with overloading:
```php
class Dispatcher implements EventDispatcherInterface
{
    public function dispatch(EventInterface $event) { ... }
    // Fatal error:  Declaration of Dispatcher::dispatch(EventInterface $event) must be compatible with EventDispatcherInterface::dispatch(object $event)
}
```

### DIFF
--- a/accepted/PSR-14-event-dispatcher.md
+++ b/accepted/PSR-14-event-dispatcher.md
@@ -123,7 +123,6 @@ class EventInterface
 {
 
 }
-
 ```
 
 ```php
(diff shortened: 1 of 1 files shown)


Crell:
PSR-14 is already approved, so no modifications are allowed.

We discussed having an Event interface extensively during development.  It was explicitly decided against.  See the meta doc and the blog series for more details on how and why that decision was made:

https://steemit.com/php/@crell/psr-14-a-major-event-in-php

Chi-teck:
> See the meta doc and the blog series for more details on how and why that decision was made

@Crell The article you referenced does not cover the subject. Can you point the exact document with rationale for using `Object` over `EventInterface`?

Crell:
That article is part 1 of a series.  If you want to understand PSR-14 in detail, I recommend reading the whole series.

In particular, the reason for not having an EventInterface was discussed in part 2: https://peakd.com/php/@crell/psr-14-all-about-events

PR #1195, opened 2019-10-14
Upgrade PSR interfaces

Jean85:
This is a draft of the new bylaw that I would like to propose to overcome the issue of evolving PSR interfaces. 

For reference, the discussion started [here on the ML](https://groups.google.com/d/topic/php-fig/OyC3plRYhqg/discussion), and this proposal was published [as a blog post](https://www.php-fig.org/blog/2019/10/upgrading-psr-interfaces/); the blog post (written by @Crell) also included a poll to gauge the interest from the community, that seems overall positive.

### DIFF
--- a/bylaws/008-psr-evolution.md
+++ b/bylaws/008-psr-evolution.md
@@ -5,6 +5,7 @@
 A PSR is often comprised of text and code, more specifically interfaces. Those interfaces are pieces of code that are released and tagged at a specific moment in time, but the PHP language doesn't stand still... it evolves over time.
 
 This means that those interfaces need to withstand that evolution, and sometimes they need to be updated, to leverage new language features that could help better enforce the behaviors exposed in the PSR itself.
+
 At the same time, a PSR cannot be changed after its release (only erratas allowed), to protect a package that declared compatibility from becoming de-facto not compatible anymore.
 
 This document defines a process to be followed in updating PSR interfaces, in a way that is not breaking in regard to behavior for end users, and with an appropriate upgrade path for the packages.
--- a/bylaws/008-psr-evolution.md
+++ b/bylaws/008-psr-evolution.md
@@ -1,0 +1,73 @@
+# PSR evolution
+
+## Scope and objectives
+
+A PSR is often comprised of text and code, more specifically interfaces; those interfaces are pieces of code that are released and tagged in a specific moment in time, but the PHP language and ecosystem doesn't stand still, it evolves over time.
+
+This means that those interfaces needs to withstand those evolutions, and sometimes they need to be updated, to leverage new language features that could help better enforce the behaviors exposed in the PSR itself.
+At the same time, a PSR cannot be changed after its release if not with erratas, to avoid that a package that declared compatibility would be de-facto not compatible anymore.    
+
+This document defines a guideline to be followed in updating PSR interfaces, in a way that is not breaking in regard to behavior for end users, and with an appropriate upgrade path for the packages.
+
+## New releases
+
+A new minor release of a PHP-FIG package containing interfaces for a PSR MUST follow these rules:
+ * the new release MUST follow [Semantic Versioning](https://semver.org/) rules;
+ * the PSR behavior MUST NOT be altered;
+ * the packages that implement the interfaces or users that are consuming them MUST NOT suffer breaking changes;
+ * the PHP version constraint of the PHP-FIG package MAY be altered to require newer language features that would aid cross-compatibility;
+ * the PHP version constraint of the PHP-FIG package MUST NOT be altered to use newer language features that would create cross-compatibility issues.
+ 
+A new major release of a PHP-FIG package containing interfaces for a PSR MUST follow the same rules, with this exception:
+ * the new major version of the package MAY contain breaking changes if the implementing packages have a reasonable upgrading path, like the possibility of releasing a cross-compatible implementation with the previous releases.
+ * the new major version of the package MAY refer to a new, superseding PSR.
+
+### Workflow
+
+Since releasing new versions of the interfaces MUST NOT alter the PSR in its behavior, those releases can be voted in with the same process as errata changes. The new releases MUST be declared and embedded in the PSR document, with eventual indications on the upgrade path in the meta document.
+
+### Practical example
+
+A common case for an upgrade in the interfaces is to add parameters and return types, since they are a new language feature introduced by PHP 7, and many PSR interfaces predates that release. In the next paragraph, we'll try to use PSR-3 as an example to show how a PSR interface should be updated.
+
+#### PSR-3: the interface
+
+PSR-3 is released with the [`psr/log` package](https://packagist.org/packages/psr/log) and it contains the `LoggerInterface`, which has this method:
+```php
+/**
+ * Logs with an arbitrary level.
+ *
+ * @param mixed  $level
+ * @param string $message
+ * @param array  $context
+ *
+ * @return void
+ */
+public function log($level, $message, array $context = array());
+```
+This method could be updated with a new minor release that adds the argument type for `$message`:
+```php
+public function log($level, string $message, array $context = []);
+```
+This change would be technically a breaking change but, thanks to the [limited contravariance possible in PHP 7.2](https://wiki.php.net/rfc/parameter-no-type-variance), we can avoid that. This means that just by requiring `"php": "^7.2"` in the `prs/log` `composer.json`, we could tag this change as a minor release, and have all the implementors be automatically cross-compatible, provided that they declare `"psr/log": "^1.0"` (or equivalent) as a constraint, which they should.
+
+After this intermediate step, it would be possible to release a new major, that would add the return type:
+```php
+public function log($level, string $message, array $context = []): void;
+```
+This must be released as a new major version of `psr/log` (2.0); any package that would implement this would be able to declare `"psr/log": "^1.2 || ^2.0"`, since backward compatibility to the first release would be impossible, due to the sum of covariance and contravariance rules.
+
+#### PSR-3: the implementation
+
+On the other side, the implementing packages would be able to do a release cycle in the opposite fashion. Using Monolog as an example, the first release looks like this:
+```php
+public function log($level, $message, array $context = array());
+```
+The second release (which at the time of this writing is already tagged as 2.0) adds the return type, maintaining compatibility with the original interface:
+```php
+public function log($level, $message, array $context = []): void;
+```
+A possible third release could be tagged in the future, adding the argument type too:
+```php
+public function log($level, string $message, array $context = []): void;
+```
--- a/bylaws/008-psr-evolution.md
+++ b/bylaws/008-psr-evolution.md
@@ -43,7 +43,9 @@ PSR-11 is released with the [`psr/container` package](https://packagist.org/pack
      */
     public function has($id);
 ```
+
 This method could be updated with a new minor release that adds the argument type for `$id`:
+
 ```php
 public function has(string $id);
 ```
--- a/bylaws/008-psr-evolution.md
+++ b/bylaws/008-psr-evolution.md
@@ -45,7 +45,7 @@ This method could be updated with a new minor release that adds the argument typ
 ```php
 public function has(string $id);
 ```
-This change would technically be a breaking change, but thanks to the [limited contravariance possible in PHP 7.2](https://wiki.php.net/rfc/parameter-no-type-variance), we can avoid that. This means that just by requiring `"php": "^7.2"` in the `prs/container` `composer.json`, we could tag this change as a minor release, and have all the implementors be automatically cross-compatible, provided that they declare `"psr/container": "^1.0"` (or equivalent) as a constraint, which they should.
+This change would technically be a breaking change, but thanks to the [limited contravariance possible in PHP 7.2](https://wiki.php.net/rfc/parameter-no-type-variance), we can avoid that. This means that just by requiring `"php": "^7.2"` in the `psr/container` `composer.json`, we could tag this change as a minor release, and have all the implementors be automatically cross-compatible, provided that they declare `"psr/container": "^1.0"` (or equivalent) as a constraint, which they should.
 
 After this intermediate step, it would be possible to release a new major version, that would add the return type:
 ```php
--- a/bylaws/008-psr-evolution.md
+++ b/bylaws/008-psr-evolution.md
@@ -35,7 +35,18 @@ Note that if the upgrade path causes the consumers to maintain multiple versions
 
 ### Workflow
 
-Since releasing new versions of the interfaces MUST NOT alter the PSR in its behavior, those releases can be voted in with the same process as errata changes. The new releases MUST be declared and embedded in the PSR document, with eventual indications on the upgrade path in the meta document.
+Since releasing new versions of the interfaces MUST NOT alter the PSR in its behavior, those releases can be voted in with the same process as errata changes. The new releases MUST be declared and embedded with a brief explanation and a link in the PSR document, like in the following example:
+
+> \`\`\`php
+interface ContainerInterface
+{
+> // code snippet here
+}
+\`\`\`
+>
+> Since [psr/container version 1.1](#), the above interface has been updated to add argument type hints.
+
+The meta document MUST be amended with eventual indications on the upgrade path for the consumers.
 
 ### Practical example
 
--- a/bylaws/008-psr-evolution.md
+++ b/bylaws/008-psr-evolution.md
@@ -12,8 +12,8 @@ This document defines a process to be followed in updating PSR interfaces, in a 
 
 ## Definitions
 
- * **Consumer** - libraries and projects that implement and/or consume the code released as part of the PSR in question;
- * **Cross-compatibility** - the ability for a consumer to support more than one code version of the PSR with a single release of their own;
+ * **Consumer** - libraries and projects that implement and/or consume the code released as part of the PSR in question.
+ * **Cross-compatibility** - the ability for a consumer to support more than one code version of the PSR with a single release of their own.
 
 ## New releases
 
--- a/bylaws/008-psr-evolution.md
+++ b/bylaws/008-psr-evolution.md
@@ -69,7 +69,9 @@ The second release would adds the return type, maintaining compatibility with th
 ```php
 public function has($id): bool;
 ```
-A third release would be tagged then, adding the argument type too:
+
+A third release would also add the argument type hint:
+
 ```php
 public function has(string $id): bool;
 ```
--- a/bylaws/008-psr-evolution.md
+++ b/bylaws/008-psr-evolution.md
@@ -4,7 +4,7 @@
 
 A PSR is often comprised of text and code, generally interfaces. Those interfaces are pieces of code that are released and tagged at a specific moment in time. However, the PHP language doesn't stand still; it evolves over time.
 
-This means that those interfaces need to both provide a stable contract, as well as evolve to leverage new language features that could help better enforce the behaviors exposed in the PSR itself.
+This means that those interfaces need to both provide a stable contract, as well as evolve to leverage new language features that could help better enforce the behaviors described in the PSR itself.
 
 At the same time, a PSR cannot be changed after its release (at which point only erratas are allowed), to protect a package that declared compatibility from becoming de facto incompatible.
 
--- a/bylaws/008-psr-evolution.md
+++ b/bylaws/008-psr-evolution.md
@@ -57,6 +57,7 @@ After this intermediate step, it would be possible to release a new major versio
 ```php
 public function has(string $id): bool;
 ```
+
 This must be released as a new major version of `psr/container` (2.0); any package that would implement this would be able to declare `"psr/container": "^1.1 || ^2.0"`, since backward compatibility to the first release would be impossible, due to the sum of covariance and contravariance rules.
 
 #### PSR-11: the implementation
--- a/bylaws/008-psr-evolution.md
+++ b/bylaws/008-psr-evolution.md
@@ -10,13 +10,10 @@ At the same time, a PSR cannot be changed after its release (at which point only
 
 This document defines a process to be followed in updating PSR interfaces, in a way that is not breaking in regard to behavior for end users, and with an appropriate upgrade path for the consumers.
 
-## Terminology
+## Definitions
 
-Here are listed some words that are used in this document, with a brief explanation of their meaning in this context:
-
- * consumer: libraries and projects that implement and/or consume the code released as part of the PSR in question;
- * cross-compatibility: the ability for a consumer to support more than one code version of the PSR with a single release of their own;
- *  
+ * **Consumer** - libraries and projects that implement and/or consume the code released as part of the PSR in question;
+ * **Cross-compatibility** - the ability for a consumer to support more than one code version of the PSR with a single release of their own;
 
 ## New releases
 
@@ -46,7 +43,9 @@ interface ContainerInterface
 >
 > Since [psr/container version 1.1](#), the above interface has been updated to add argument type hints.
 
-The meta document MUST be amended with eventual indications on the upgrade path for the consumers.
+In the example above, the last line is indicative of what should be added to the specification.
+
+The meta document MUST be amended with information detailing the consumer upgrade path.
 
 ### Practical example
 
--- a/bylaws/008-psr-evolution.md
+++ b/bylaws/008-psr-evolution.md
@@ -22,6 +22,8 @@ A new minor release of a PHP-FIG package containing interfaces for a PSR MUST fo
 A new major release of a PHP-FIG package containing interfaces for a PSR MUST follow the same rules, with this exception:
  * the new major version of the package MAY contain breaking changes if the implementing packages have a reasonable upgrade path, like the possibility of releasing a cross-compatible implementation with the previous releases;
  * the new major version of the package MAY refer to a new, superseding PSR.
+ 
+Note that if the upgrade path causes the consumers to maintain multiple versions of their libraries side-by-side, only to support multiple versions of the same PSR, the upgrade path is to be considered too steep.
 
 ### Workflow
 
--- a/bylaws/008-psr-evolution.md
+++ b/bylaws/008-psr-evolution.md
@@ -28,46 +28,42 @@ Since releasing new versions of the interfaces MUST NOT alter the PSR in its beh
 
 ### Practical example
 
-A common case for an upgrade in the interfaces is to add types for parameters and returns, since they are new language features introduced by PHP 7, and many PSR interfaces predate that release. We'll use a method from PSR-3 as an example of how a PSR interface could be updated.
+A common case for an upgrade in the interfaces is to add types for parameters and returns, since they are new language features introduced by PHP 7, and many PSR interfaces predate that release. We'll use a method from PSR-11 as an example of how a PSR interface could be updated.
 
-#### PSR-3: the interface
+#### PSR-11: the interface
 
-PSR-3 is released with the [`psr/log` package](https://packagist.org/packages/psr/log) and contains the `LoggerInterface`, which has this method:
+PSR-11 is released with the [`psr/container` package](https://packagist.org/packages/psr/container) and it holds the `ContainerInterface`, which has this method:
 ```php
-/**
- * Logs with an arbitrary level.
- *
- * @param mixed  $level
- * @param string $message
- * @param array  $context
- *
- * @return void
- */
-public function log($level, $message, array $context = array());
+    /**
+     * @param string $id Identifier of the entry to look for.
+     *
+     * @return bool
+     */
+    public function has($id);
 ```
-This method could be updated with a new minor release that adds the argument type for `$message`:
+This method could be updated with a new minor release that adds the argument type for `$id`:
 ```php
-public function log($level, string $message, array $context = []);
+public function has(string $id);
 ```
-This change would technically be a breaking change, but thanks to the [limited contravariance possible in PHP 7.2](https://wiki.php.net/rfc/parameter-no-type-variance), we can avoid that. This means that just by requiring `"php": "^7.2"` in the `prs/log` `composer.json`, we could tag this change as a minor release, and have all the implementors be automatically cross-compatible, provided that they declare `"psr/log": "^1.0"` (or equivalent) as a constraint, which they should.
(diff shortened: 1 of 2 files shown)


stefanotorresi:
It all looks good to me, except one thing: "PSR behavior" is not clearly defined and it's a very ambiguous concept; e.g. some users might argue that method signature changes may constitute a behavioural change.
That is, stating that a new release must not alter  "PSR behavior" is a rule that may be interpreted in different ways and, as such, I would suggest to clarify it further.

bylaws/008-psr-evolution.md
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * @return void
+ */
+public function log($level, $message, array $context = array());
+```
+This method could be updated with a new minor release that adds the argument type for `$message`:
+```php
+public function log($level, string $message, array $context = []);
+```
+This change would be technically a breaking change but, thanks to the [limited contravariance possible in PHP 7.2](https://wiki.php.net/rfc/parameter-no-type-variance), we can avoid that. This means that just by requiring `"php": "^7.2"` in the `prs/log` `composer.json`, we could tag this change as a minor release, and have all the implementors be automatically cross-compatible, provided that they declare `"psr/log": "^1.0"` (or equivalent) as a constraint, which they should.
```
drealecs: Indeed, this would not be a breaking change for implementing library.
But, as I was explaining on the thread, it will be a breaking change for project and libraries that calls the `log` method.
Even if the docblock specified what message parameter type is, there was no strong enforcement. After a minor release, there will be a language runtime enforcement.

Just as return types can be reduced for callers but cannot be reduced for implementors, similarly, parameter types can be reduced for implementors but cannot be reduced for callers.

simensen:
I'm having a hard time following the Monolog example. It looks wrong to me as-is, and I'm not sure if I need to have read everything else about this in order to understand it or not. It would be quite useful to use some example concrete version numbers for the interfaces that can be referenced by the Monolog examples to show how they work. Likewise, it would probably be helpful to have a snippet of a `composer.json` showing any PHP / PSR version constraints needed/required for the packages to work as advertised.

For example:

> The second release (which at the time of this writing is already tagged as 2.0) adds the return type, maintaining compatibility with the original interface:

Just adding a `void` return type is still compatible with the original interface? If so, why would we add the argument types first to the interface instead of return types? I suspect this requires Monolog to also do something like restrict PHP to "^7.2", but that isn't stated here so it could be a bit confusing.

Example:

> (rewritten)
>
> The second release (which at the time of this writing is already tagged as 2.0) adds the return type. 
>
> ```php
> public function log($level, $message, array $context = []): void;
> ```
>
> This would be a breaking change for versions of PHP earlier than PHP 7.2. However, from PHP 7.2 on, PHP will allow the implementation to add a return type even if the underlying interface does not specify one. Therefore, the new version of this package must restrict PHP versions to those at or after PHP 7.2.
>
> ```json
> {
>   "name": "monolog/monolog",
>   "require": {
>     "psr/log": "^1.0",
>     "php": "^7.2"
>   }
> }
> ```

I may be completely wrong with what I've written here, but was merely an attempt to show what I think might have helped me out to better understand things by showing the `composer.json` snippet and required constraints that should be used by an implementation that would make this all work.

----

> [...] do a release cycle in the opposite fashion.

Might be nice to explain why it should be done "opposite" and how this is helpful/is required. I've read a few of the posts on this discussion, but it still isn't clear to me why we introduce argument type hints first, and then the return types, on interfaces, but ask implementations to roll out return type hints first, and then the argument type hints...

That might be my bad for not reading closely enough on the out of band discussion, but it should be documented here so people understand, just from reading this bylaw, why this order is important since we are stating that it is...

bylaws/008-psr-evolution.md
Jean85: That's not true: the fact that the interface is requiring a certain type is not enforced until the library adds the same type too, thanks to PHP 7.2. At that point, it should be up to the libraries to decide if they want to adhere strictly to the interface, add the types and create a breaking change or not.

Due to LSP, implementing libraries are free to accept MORE types than the one required by the interface.

Jean85:
> Might be nice to explain why it should be done "opposite" and how this is helpful/is required. I've read a few of the posts on this discussion, but it still isn't clear to me why we introduce argument type hints first, and then the return types, on interfaces, but ask implementations to roll out return type hints first, and then the argument type hints...

The reason is following LSP principle, and hence covariance/contravariance rules: to respect LSP, implementing libraries can augment the type of the arguments, and restrict the type of the return, and the opposite is true for the PSR interfaces.

So, Monolog could accept any stringable object there, not adding any argument type, and still be PSR-compliant.

> Just adding a void return type is still compatible with the original interface? If so, why would we add the argument types first to the interface instead of return types? I suspect this requires Monolog to also do something like restrict PHP to "^7.2", but that isn't stated here so it could be a bit confusing.

Just adding a return type in the implementing library does NOT violate LSP, and does NOT require PHP 7.2 in on itself: https://3v4l.org/Yeb9A

Doing the opposite (adding argument type on the implementing library) breaks LSP, always: https://3v4l.org/VDujN

If you need more practical examples, I posted a spreadsheet with 3v4l links on the ML, which can be found here: https://docs.google.com/spreadsheets/d/1QY7AUVQbWnVCbOZqJX11tbz2gKr-pU0CeZezcyJsEPA

bylaws/008-psr-evolution.md
```diff
@@ -0,0 +1,73 @@
...
+```php
+public function log($level, $message, array $context = array());
+```
+The second release (which at the time of this writing is already tagged as 2.0) adds the return type, maintaining compatibility with the original interface:
+```php
+public function log($level, $message, array $context = []): void;
+```
+A possible third release could be tagged in the future, adding the argument type too:
+```php
+public function log($level, string $message, array $context = []): void;
```
gmponos: This is wrong and you might need to be extra careful here...as you want to evolve.. 

Docblock says that the argument is a string but the PSR in some point mentions that it could be an object that can be cast to a string.

gmponos: This is wrong and you might need to be extra careful here...as you want to evolve..

Docblock says that the argument is a string but the PSR in some point mentions that it could be an object that can be cast to a string.

gmponos: For the record

https://github.com/php-fig/log/pull/60

Jean85: You're right, probably PSR-3 is a bad example. I should use a different interface then.

bylaws/008-psr-evolution.md
drealecs: I agree that when adding types to parameters there is no breaking changing on the implementing libraries.
I argue that there would be a breaking change on the consuming libraries.

Jean85: That depends totally on the implementing libraries. As highlighted in https://github.com/php-fig/fig-standards/pull/1195#discussion_r334586301, for Monolog specifically they could choose to not enforce `string` due to stringable objects being accepted.

arcanisgk:
Hi guys, I have a serious question.
What do we do with arrays?

[Reddit Link](
https://www.reddit.com/r/PHPhelp/comments/dl6tsq/understanding_the_standards_psr_or_others_i_am/)

ARRAY intended to Store Objects

```
<?php
class Example {
    function fooFunction() {
        return "Hello World! Object<br>";
    }
}
$var_object = new Example;

//Check This build #1 and #2, Wich is correctly???:
$Test1 = array('key' => $var_object);  //is this correctly??
$Test2 = array('KEY' => $var_object);  //is this correctly??

//Calls best way????:
echo $Test1['key']->fooFunction();
echo $Test2['KEY']->fooFunction();
```

or so to Store Info....

Jean85:
@arcanisgk this is not the place for this kind of question. Keep those on a social network, like the linked Reddit thread.

Jean85:
I've switched the example from PSR-3 to PSR-11 to avoid the pitfall highlighted in https://github.com/php-fig/fig-standards/pull/1195#discussion_r334586301.

Please share your comments and feedback; if no blockers arise, I would like to put this officially in front of the CC and call a vote afterwards.

bylaws/008-psr-evolution.md
```diff
@@ -0,0 +1,69 @@
...
+
+A new minor release of a PHP-FIG package containing interfaces for a PSR MUST follow these rules:
+ * the new release MUST follow [Semantic Versioning](https://semver.org/) rules;
+ * the PSR behavior MUST NOT be altered;
+ * the packages that implement the interfaces or users that are consuming them MUST NOT suffer breaking changes;
+ * the PHP version constraint of the PHP-FIG package MAY be altered to require newer language features that would aid cross-compatibility;
+ * the PHP version constraint of the PHP-FIG package MUST NOT be altered to use newer language features that would create cross-compatibility issues.
+ 
+A new major release of a PHP-FIG package containing interfaces for a PSR MUST follow the same rules, with this exception:
+ * the new major version of the package MAY contain breaking changes if the implementing packages have a reasonable upgrade path, like the possibility of releasing a cross-compatible implementation with the previous releases;
```
johmanx10: It is my concern that the language used, "have a reasonable upgrade path", to define an exception, is broad and vague and candidate for easy abuse. Perhaps it is safe to bind this to a clear example of what is explicitly NOT "reasonable".

I believe something specific can be formulated along the lines of:
> If the upgrade path causes userland to maintain multiple versions of their software side-by-side, only to support multiple versions of the same PSR, the upgrade path is too steep.

I'm unclear whether that's an improvement, but would like at least to have this considered by people more clever than myself.

bylaws/008-psr-evolution.md
```diff
@@ -0,0 +1,69 @@
...
+     *
+     * @return bool
+     */
+    public function has($id);
+```
+This method could be updated with a new minor release that adds the argument type for `$id`:
+```php
+public function has(string $id);
+```
+This change would technically be a breaking change, but thanks to the [limited contravariance possible in PHP 7.2](https://wiki.php.net/rfc/parameter-no-type-variance), we can avoid that. This means that just by requiring `"php": "^7.2"` in the `prs/container` `composer.json`, we could tag this change as a minor release, and have all the implementors be automatically cross-compatible, provided that they declare `"psr/container": "^1.0"` (or equivalent) as a constraint, which they should.
```
nicolas-grekas: s/prs/psr/

bylaws/008-psr-evolution.md
```diff
@@ -0,0 +1,69 @@
+# PSR evolution
+
+## Scope and objectives
+
+A PSR is often comprised of text and code, more specifically interfaces. Those interfaces are pieces of code that are released and tagged at a specific moment in time, but the PHP language doesn't stand still... it evolves over time.
+
+This means that those interfaces need to withstand that evolution, and sometimes they need to be updated, to leverage new language features that could help better enforce the behaviors exposed in the PSR itself.
+At the same time, a PSR cannot be changed after its release (only erratas allowed), to protect a package that declared compatibility from becoming de-facto not compatible anymore.
```
Crell:
```suggestion

At the same time, a PSR cannot be changed after its release (only erratas allowed), to protect a package that declared compatibility from becoming de-facto not compatible anymore.
```

bylaws/008-psr-evolution.md
```diff
@@ -0,0 +1,69 @@
...
+ * the PHP version constraint of the PHP-FIG package MAY be altered to require newer language features that would aid cross-compatibility;
+ * the PHP version constraint of the PHP-FIG package MUST NOT be altered to use newer language features that would create cross-compatibility issues.
+ 
+A new major release of a PHP-FIG package containing interfaces for a PSR MUST follow the same rules, with this exception:
+ * the new major version of the package MAY contain breaking changes if the implementing packages have a reasonable upgrade path, like the possibility of releasing a cross-compatible implementation with the previous releases;
+ * the new major version of the package MAY refer to a new, superseding PSR.
+
+### Workflow
+
+Since releasing new versions of the interfaces MUST NOT alter the PSR in its behavior, those releases can be voted in with the same process as errata changes. The new releases MUST be declared and embedded in the PSR document, with eventual indications on the upgrade path in the meta document.
```
Crell: This would be the first time we make non-trivial changes to the spec document rather than the metadoc.  It... kinda makes sense in context here but the end result is that PSR-3 would end up with 3 versions of all its code in the spec doc.  That's... a lot of change.

I don't have a better idea off hand, but this worries me.

bylaws/008-psr-evolution.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Practical example
+
+A common case for an upgrade in the interfaces is to add types for parameters and returns, since they are new language features introduced by PHP 7, and many PSR interfaces predate that release. We'll use a method from PSR-11 as an example of how a PSR interface could be updated.
+
+#### PSR-11: the interface
+
+PSR-11 is released with the [`psr/container` package](https://packagist.org/packages/psr/container) and it holds the `ContainerInterface`, which has this method:
+```php
```
Crell:
```suggestion

```php
```

bylaws/008-psr-evolution.md
```diff
@@ -0,0 +1,69 @@
...
+     *
+     * @return bool
+     */
+    public function has($id);
+```
+This method could be updated with a new minor release that adds the argument type for `$id`:
+```php
+public function has(string $id);
+```
+This change would technically be a breaking change, but thanks to the [limited contravariance possible in PHP 7.2](https://wiki.php.net/rfc/parameter-no-type-variance), we can avoid that. This means that just by requiring `"php": "^7.2"` in the `psr/container` `composer.json`, we could tag this change as a minor release, and have all the implementors be automatically cross-compatible, provided that they declare `"psr/container": "^1.0"` (or equivalent) as a constraint, which they should.
```
Crell: I thought we were going to go with major versions here to avoid the different-environment issue that @nicolas-grekas pointed out.

Jean85: This could be discussed. I understand Nicolas motivation, but OTOH this approach would not require action from the implementing libraries on the first step, and that's why I preferred it.

nicolas-grekas: I've thought twice about this: given that no apps require any psr packages directly, it means that they come transitively. If we go with v2 here, and if all libs update their composer.json to allow it, then the different-environment issue appears also.
This means my argument is fragilized here, maybe not worth it. For thoughts.

drealecs: Actually, I believe that the expectancy would be that apps that calls the interfaces in their code (and only link interfaces to implementation some other code) actually require psr packages directly.
As this is what they should do. If you use a namespace in your code, you should have the package providing it in composer.json.

But I also believe that apps do not require psr packages. And a minor change will trigger a composer update to upgrade it to 1.1 from 1.0.
And mainly that's what I also argued earlier that a major version is needed as otherwise $container->has(1) that returns false, after change, will throw a `TypeError`.

Jean85: > [...] otherwise $container->has(1) that returns false, after change, will throw a TypeError.

That's true only if you're using `strict_types=1`, and only at the implementation change, not on the interface change; so the implementation can choose on its own if the change warrants a major or not.

gmponos:
I was wondering do you believe that there is any catch with packages that are not required but suggested.

I mean in my code I might easily have a check if a `class_exists` then use this `container` which implements PSR-11 or other `container` that does not

bylaws/008-psr-evolution.md
```diff
@@ -5,6 +5,7 @@
 A PSR is often comprised of text and code, more specifically interfaces. Those interfaces are pieces of code that are released and tagged at a specific moment in time, but the PHP language doesn't stand still... it evolves over time.
 
 This means that those interfaces need to withstand that evolution, and sometimes they need to be updated, to leverage new language features that could help better enforce the behaviors exposed in the PSR itself.
+
 At the same time, a PSR cannot be changed after its release (only erratas allowed), to protect a package that declared compatibility from becoming de-facto not compatible anymore.
 
 This document defines a process to be followed in updating PSR interfaces, in a way that is not breaking in regard to behavior for end users, and with an appropriate upgrade path for the packages.
```
arcanisgk: good work

bylaws/008-psr-evolution.md
drealecs: Thank you @Jean85 for explaining it again. You're right, I understood it now. It's true that IDE might complain about you using an interface method with a wrong parameter type but at run-time the check is done against the actual implementing class, not against the interface... obviously, I would say now.

I don't have any other issue with it being a minor version of psr package with php ^7.2 required.

Jean85:
> I was wondering do you believe that there is any catch with packages that are not required but suggested.

This kind of behavior is possible, but I fear that there's nothing we can do to avoid issues. A patch release from their end would fix this.

bylaws/008-psr-evolution.md
```diff
@@ -0,0 +1,71 @@
...
+
+This document defines a process to be followed in updating PSR interfaces, in a way that is not breaking in regard to behavior for end users, and with an appropriate upgrade path for the packages.
+
+## New releases
+
+A new minor release of a PHP-FIG package containing interfaces for a PSR MUST follow these rules:
+ * the new release MUST follow [Semantic Versioning](https://semver.org/) rules;
+ * the PSR behavior MUST NOT be altered;
+ * the packages that implement the interfaces or users that are consuming them MUST NOT suffer breaking changes;
+ * the PHP version constraint of the PHP-FIG package MAY be altered to require newer language features that would aid cross-compatibility;
```
weierophinney: I think you need to define what is meant by the term "cross-compatibility" before using it here. Are we talking "forwards compatibility", whereby the code adds features that would make it compatible with later versions of PHP, while retaining compatibility with current versions? Or are we talking compatibility with both consumer and implementor? or something else?

bylaws/008-psr-evolution.md
weierophinney: I agree with @johmanx10 here - "reasonable" is very subjective. Having examples of unreasonable and/or reasonable paths would help remove uncertainty and the possibility of unfair handling of different PSR evolutions.

I think the sentence Jan-Marten suggests is spot on.

bylaws/008-psr-evolution.md
weierophinney: It depends on what is meant by "declared and embedded in the PSR document". Does this mean embedding the different versions of the interface? Or does it mean a link to the tagged version of the new interfaces in their repo (e.g., https://github.com/php-fig/http-message/releases/tag/1.0.1)?

We need clarification on what that phrase means.

bylaws/008-psr-evolution.md
```diff
@@ -0,0 +1,71 @@
...
+
+```php
+    /**
+     * @param string $id Identifier of the entry to look for.
+     *
+     * @return bool
+     */
+    public function has($id);
+```
+This method could be updated with a new minor release that adds the argument type for `$id`:
```
weierophinney:
```suggestion

This method could be updated with a new minor release that adds the argument type for `$id`:

```

bylaws/008-psr-evolution.md
```diff
@@ -0,0 +1,71 @@
...
+```php
+public function has(string $id);
+```
+This change would technically be a breaking change, but thanks to the [limited contravariance possible in PHP 7.2](https://wiki.php.net/rfc/parameter-no-type-variance), we can avoid that. This means that just by requiring `"php": "^7.2"` in the `psr/container` `composer.json`, we could tag this change as a minor release, and have all the implementors be automatically cross-compatible, provided that they declare `"psr/container": "^1.0"` (or equivalent) as a constraint, which they should.
```
weierophinney:
```suggestion

This change would technically be a breaking change, but thanks to the [limited contravariance possible in PHP 7.2](https://wiki.php.net/rfc/parameter-no-type-variance), we can avoid that. This means that just by requiring `"php": "^7.2"` in the `psr/container` `composer.json`, we could tag this change as a minor release, and have all the implementors be automatically cross-compatible, provided that they declare `"psr/container": "^1.0"` (or equivalent) as a constraint.
```

(Adds whitespace before the paragraph, and removes editorializing.)

bylaws/008-psr-evolution.md
```diff
@@ -0,0 +1,71 @@
...
+
+After this intermediate step, it would be possible to release a new major version, that would add the return type:
```
weierophinney:
```suggestion
After this intermediate step, it would be possible to release a new major version, adding a return type hint:

```

bylaws/008-psr-evolution.md
```diff
@@ -0,0 +1,71 @@
...
+```php
+public function has(string $id): bool;
+```
+This must be released as a new major version of `psr/container` (2.0); any package that would implement this would be able to declare `"psr/container": "^1.1 || ^2.0"`, since backward compatibility to the first release would be impossible, due to the sum of covariance and contravariance rules.
```
weierophinney:
```suggestion

This must be released as a new major version of `psr/container` (2.0); any package that would implement this would be able to declare `"psr/container": "^1.1 || ^2.0"`, since backward compatibility to the first release would be impossible, due to the sum of covariance and contravariance rules.
```

Adds whitespace before paragraph.

bylaws/008-psr-evolution.md
```diff
@@ -0,0 +1,71 @@
...
+
+#### PSR-11: the implementation
+
+On the other side, the implementing packages would be able to do a release cycle in the opposite fashion. The first release looks like this:
```
weierophinney:
```suggestion
On the other side, the implementing packages would be able to do a release cycle in the opposite fashion. The first release looks like this:

```

(whitespace)

bylaws/008-psr-evolution.md
```diff
@@ -0,0 +1,71 @@
...
+```php
+public function has($id);
+```
+The second release would adds the return type, maintaining compatibility with the original interface:
```
weierophinney:
```suggestion

The second release would add the return type, maintaining compatibility with the original interface:

```

(whitespace, and removal of plural form)

bylaws/008-psr-evolution.md
```diff
@@ -0,0 +1,71 @@
...
+```php
+public function has($id): bool;
+```
+A third release would be tagged then, adding the argument type too:
```
weierophinney:
```suggestion

A third release would also add the argument type hint:

```

(whitespace, and active form)

bylaws/008-psr-evolution.md
Jean85: I wrote this thinking about the current specific case of bumping to 7.2 to allow dropping types. Would a reword like this one work for you?

```suggestion
 * the PHP version constraint of the PHP-FIG package MAY be increased to leverage newer language features, especially those that would aid cross-compatibility of consumers with the old and the new versions;
```

bylaws/008-psr-evolution.md
Jean85: I had in mind adding a new code snippet with the changes, something along the lines of "Since version 1.1, the interface was changed to receive argument types: <code snippet>".

Would this phrasing be fine for you?

```suggestion
Since releasing new versions of the interfaces MUST NOT alter the PSR in its behavior, those releases can be voted in with the same process as errata changes. The new releases MUST be declared and embedded in the PSR document following the original code snippets in the original document, with eventual indications on the upgrade path in the meta document.
```

Jean85:
Thanks @weierophinney for the very thorough review. I've applied all your suggestion, and left two discussions open with two proposed phrasings. Let me know if they are fine.

bylaws/008-psr-evolution.md
weierophinney: That sounds perfect!

weierophinney: Only part is: I'd put that in a "Terminology" section at the beginning (we do that in a few PSRs already, so it will be familiar to PSR readers). That way, when it's referenced elsewhere, the context is fully known.

bylaws/008-psr-evolution.md
weierophinney: The problem with saying "following the original code snippets" is that the current "code snippets" are full interface definitions generally - or do you mean something like this:

```markdown
\`\`\`php
interface ContainerInterface
{
    /**
     * Test whether or not a service is registered in the container.
     *
     * @param string $name
     * @return bool
     */
    public function has($name);


    /**
     * Retrieve a service from the container:
     *
     * @param string $name
     * @return object
     */
    public function get($name);
}
\`\`\`

> Since psr/container version 1.1, the above interface has been updated to add argument type hints.
```

(added escape characters on the embedded code sample so it would highlight properly)

Basically, I think even with clarifications to the verbiage, we may also need an example.

Jean85: Is e0c9954 enough for this?

bylaws/008-psr-evolution.md
Jean85: I've added that with 94b5cc79e6b9c35788737dd76164f87690a0cc71, is it ok?

bylaws/008-psr-evolution.md
stefanotorresi: Maybe there is another way to do this entirely: we are already versioning the spec document with Git, so why don't we just update the document with the latest changes, specifying the version in the header, and list all the various versions in the website?

Jean85: I fear that that approach would complicate the site a lot, since it would have to use the submodule at multiple commits, and commits would've have to be fixed. Or we would have to have branches. Too complicated for a simple document.

bylaws/008-psr-evolution.md
```diff
@@ -0,0 +1,101 @@
+# PSR evolution
+
+## Scope and objectives
+
+A PSR is often comprised of text and code, generally interfaces. Those interfaces are pieces of code that are released and tagged at a specific moment in time. However, the PHP language doesn't stand still; it evolves over time.
+
+This means that those interfaces need to both provide a stable contract, as well as evolve to leverage new language features that could help better enforce the behaviors exposed in the PSR itself.
+
+At the same time, a PSR cannot be changed after its release (at which point only erratas are allowed), to protect a package that declared compatibility from becoming de-facto not compatible anymore.
```
iansltx:
```suggestion
At the same time, a PSR cannot be changed after its release (at which point only erratas are allowed), to protect a package that declared compatibility from becoming de facto incompatible.
```

bylaws/008-psr-evolution.md
```diff
@@ -0,0 +1,101 @@
+# PSR evolution
+
+## Scope and objectives
+
+A PSR is often comprised of text and code, generally interfaces. Those interfaces are pieces of code that are released and tagged at a specific moment in time. However, the PHP language doesn't stand still; it evolves over time.
+
+This means that those interfaces need to both provide a stable contract, as well as evolve to leverage new language features that could help better enforce the behaviors exposed in the PSR itself.
```
iansltx:
```suggestion
This means that those interfaces need to both provide a stable contract, as well as evolve to leverage new language features that could help better enforce the behaviors described in the PSR itself.
```

bylaws/008-psr-evolution.md
```diff
@@ -0,0 +1,101 @@
...
+
+At the same time, a PSR cannot be changed after its release (at which point only erratas are allowed), to protect a package that declared compatibility from becoming de-facto not compatible anymore.
+
+This document defines a process to be followed in updating PSR interfaces, in a way that is not breaking in regard to behavior for end users, and with an appropriate upgrade path for the consumers.
+
+## Definitions
+
+ * **Consumer** - libraries and projects that implement and/or consume the code released as part of the PSR in question.
```
iansltx:
```suggestion
 * **Consumer** - a library or project that implements and/or consumes the code released as part of the PSR in question.
```

bylaws/008-psr-evolution.md
iansltx: @Jean85 Please either commit your suggestion (which looks good to me as well) or mark this conversation as resolved. The Terminology section helps, but it wouldn't hurt to have the extra clarity here.

bylaws/008-psr-evolution.md
Crell: One note on this part.  In PSRs, we tend to differentiate between an *Implementer* (who implements a PSR interface) and a *Consumer* (who composes it).  Strictly speaking both are "consumers" (in the economic sense) of a PSR, but I fear it may cause some confusion down the line.  Can we use the more standard terminology here, even if it makes it slightly longer?

Jean85: Done with aa3e198 :+1:

Jean85:
The bylaw vote on this document has been called:
 * CC vote thread: https://groups.google.com/d/topic/php-fig/L3kmFRNsZYo/discussion
 * Project Representatives vote thread: https://groups.google.com/d/topic/php-fig/F2-kYPW8UDk/discussion

Since the vote is in progress, no more edits will be accepted on the document; so, commit 
aa3e198 is to be considered the final form of this proposal.

bylaws/008-psr-evolution.md
```diff
@@ -0,0 +1,102 @@
...
+ * the new major version of the package MAY contain breaking changes if the implementing packages have a reasonable upgrade path, like the possibility of releasing a cross-compatible implementation with the previous releases;
+ * the new major version of the package MAY refer to a new, superseding PSR.
+ 
+Note that if the upgrade path causes the consumers or implementers to maintain multiple versions of their libraries side-by-side, only to support multiple versions of the same PSR, the upgrade path is to be considered too steep.
+
+### Workflow
+
+Since releasing new versions of the interfaces MUST NOT alter the PSR in its behavior, those releases can be voted in with the same process as errata changes. The new releases MUST be declared and embedded with a brief explanation and a link in the PSR document, like in the following example:
+
+> \`\`\`php
```
simPod: This does not render properly (at least on github) though.

Jean85: We lose the whitespace inside the code snippet, but the rest of the content is good:

immagine

bylaws/008-psr-evolution.md
```diff
@@ -0,0 +1,102 @@
...
+ * the PSR behavior MUST NOT be altered;
+ * consumers or implementers of those interfaces MUST NOT suffer breaking changes;
+ * the PHP version constraint of the PHP-FIG package MAY be increased to leverage newer language features, especially those that would aid cross-compatibility of consumers and implementers with the old and the new versions;
+ * the PHP version constraint of the PHP-FIG package MUST NOT be altered to use newer language features that would create cross-compatibility issues.
+ 
+A new major release of a PHP-FIG package containing interfaces for a PSR MUST follow the same rules, with this exception:
+ * the new major version of the package MAY contain breaking changes if the implementing packages have a reasonable upgrade path, like the possibility of releasing a cross-compatible implementation with the previous releases;
+ * the new major version of the package MAY refer to a new, superseding PSR.
+ 
+Note that if the upgrade path causes the consumers or implementers to maintain multiple versions of their libraries side-by-side, only to support multiple versions of the same PSR, the upgrade path is to be considered too steep.
```
marc1706: This IMHO is a bit unclear: `[...] the upgrade path is to be considered too steep.`
To steep for what? A minor release or a major release or a new release of the same PSR?
Also, I'd then state what the recommended step is if the upgrade path is considered too steep.

Jean85: The point here is that an approach that is to be considered too steep is to be discarded completely. As a PSR, we cannot try to outskirt the issues, we have less ways to circumvent problems because interop an easiness of use is paramount.

PR #1196, opened 2019-10-16
wording adjustments

ashnazg:


bylaws/008-psr-evolution.md
```diff
@@ -28,11 +28,11 @@ Since releasing new versions of the interfaces MUST NOT alter the PSR in its beh
 
 ### Practical example
 
-A common case for an upgrade in the interfaces is to add parameters and return types, since they are a new language feature introduced by PHP 7, and many PSR interfaces predates that release. In the next paragraph, we'll try to use PSR-3 as an example to show how a PSR interface should be updated.
+A common case for an upgrade in the interfaces is to add typehints for parameters and returns, since they are new language features introduced by PHP 7, and many PSR interfaces predate that release. We'll use a method from PSR-3 as an example of how a PSR interface could be updated.
```
Jean85: Those are not "hints", but full-blown types...
```suggestion
A common case for an upgrade in the interfaces is to add types for parameters and returns, since they are new language features introduced by PHP 7, and many PSR interfaces predate that release. We'll use a method from PSR-3 as an example of how a PSR interface could be updated.
```

bylaws/008-psr-evolution.md
```diff
@@ -2,12 +2,12 @@
 
 ## Scope and objectives
 
-A PSR is often comprised of text and code, more specifically interfaces; those interfaces are pieces of code that are released and tagged in a specific moment in time, but the PHP language and ecosystem doesn't stand still, it evolves over time.
+A PSR is often comprised of text and code, more specifically interfaces. Those interfaces are pieces of code that are released and tagged at a specific moment in time, but the PHP language doesn't stand still... it evolves over time.
 
-This means that those interfaces needs to withstand those evolutions, and sometimes they need to be updated, to leverage new language features that could help better enforce the behaviors exposed in the PSR itself.
-At the same time, a PSR cannot be changed after its release if not with erratas, to avoid that a package that declared compatibility would be de-facto not compatible anymore.    
+This means that those interfaces need to withstand those evolutions, and sometimes they need to be updated, to leverage new language features that could help better enforce the behaviors exposed in the PSR itself.
```
Crell:
```suggestion
This means that those interfaces need to withstand that evolution, and sometimes they need to be updated, to leverage new language features that could help better enforce the behaviors exposed in the PSR itself.
```

bylaws/008-psr-evolution.md
```diff
@@ -19,7 +19,7 @@ A new minor release of a PHP-FIG package containing interfaces for a PSR MUST fo
  * the PHP version constraint of the PHP-FIG package MUST NOT be altered to use newer language features that would create cross-compatibility issues.
  
 A new major release of a PHP-FIG package containing interfaces for a PSR MUST follow the same rules, with this exception:
- * the new major version of the package MAY contain breaking changes if the implementing packages have a reasonable upgrading path, like the possibility of releasing a cross-compatible implementation with the previous releases.
+ * the new major version of the package MAY contain breaking changes if the implementing packages have a reasonable upgrade path, like the possibility of releasing a cross-compatible implementation with the previous releases;
  * the new major version of the package MAY refer to a new, superseding PSR.
```
Crell: What's the intent here?  Would this be, say, psr-21, which is a new PHP 8-centric version of Request/Response messages, which then uses the same namespace as PSR-7?

bylaws/008-psr-evolution.md
ashnazg: Done

bylaws/008-psr-evolution.md
ashnazg: That sounds like a question more directly for @Jean85 's PR, rather than mine.  I'm only doing some English touch-ups of his via my PR.

Jean85: Responding to @Crell, yes it would. Obviously that would mean that the PHP-FIG vision would see that as an upgrade path of PSR-7/15/17/18

bylaws/008-psr-evolution.md
xiazemin: https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#52-author
5.14. @throws is that correct?  if I have catched ,should I still comment on it ?

PR #1197, opened 2019-11-04
@copyleft as near-synonym for @copyright

pecuchet:
A proposal to allow the usage of `@copyleft` to more accurately denote reciprocal [copyleft software licenses](https://en.wikipedia.org/wiki/Copyleft) included in a code base.

### DIFF
--- a/proposed/phpdoc-tags.md
+++ b/proposed/phpdoc-tags.md
@@ -16,22 +16,23 @@ PSR-19: PHPDoc tags
 - [5. Tags](#5-tags)
   - [5.1.  @api](#51-api)
   - [5.2.  @author](#52-author)
-  - [5.3.  @copyright](#53-copyright)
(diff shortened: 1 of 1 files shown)


ashnazg:
I am a copyleft fan... but I don't think this warrants its own tag, since a copyleft designation can be reasonably interpreted from the existing copyright and/or license tags.

@JanTvrdik @muglug , any thoughts?

Crell:
I concur.  copyleft is not a legal term.  What you really want here is `@license GPLv2-or-later` or something like that.  Just `@copyleft` doesn't actually tell you anything useful.

phackwer:
Yes, you should always be specific in licensing. Public domain/copyleft
leads to a number of problems and possibilities. You should never leave it
lightly defined as the use of the term "free" is too subjective.

PR #1203, opened 2020-01-16
Update reference to PSR-2 to PSR-12

skylerfenn:
PSR-2 is now deprecated.


Jean85:
This is advisable, but since it's a `MUST`, we cannot merge it due to the fact that it's not backward compatible, and it would possibly break the spec for an implementing library.

benedmunds:
Bumping this @Jean85

Jean85:
Wow, damn be the pandemic, I definitely dropped this!

I'll call the vote.

skylerfenn:
> There's a typo in this but the vote has now started - there shouldn't be a comma after 'and'

Fixed!

samdark (approved):

yuparach99 (approved):

Jean85 (approved):
Sorry I'm VERY late!

Both votes are passed! :+1: 

https://groups.google.com/g/php-fig/c/GK-fjLdQFBk
https://groups.google.com/g/php-fig/c/FfJcCx4nHx0

PR #1206, opened 2020-01-24
Propose errata for anonymous classes

Jean85:
Reference ML discussion: https://groups.google.com/d/topic/php-fig/F0IVXx1LL_Q/discussion

### DIFF
--- a/accepted/PSR-12-extended-coding-style-guide-meta.md
+++ b/accepted/PSR-12-extended-coding-style-guide-meta.md
@@ -212,5 +212,4 @@ _**Note:** Order descending chronologically._
 
 # 9. Errata and clarifications
 
- * The code example for anonymous classes has ben fixed (see [#1206](https://github.com/php-fig/fig-standards/pull/1206)) because it was wrong; omitting parentheses after the `class` keyword should not be allowed, since it would be in contrast with what is prescribed for normal classes.
-
+ * The code example for anonymous classes was fixed (see [#1206](https://github.com/php-fig/fig-standards/pull/1206)) because it contradicted the actual rule: omitting parentheses after the `class` keyword when creating a new instance is not allowed, as prescribed for normal classes.
--- a/accepted/PSR-12-extended-coding-style-guide-meta.md
+++ b/accepted/PSR-12-extended-coding-style-guide-meta.md
@@ -209,3 +209,8 @@ _**Note:** Order descending chronologically._
 
 * [Inspiration Mailing List Thread](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/php-fig/wh9avopSR9k)
 * [Initial Mailing List PSR Proposal Thread](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/php-fig/MkFacLdfGso)
+
+# 9. Errata and clarifications
+
+ * The code example for anonymous classes has ben fixed (see [#1206](https://github.com/php-fig/fig-standards/pull/1206)) because it was wrong; omitting parentheses after the `class` keyword should not be allowed, since it would be in contrast with what is prescribed for normal classes.
+
--- a/accepted/PSR-12-extended-coding-style-guide.md
+++ b/accepted/PSR-12-extended-coding-style-guide.md
@@ -232,7 +232,7 @@ declare(ticks=1) {
 
 ## 4. Classes, Properties, and Methods
 
-The term "class" refers to all classes, anonymous classes, interfaces, and traits.
+The term "class" refers to all classes, interfaces, and traits.
 
 Any closing brace MUST NOT be followed by any comment or statement on the
 same line.
--- a/accepted/PSR-12-extended-coding-style-guide.md
+++ b/accepted/PSR-12-extended-coding-style-guide.md
@@ -232,7 +232,7 @@ declare(ticks=1) {
 
 ## 4. Classes, Properties, and Methods
 
-The term "class" refers to all classes, interfaces, and traits.
+The term "class" refers to all classes, anonymous classes, interfaces, and traits.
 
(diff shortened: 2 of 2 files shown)


accepted/PSR-12-extended-coding-style-guide.md
```diff
@@ -232,7 +232,7 @@ declare(ticks=1) {
 
 ## 4. Classes, Properties, and Methods
 
-The term "class" refers to all classes, interfaces, and traits.
+The term "class" refers to all classes, anonymous classes, interfaces, and traits.
```
GrahamCampbell: I don't think this is right, because the position of the opening brace for genuine classes is different to that of anonymous classes.

mvorisek: PR is legit, but I would also remove this part which only make thing more confusing by recursive definition.

Jean85:
I've added an errata paragraph, but I'm not to happy with that; any suggestion on how to rephrase that?

accepted/PSR-12-extended-coding-style-guide-meta.md
```diff
@@ -209,3 +209,8 @@ _**Note:** Order descending chronologically._
 
 * [Inspiration Mailing List Thread](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/php-fig/wh9avopSR9k)
 * [Initial Mailing List PSR Proposal Thread](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/php-fig/MkFacLdfGso)
+
+# 9. Errata and clarifications
+
+ * The code example for anonymous classes has ben fixed (see [#1206](https://github.com/php-fig/fig-standards/pull/1206)) because it was wrong; omitting parentheses after the `class` keyword should not be allowed, since it would be in contrast with what is prescribed for normal classes.
```
samdark:
```suggestion
 * The code example for anonymous classes was fixed (see [#1206](https://github.com/php-fig/fig-standards/pull/1206)) because it contradicted actual rule: omitting parentheses after the `class` keyword when creating a new instance should not be allowed as prescribed for normal classes.
```

PowerKiKi: I suggest using "is" instead of "should", because the original rule is a "MUST", not a "SHOULD".

```suggestion
 * The code example for anonymous classes has ben fixed (see [#1206](https://github.com/php-fig/fig-standards/pull/1206)) because it was wrong; omitting parentheses after the `class` keyword is not allowed, since it would be in contrast with what is prescribed for normal classes.
```

Jean85:
I've committed a mix of @samdark & @PowerKiKi suggestions.

mvorisek:
@Jean85 Ready to be merged?

Jean85:
This needs to go through a Core Committee vote to be approved.

mvorisek:
@Jean85 Is that already scheduled and when the outcome can be excepted?

Jean85:
It's up to @KorvinSzanto as the editor of the PSR to approve and call the vote on this.

mbessolov:
Hello everyone,

While it all looks nice and tide without any constructor arguments, once they get introduced into the example, it goes haywire.

> 4.1 Extends and Implements
> The extends and implements keywords MUST be declared on the same line as the class name.

Let's see how this plays with constructor arguments present:

```
$instance = new class($firstConstructorArgument, $secondArgument, $thirdArgument, $andOneMore) extends \Foo implements \HandleableInterface {
```

or as many of us work with object properties and getters, it aggravates the problem further:
```
class Sometest {
    public function testSomething()
    {
        $instance = new class($this->getSomething(), $this->somethingElse, $this->andTheThirdProperty(), $this->letsBreakDemeterLawToo()->getX()) extends \Foo implements \HandleableInterface {
```

While PSR-2 imposes only a *soft* limit of 120 characters on line length, and there is stricly speaking no problem from the PSR point of view, I think that enforcing `extends` to be on the same line as `new class()` makes it barely readable (in the first example the line length is 142 characters, and in the second example it is 192 characters including indentation).

In my opinion a clarification is needed that `class` in `new class() extends` is not really "the class name" (they are anonymous classes after all, not having names by definition) and extends and implements keywords may be on the line with the closing parenthesis of a multi-line list of constructor arguments like so:
```
$instance = new class(
    $foo,
    $bar
) extends \Foo implements \HandleableInterface {
```

Could this last code snippet be included as an example in addition to/instead of the current one?

drupol:
What's the status of this ?

Jean85:
There's noone pushing those kind of changes.

The main problem here is that we do not have a way to handle these changes in an iterating way: the language is evolving really fast (enum, union types, matches are just some of the new stuff) and the PSR process is "granitic", since a PSR is supposed to be immutable. We would need a "living document" kind of process, not the simple PSR way as of now, but we're out of ideas for now.

If someone has some idea on how to proceed, it's welcome to discuss that with us, so we can try to define a formal process and make these kind of adjustments easy.

WinterSilence:
@Jean85 standard is immutables, but extendable (for example, RFC standards).
i.e. PSR-12.1 add rules rules anonymous classes, PSR-12.2 add rules for ENUM and Annotations and etc.

Crell:
There is no "extendable with sub-spec" process for PSRs.  That's not a thing.

However, we do now have a PER process for evolving standards, and there is a working group for a Coding Standards PER to supplant PSR-12.  So that's happening, just not here.

Closing this issue accordingly.

WinterSilence:
@Crell
> PER process for evolving standards

where I can read more about it?

Crell:
https://www.php-fig.org/bylaws/per-workflow/

WinterSilence:
@Crell voting to modernize deprecated code looks like strange

PR #1207, opened 2020-01-26
Use github code owners instead of pullappprove

goetas:
pullapprove is a paid service now, github code owners offers a similar feature


ashnazg:
The decision to change LGTM

CODEOWNERS
```diff
@@ -0,0 +1,4 @@
+/proposed/ @php-fig/secretaries
+/bylaws/ @php-fig/secretaries
+/proposed/security* @michaelcullum @php-fig/psr-9-10
```
Jean85: Michael is included in that team, so it shouldn't be needed here

```suggestion
/proposed/security* @php-fig/psr-9-10
```

michaelcullum:
Should also probably have a catch-all `*` for secretaries. Shame code owners doesn't work for labels. That was nice.

Ah well.

CODEOWNERS
```diff
@@ -0,0 +1,4 @@
+/proposed/ @php-fig/secretaries
```
michaelcullum:
```suggestion
* @php-fig/secretaries
/proposed/ @php-fig/secretaries
```

goetas:
suggested changes applied

goetas:
Then it is!

PR #1209, opened 2020-03-13
Fix spelling mistake in PSR-3-logger-interface.md

jhfrintrop:
Shouldn't it be `[...] can be cast [...]` instead of `[...] can be casted [...]`?


Crell:
I believe `cast` is more correct in English, yes.  This is also a non-substantive change in just a comment in non-normative code, so I think it's safe to make.

PR #1212, opened 2020-04-20
CacheInterface::getOrSet()

ElGigi:
Fetches a value from the cache, if does not exists, persists data of callback result in the cache, uniquely referenced by a key with an optional expiration TTL time.


Crell:
While a useful feature, the spec deliberately chose to avoid it to keep the API surface simple.  PSR-6 is already passed and thus cannot take new features, only trivial clarifications.

PR #1213, opened 2020-05-26
Return type hint compatible with code analyzers

Wosi:
Common code analyzers such as PhpStan and Psalm don't understand the notation
`@return iterable[callable]`
But they work as expected with
`@return iterable<callable>`

(A corresponding PR in the composer package repository has already been merged (https://github.com/php-fig/event-dispatcher/pull/25) 
@Crell asked me to create this PR


goetas:
Thanks @Wosi

PR #1214, opened 2020-06-16
Use "Structural Elements" in reference @internal

jaapio:
API components is a new definition which does not allign with the rest of the document where we are using "Structural Elements"


jaapio:
@ashnazg :-)

ashnazg (approved):

PR #1217, opened 2020-07-18
Add TYPO3 project

goetas:
As the result of https://groups.google.com/d/msg/php-fig/QWcR0RU34rs/w5IWT2t2AwAJ TYPO3 is now a project member of PHP-FIG
/cc @bmack


SlvrEagle23:
@goetas Shouldn't this PR also specify where the `[Typo3]` link goes? As far as I can tell it won't resolve as is.

goetas:
@SlvrEagle23  good catch! Updated!

bmack:
Thanks a lot for pushing it forward. One slight remark / question. Since the product is called "TYPO3" not and not "Typo3", can we update that in the Project Name (similar to SugarCRM or CakePHP where we have some uppercase mixed in the game)... is that OK?

goetas:
@bmack done

goetas:
thank you all! TYPO3 Welcome onboard!

bmack:
Thank you all for having me/us on board!

PR #1219, opened 2020-12-02
Move Beau Simensen to Former Core Committee; Add Vacancy

SlvrEagle23:


personnel.md
```diff
@@ -17,7 +17,7 @@ Feel free to contact the secretaries at info AT php-fig.org. For more informatio
 
 | Name                                  | Term                    |
 |---------------------------------------|-------------------------|
-| Beau Simensen ([@beausimensen])       | 2016-12-24 - 2020-08-31 |
+| Vacant                                |                         |
 | Larry Garfield ([@Crell])             | 2016-12-24 - 2021-05-30 |
 | Matthew Weier O'Phinney ([@mwop])     | 2016-12-24 - 2021-05-30 |
 | Cees-Jan Kiewiet ([@WyriHaximus])     | 2016-12-24 - 2020-08-31 |
```
Jean85, line 23: We should also update the term of the CC members that got reelected.

Jean85 (changes requested):

personnel.md
SlvrEagle23: @Jean85 When did that re-election happen and what is their term extended to?

Jean85: As per our bylaws https://www.php-fig.org/bylaws/elections-and-vacancies/
> Both Secretaries and Core Committee members are elected for two year terms with votes staggered eight months apart.
> [...]
> The new Secretary and Core Committee members will take office at 5pm UTC time on the last Sunday of the month, and the voting period must conclude at least 24 hours prior to that.

SlvrEagle23: @Jean85 Updated terms, moved inactive members and added Ben Edmunds.

Jean85 (approved):

PR #1228, opened 2021-04-08
A propsal for enums

ahmadmayahi:


accepted/PSR-1-basic-coding-standard.md
```diff
@@ -97,14 +97,14 @@ if (! function_exists('bar')) {
 }
 ~~~
 
-## 3. Namespace and Class Names
+## 3. Namespace, Class Names and Enums
```
drealecs, line 100: I think the only needed change is in section 4.:
Changing
```
The term "class" refers to all classes, interfaces, and traits.
```
to
```
The term "class" refers to all classes, interfaces, enums and traits.
```
Or something similar.

drealecs: But we can also consider that enum is just a special type of class. So no changes would be needed.

Crell:
We don't modify already-approved specs.  But as @drealecs noted, class includes enums for most practical purposes.

(Except for the brace-on-own-line bit, because that's intensely stupid for classes already and makes even less sense on enums, but that's a different matter.)

Jean85:
Closing this as already explained by Crell.

PR #1230, opened 2021-04-14
Modify meta doc

heiglandreas:
This PR contains some changes that should make it more clear why certain decissions were taken the way we took them.

### DIFF
--- a/proposed/clock-meta.md
+++ b/proposed/clock-meta.md
@@ -81,25 +81,6 @@ final class SystemClock implements \Psr\Clock\ClockInterface
     }
 }
 
-//
-
-final class UTCClock implements \Psr\Clock\ClockInterface
-{
-    private \DateTimeZone $utcTimeZone;
-
-    public function __construct()
-    {
-        $this->utcTimeZone = new \DateTimeZone('UTC');
-    }
-
-    public function now(): \DateTimeImmutable
-    {
-        return new \DateTimeImmutable('now', $this->utcTimeZone);
-    }
-}
-
-//
-
 final class FrozenClock implements \Psr\Clock\ClockInterface
 {
     private \DateTimeImmutable $now;
--- a/proposed/clock-meta.md
+++ b/proposed/clock-meta.md
@@ -66,7 +66,7 @@ Time by now is defined by interaction of electromagnetic radiation with the exci
 
 This means that for the sake of this PSR the timezone is irrelevant as it is an implementation detail of the code implementing the interface. 
 
-As there is no way to assert a certain timezone in the returnd `DateTimeImmutable` it is up to the implementation to make sure that the timezone is handled according to the Business-Logic. That is either by making sure that a call to `now()` will only return a `DateTimeImmutable`-object with a known timezone (implicit contract) or by making sure that the Business-logic handles the timezone explicitly. This an be done by calling `setTimezone` or by using `getTimestamp` which will automatically return the UTC based Unix-Timestamp.
+It is up to the implementation to make sure that the timezone is handled according to the business logic of the application. That is either by making sure that a call to `now()` will only return a `DateTimeImmutable` object with a known timezone (implicit contract) or by explicitly changing the timezone to be correct for the application. This an be done by calling `setTimezone()` to create a new `DateTimeImmutable` object with the given timezone. 
 
 These are though actions that are not part of the contract which this interface defines.
 
--- a/proposed/clock-meta.md
+++ b/proposed/clock-meta.md
@@ -6,7 +6,7 @@ Getting the current time in applications is a rather straight forward thing give
 
 Due to the nature of time as constantly progressing it becomes a nuisance to use these functions if you need predictable results like for testing.
 
-This `ClockInterface` aims to provide a standard way to consume time, that allows interopability not only when consuming the "real" time but also when predictable results need to be available. This allows to skip using PHP Extensions for testing or using hacks like re-declaring the `time()` function in other namespaces. 
+This `ClockInterface` aims to provide a standard way to consume time, that allows interoperability not only when consuming the "real" time but also when predictable results need to be available. This allows the avoidance of using PHP Extensions for testing or using hacks like re-declaring the `time()` function in other namespaces. 
 
 ## 2. Why Bother?
 
--- a/proposed/clock-meta.md
+++ b/proposed/clock-meta.md
@@ -54,7 +54,6 @@ We have decided to formalize the existing practices, used by several other packa
 
 (This list is not exhaustive!)
 
-Some providing
 interfaces, and some relying on overloading (extending) the Clock class to mock the
 current time.
 
--- a/proposed/clock-meta.md
+++ b/proposed/clock-meta.md
@@ -2,7 +2,7 @@
 
 ## 1. Summary
 
-Getting the current time in applications is a rather straightforward thing given the language's native `time()` function (or one of its similar functionalities like `utime()` or `new DateTimeImmutable()` etc.)
+Getting the current time in applications is a rather straightforward thing given the language's native `time()` function (or one of its similar functionalities like `microtime()` or `new DateTimeImmutable()` etc.)
 
 Due to the nature of time as constantly progressing it becomes a nuisance to use these functions if you need predictable results, for example when testing.
 
--- a/proposed/clock-meta.md
+++ b/proposed/clock-meta.md
@@ -14,7 +14,7 @@ There are currently a few libraries that provide this functionality, however the
 
 Symfony provides a package called `symfony/phpunit-bridge` that has a `Symfony\Bridge\PhpUnit\ClockMock` class, which allows mocking PHP's built-in time and date functions, however this does not solve mocking calls to `new \DateTimeImmutable()`. It also does not fully mock time when called from other libraries that rely on the system time.
 
-`Carbon\Carbon` – and being forked from that also `Cake\Chronos\Chronos` – do provide mocking via a static `setTestNow()` method, but this has its own pitfalls as it provides no isolation and must be called again to stop mocking.
+`Carbon\Carbon`, and its fork `Cake\Chronos\Chronos`, do provide mocking via a static `setTestNow()` method, but this provides no isolation and must be called again to stop mocking.
 
 Pros:
 
--- a/proposed/clock-meta.md
+++ b/proposed/clock-meta.md
@@ -4,7 +4,7 @@
 
 Getting the current time in applications is a rather straight forward thing given the languages native `time()` function (or one of it's similar functionalities like `utime()` or `new DateTimeImmutable()` etc.)
 
-Due to the nature of time as constantly progressing it becomes a nuisance to use these functions if you need predictable results like for testing.
+Due to the nature of time as constantly progressing it becomes a nuisance to use these functions if you need predictable results, for example when testing.
 
 This `ClockInterface` aims to provide a standard way to consume time, that allows interoperability not only when consuming the "real" time but also when predictable results need to be available. This allows the avoidance of using PHP Extensions for testing or using hacks like re-declaring the `time()` function in other namespaces. 
 
--- a/proposed/clock-meta.md
+++ b/proposed/clock-meta.md
@@ -64,7 +64,7 @@ These implementations all provide a `now()` method which returns a `DateTimeImmu
 
 Time by now is defined by interaction of electromagnetic radiation with the excited states of certain atoms where the SI defines one second as the duration of 9192631770 cycles of radiation corresponding to the transition between two energy levels of the ground state of the caesium-133 atom at 0K. This means that retrieving the current time will always return the same time, no matter where it is observed. While the timezone defines *where* the time was observed it does not modify the actual "slice" of time.
 
-This means that for the sake of this PSR the timezone is irrelevant as it is an implementation detail of the code implementing the interface. 
+This means that for the sake of this PSR the timezone is considered an implementation detail of the interface. 
 
 It is up to the implementation to make sure that the timezone is handled according to the business logic of the application. That is either by making sure that a call to `now()` will only return a `DateTimeImmutable` object with a known timezone (implicit contract) or by explicitly changing the timezone to be correct for the application. This an be done by calling `setTimezone()` to create a new `DateTimeImmutable` object with the given timezone. 
 
--- a/proposed/clock-meta.md
+++ b/proposed/clock-meta.md
@@ -39,7 +39,7 @@ calling `time()` or `date()`.
   described in this document, so it is not a coding standard;
 * This PSR does not provide a recommendation on how to handle timezones when 
   retrieving the current time. This is left up to the implementation.
-* This PSR does not handle any Scheduler specific tasks like `sleep` or `wait` as those – even though they take a time or an interval as parameter – are not related to retrieving the current time. Handling such tasks is left to the respective implementations or a different PSR.
+* This PSR does not handle any scheduling methods like `sleep()` or `wait()` because such methods are not related to retrieving the current time.
 
 ## 4. Approaches
 
--- a/proposed/clock-meta.md
+++ b/proposed/clock-meta.md
@@ -67,7 +67,7 @@ This means that for the sake of this PSR the timezone is considered an implement
 
 It is up to the implementation to make sure that the timezone is handled according to the business logic of the application. That is either by making sure that a call to `now()` will only return a `DateTimeImmutable` object with a known timezone (implicit contract) or by explicitly changing the timezone to be correct for the application. This can be done by calling `setTimezone()` to create a new `DateTimeImmutable` object with the given timezone. 
 
-These are though actions that are not part of the contract which this interface defines.
+These actions are not defined in this interface.
 
 
 ### 4.2 Example Implementations
--- a/proposed/clock-meta.md
+++ b/proposed/clock-meta.md
@@ -2,21 +2,19 @@
 
 ## 1. Summary
 
-The purpose of using the `ClockInterface` is to provide a standard way to access the system 
-time, that would allow interopability when testing code that relies on the current time 
-rather than relying on installing PHP extensions or use hacks like re-declaring the `time()`
-function in other namespaces. 
+Getting the current time in applications is a rather straight forward thing given the languages native `time()` function (or one of it's similar functionlaities like `utime()` or `new DateTimeImmutable()` etc.)
+
+Due to the nature of time as constantly progressing it becomes a nuisance to use these functions if you need predictable results like for testing.
+
+This `ClockInterface` aims to provide a standard way to consume time, that allows interopability not only when consuming the "real" time but also when predictable results need to be available. This allows to skip using PHP Extensions for testing or using hacks like re-declaring the `time()` function in other namespaces. 
 
 ## 2. Why Bother?
 
-There are currently a few libraries that provide this functionality, however 
-there is no interopability between these different libraries, as they ship with their own 
-clock interfaces. Symfony provides a package called `symfony/phpunit-bridge` that has a
-`Symfony\Bridge\PhpUnit\ClockMock` class, which allows mocking PHP's built-in time & date 
-functions, however this does not solve mocking calls to `new \DateTimeImmutable()`. It does
-not fully mock time when called from other libraries that rely on the system time. 
-`Cake\Chronos\Chronos` does provide mocking, however it is set via a global (static class 
-property), and this has its own pitfalls as it provides no isolation.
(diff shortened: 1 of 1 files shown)


proposed/clock-meta.md
```diff
@@ -41,17 +39,37 @@ calling `time()` or `date()`.
...
+* [`ergebnis/clock`](https://packagist.org/packages/ergebnis/clock)
+* [`mangoweb/clock`](https://packagist.org/packages/mangoweb/clock)
+
+(This list is not exhaustive!)
+
+Some providing
 interfaces, and some relying on overloading (extending) the Clock class to mock the
 current time.
 
+These implementations all provide a `now`-method which returns a `DateTimeImmutable`-object. As the `DateTimeImmutable`-object allows to retieve the Unix-Timestamp by calling `getTimestamp()` or `format('u.U')` this interface does not define any special methods to retieve a Unix-Timestamp or any other form of time-information apart from retrieving the `DateTimeImmutable`-object.
```
shadowhand, line 61:
```suggestion
These implementations all provide a `now()` method which returns a `DateTimeImmutable` object. As the `DateTimeImmutable` object allows retrieving the Unix timestamp, by calling `getTimestamp()` or `format('u.U')`, this interface does not define any special methods to retrieve a Unix timestamp or any other time information that is not available from a `DateTimeImmutable` object. 
```

proposed/clock-meta.md
```diff
@@ -41,17 +39,37 @@ calling `time()` or `date()`.
...
+
+Time by now is defined by interaction of electromagnetic radiation with the excited states of certain atoms where the SI defines one second as the duration of 9192631770 cycles of radiation corresponding to the transition between two energy levels of the ground state of the caesium-133 atom at 0K. This means that retrieving the current time will always return the same time, no matter where it is observed. While the timezone defines *where* the time was observed it does not modify the actual "slice" of time.
+
+This means that for the sake of this PSR the timezone is irrelevant as it is an implementation detail of the code implementing the interface. 
+
+As there is no way to assert a certain timezone in the returnd `DateTimeImmutable` it is up to the implementation to make sure that the timezone is handled according to the Business-Logic. That is either by making sure that a call to `now()` will only return a `DateTimeImmutable`-object with a known timezone (implicit contract) or by making sure that the Business-logic handles the timezone explicitly. This an be done by calling `setTimezone` or by using `getTimestamp` which will automatically return the UTC based Unix-Timestamp.
```
shadowhand, line 69:
```suggestion
It is up to the implementation to make sure that the timezone is handled according to the business logic of the application. That is either by making sure that a call to `now()` will only return a `DateTimeImmutable` object with a known timezone (implicit contract) or by explicitly changing the timezone to be correct for the application. This an be done by calling `setTimezone()` to create a new `DateTimeImmutable` object with the given timezone. 
```

proposed/clock-meta.md
```diff
@@ -41,17 +39,37 @@ calling `time()` or `date()`.
...
 interfaces, and some relying on overloading (extending) the Clock class to mock the
 current time.
 
+These implementations all provide a `now`-method which returns a `DateTimeImmutable`-object. As the `DateTimeImmutable`-object allows to retieve the Unix-Timestamp by calling `getTimestamp()` or `format('u.U')` this interface does not define any special methods to retieve a Unix-Timestamp or any other form of time-information apart from retrieving the `DateTimeImmutable`-object. 
+
+### 4.2 Timezones
+
+Time by now is defined by interaction of electromagnetic radiation with the excited states of certain atoms where the SI defines one second as the duration of 9192631770 cycles of radiation corresponding to the transition between two energy levels of the ground state of the caesium-133 atom at 0K. This means that retrieving the current time will always return the same time, no matter where it is observed. While the timezone defines *where* the time was observed it does not modify the actual "slice" of time.
+
+This means that for the sake of this PSR the timezone is irrelevant as it is an implementation detail of the code implementing the interface.
```
shadowhand, line 67:
```suggestion
This means that for the sake of this PSR the timezone is considered an implementation detail of the interface. 
```

proposed/clock-meta.md
```diff
@@ -41,17 +39,37 @@ calling `time()` or `date()`.
...
+
+This means that for the sake of this PSR the timezone is irrelevant as it is an implementation detail of the code implementing the interface. 
+
+As there is no way to assert a certain timezone in the returnd `DateTimeImmutable` it is up to the implementation to make sure that the timezone is handled according to the Business-Logic. That is either by making sure that a call to `now()` will only return a `DateTimeImmutable`-object with a known timezone (implicit contract) or by making sure that the Business-logic handles the timezone explicitly. This an be done by calling `setTimezone` or by using `getTimestamp` which will automatically return the UTC based Unix-Timestamp.
+
+These are though actions that are not part of the contract which this interface defines.
```
shadowhand, line 71:
```suggestion
```

proposed/clock-meta.md
```diff
@@ -64,25 +82,6 @@ final class SystemClock implements \Psr\Clock\ClockInterface
...
-    public function __construct()
-    {
-        $this->utcTimeZone = new \DateTimeZone('UTC');
-    }
-
-    public function now(): \DateTimeImmutable
-    {
-        return new \DateTimeImmutable('now', $this->utcTimeZone);
-    }
-}
```
shadowhand, line 82: Why was this removed?

proposed/clock-meta.md
```diff
@@ -41,17 +39,37 @@ calling `time()` or `date()`.
   described in this document, so it is not a coding standard;
 * This PSR does not provide a recommendation on how to handle timezones when 
   retrieving the current time. This is left up to the implementation.
+* This PSR does not handle any Scheduler specific tasks like `sleep` or `wait` as those – even though they take a time or an interval as parameter – are not related to retrieving the current time. Handling such tasks is left to the respective implementations or a different PSR.
```
shadowhand, line 42:
```suggestion
* This PSR does not handle any scheduling methods like `sleep()` or `wait()` because such methods are not related to retrieving the current time.
```

proposed/clock-meta.md
```diff
@@ -2,21 +2,19 @@
...
-`Symfony\Bridge\PhpUnit\ClockMock` class, which allows mocking PHP's built-in time & date 
-functions, however this does not solve mocking calls to `new \DateTimeImmutable()`. It does
-not fully mock time when called from other libraries that rely on the system time. 
-`Cake\Chronos\Chronos` does provide mocking, however it is set via a global (static class 
-property), and this has its own pitfalls as it provides no isolation.
+There are currently a few libraries that provide this functionality, however there is no interopability between these different libraries, as they ship with their own clock interfaces. 
+
+Symfony provides a package called `symfony/phpunit-bridge` that has a
+`Symfony\Bridge\PhpUnit\ClockMock` class, which allows mocking PHP's built-in time & date functions, however this does not solve mocking calls to `new \DateTimeImmutable()`. It does not fully mock time when called from other libraries that rely on the system time. 
+`Cake\Chronos\Chronos` does provide mocking, however it is set via a global (static class property), and this has its own pitfalls as it provides no isolation.
```
shadowhand, line 17:
```suggestion
Symfony provides a package called `symfony/phpunit-bridge` that has a `Symfony\Bridge\PhpUnit\ClockMock` class, which allows mocking PHP's built-in time and date functions, however this does not solve mocking calls to `new \DateTimeImmutable()`. It also does not fully mock time when called from other libraries that rely on the system time.

`Cake\Chronos\Chronos` does provide mocking via a static `setTestNow()` method, but this has its own pitfalls as it provides no isolation and must be called again to stop mocking.
```

shadowhand (changes requested):

proposed/clock-meta.md
heiglandreas: If we are proposing that the timezone is not part of the Interface we should not provide a minimum implementation that provides a timezone specific implementation. And for a minimum implementation the UTCClock is not necessary. If people want a more sophisticated example they can check out one of the multiple provided implementations

proposed/clock-meta.md
```diff
@@ -2,21 +2,19 @@
 
 ## 1. Summary
 
-The purpose of using the `ClockInterface` is to provide a standard way to access the system 
-time, that would allow interopability when testing code that relies on the current time 
-rather than relying on installing PHP extensions or use hacks like re-declaring the `time()`
-function in other namespaces. 
+Getting the current time in applications is a rather straight forward thing given the languages native `time()` function (or one of it's similar functionlaities like `utime()` or `new DateTimeImmutable()` etc.)
```
benedmunds, line 5:
```suggestion
Getting the current time in applications is a rather straight forward thing given the languages native `time()` function (or one of it's similar functionalities like `utime()` or `new DateTimeImmutable()` etc.)
```

proposed/clock-meta.md
```diff
@@ -2,21 +2,19 @@
...
+
+Due to the nature of time as constantly progressing it becomes a nuisance to use these functions if you need predictable results like for testing.
```
benedmunds, line 7:
```suggestion
Due to the nature of time as constantly progressing it becomes a nuisance to use these functions if you need predictable results, for example when testing.
```

proposed/clock-meta.md
```diff
@@ -2,21 +2,19 @@
...
+
+This `ClockInterface` aims to provide a standard way to consume time, that allows interopability not only when consuming the "real" time but also when predictable results need to be available. This allows to skip using PHP Extensions for testing or using hacks like re-declaring the `time()` function in other namespaces.
```
benedmunds, line 9:
```suggestion
This `ClockInterface` aims to provide a standard way to consume time, that allows interoperability not only when consuming the "real" time but also when predictable results need to be available. This allows the avoidance of using PHP Extensions for testing or using hacks like re-declaring the `time()` function in other namespaces. 
```

proposed/clock-meta.md
```diff
@@ -41,17 +39,37 @@ calling `time()` or `date()`.
...
+We have decided to formalize the existing practices, used by several other packages out in the wild. Some of the popular packages providing this functionality are: 
+
+* [`lcobucci/clock`](https://packagist.org/packages/lcobucci/clock)
+* [`kreait/clock`](https://packagist.org/packages/kreait/clock)
+* [`ergebnis/clock`](https://packagist.org/packages/ergebnis/clock)
+* [`mangoweb/clock`](https://packagist.org/packages/mangoweb/clock)
+
+(This list is not exhaustive!)
+
+Some providing
```
benedmunds, line 57:
```suggestion
```

benedmunds (changes requested):

proposed/clock-meta.md
ashnazg, line 5: s/straight forward/straightforward/ (it's one word)

s/given the languages native/given the language's native/ (make it possessive)

s/one of it's similar/one of its similar/ (make it possessive rather than "it is" contraction)

heiglandreas: Fixed with f5c2188

proposed/clock-meta.md
heiglandreas: We should make it explicit that handling the timezones are not part of the interfaces contract!

cseufert:
Looks great thanks @heiglandreas

heiglandreas:
Anything I can do  to go on with this PSR?

proposed/clock-meta.md
```diff
@@ -2,21 +2,19 @@
 
 ## 1. Summary
 
-The purpose of using the `ClockInterface` is to provide a standard way to access the system 
-time, that would allow interopability when testing code that relies on the current time 
-rather than relying on installing PHP extensions or use hacks like re-declaring the `time()`
-function in other namespaces. 
+Getting the current time in applications is a rather straightforward thing given the language's native `time()` function (or one of its similar functionalities like `utime()` or `new DateTimeImmutable()` etc.)
```
nicolas-grekas, line 5: s/utime/microtime I suppose?

heiglandreas: Good catch! Thanks. Mixed up the underlying C-functions there 😁

proposed/clock-meta.md
```diff
@@ -2,21 +2,19 @@
...
-`Symfony\Bridge\PhpUnit\ClockMock` class, which allows mocking PHP's built-in time & date 
-functions, however this does not solve mocking calls to `new \DateTimeImmutable()`. It does
-not fully mock time when called from other libraries that rely on the system time. 
-`Cake\Chronos\Chronos` does provide mocking, however it is set via a global (static class 
-property), and this has its own pitfalls as it provides no isolation.
+There are currently a few libraries that provide this functionality, however there is no interopability between these different libraries, as they ship with their own clock interfaces. 
+
+Symfony provides a package called `symfony/phpunit-bridge` that has a `Symfony\Bridge\PhpUnit\ClockMock` class, which allows mocking PHP's built-in time and date functions, however this does not solve mocking calls to `new \DateTimeImmutable()`. It also does not fully mock time when called from other libraries that rely on the system time.
+
+`Cake\Chronos\Chronos` does provide mocking via a static `setTestNow()` method, but this has its own pitfalls as it provides no isolation and must be called again to stop mocking.
```
kylekatarnls, line 17: `Chronos` just inherits `setTestNow()` from `Carbon`, the project they forked, which implemented it years before. By respect for @briannesbitt, the original author, please mention `Carbon` instead.

proposed/clock-meta.md
```diff
@@ -41,17 +39,36 @@ calling `time()` or `date()`.
...
+We have decided to formalize the existing practices, used by several other packages out in the wild. Some of the popular packages providing this functionality are: 
+
+* [`lcobucci/clock`](https://packagist.org/packages/lcobucci/clock)
+* [`kreait/clock`](https://packagist.org/packages/kreait/clock)
+* [`ergebnis/clock`](https://packagist.org/packages/ergebnis/clock)
+* [`mangoweb/clock`](https://packagist.org/packages/mangoweb/clock)
+
+(This list is not exhaustive!)
+
 interfaces, and some relying on overloading (extending) the Clock class to mock the
```
kylekatarnls, line 57: "Some providing" got truncated.
```suggestion
Some providing interfaces, and some relying on overloading (extending) the Clock class to mock the
```

proposed/clock-meta.md
```diff
@@ -41,17 +39,36 @@ calling `time()` or `date()`.
...
+
+### 4.2 Timezones
+
+Time by now is defined by interaction of electromagnetic radiation with the excited states of certain atoms where the SI defines one second as the duration of 9192631770 cycles of radiation corresponding to the transition between two energy levels of the ground state of the caesium-133 atom at 0K. This means that retrieving the current time will always return the same time, no matter where it is observed. While the timezone defines *where* the time was observed it does not modify the actual "slice" of time.
+
+This means that for the sake of this PSR the timezone is considered an implementation detail of the interface. 
+
+It is up to the implementation to make sure that the timezone is handled according to the business logic of the application. That is either by making sure that a call to `now()` will only return a `DateTimeImmutable` object with a known timezone (implicit contract) or by explicitly changing the timezone to be correct for the application. This an be done by calling `setTimezone()` to create a new `DateTimeImmutable` object with the given timezone.
```
kylekatarnls, line 68: Yes x 1000! This paragraph is the most important to me. I hope the one above about caesium-133 atom won't loose readers before they reach this 🤣

kylekatarnls (changes requested):
Few typos sound to have truncated word/sentence

proposed/clock-meta.md
heiglandreas: This should now be fixed by mentioning Carboin as well as Chronos

proposed/clock-meta.md
heiglandreas: Thanks! Fixed that

proposed/clock-meta.md
heiglandreas: Good catch! Thanks for spotting!

proposed/clock-meta.md
```diff
@@ -41,17 +39,36 @@ calling `time()` or `date()`.
...
+We have decided to formalize the existing practices, used by several other packages out in the wild. Some of the popular packages providing this functionality are: 
+
+* [`lcobucci/clock`](https://packagist.org/packages/lcobucci/clock)
+* [`kreait/clock`](https://packagist.org/packages/kreait/clock)
+* [`ergebnis/clock`](https://packagist.org/packages/ergebnis/clock)
+* [`mangoweb/clock`](https://packagist.org/packages/mangoweb/clock)
+
+(This list is not exhaustive!)
+
+Some of these providing interfaces, and some relying on overloading (extending) the Clock class to mock the
```
shadowhand, line 57:
```suggestion
Some of these provide interfaces and some rely on extending a clock class to mock the
```

proposed/clock-meta.md
```diff
@@ -2,21 +2,19 @@
 
 ## 1. Summary
 
-The purpose of using the `ClockInterface` is to provide a standard way to access the system 
-time, that would allow interopability when testing code that relies on the current time 
-rather than relying on installing PHP extensions or use hacks like re-declaring the `time()`
-function in other namespaces. 
+Getting the current time in applications is a rather straightforward thing given the language's native `time()` function (or one of its similar functionalities like `microtime()` or `new DateTimeImmutable()` etc.)
```
shadowhand, line 5:
```suggestion
Getting the current time in applications is typically achieved using the `time()` or `microtime` functions, or by using a `new \DateTimeImmutable()` class.
```

proposed/clock-meta.md
```diff
@@ -2,21 +2,19 @@
...
+
+Due to the nature of time as constantly progressing it becomes a nuisance to use these functions if you need predictable results, for example when testing.
```
shadowhand, line 7:
```suggestion
Due to the nature of time progression these methods cannot be used when predictable results are needed, such as during testing.
```

shadowhand (changes requested):

proposed/clock-meta.md
```diff
@@ -2,21 +2,19 @@
...
+
+This `ClockInterface` aims to provide a standard way to consume time, that allows interoperability not only when consuming the "real" time but also when predictable results need to be available. This allows the avoidance of using PHP Extensions for testing or using hacks like re-declaring the `time()` function in other namespaces.
```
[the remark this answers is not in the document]
heiglandreas: In that context rewriting a function in a namespace *is* a hack ;-)

But I do get your point :wink:

proposed/clock-meta.md
```diff
@@ -41,17 +39,36 @@ calling `time()` or `date()`.
...
-interfaces, and some relying on overloading (extending) the Clock class to mock the
+We have decided to formalize the existing practices, used by several other packages out in the wild. Some of the popular packages providing this functionality are: 
+
+* [`lcobucci/clock`](https://packagist.org/packages/lcobucci/clock)
+* [`kreait/clock`](https://packagist.org/packages/kreait/clock)
+* [`ergebnis/clock`](https://packagist.org/packages/ergebnis/clock)
+* [`mangoweb/clock`](https://packagist.org/packages/mangoweb/clock)
+
+(This list is not exhaustive!)
+
```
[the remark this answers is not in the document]
heiglandreas: Even though that should not be necessary I'd rather keep it to make clear that there might be other packages that provide the same value. Yes, we introduced that list with "some popular packages … are". But there will be people not reading that. And the oneliner right underneath the table is catching the eye better than the message in the copy-text above.

proposed/clock-meta.md
```diff
@@ -41,17 +39,36 @@ calling `time()` or `date()`.
...
+### 4.2 Timezones
+
+Time by now is defined by interaction of electromagnetic radiation with the excited states of certain atoms where the SI defines one second as the duration of 9192631770 cycles of radiation corresponding to the transition between two energy levels of the ground state of the caesium-133 atom at 0K. This means that retrieving the current time will always return the same time, no matter where it is observed. While the timezone defines *where* the time was observed it does not modify the actual "slice" of time.
+
+This means that for the sake of this PSR the timezone is considered an implementation detail of the interface. 
+
+It is up to the implementation to make sure that the timezone is handled according to the business logic of the application. That is either by making sure that a call to `now()` will only return a `DateTimeImmutable` object with a known timezone (implicit contract) or by explicitly changing the timezone to be correct for the application. This can be done by calling `setTimezone()` to create a new `DateTimeImmutable` object with the given timezone. 
+
+These are though actions that are not part of the contract which this interface defines.
```
shadowhand, line 70:
```suggestion
These actions are not defined in this interface.
```

shadowhand (approved):
One remaining suggestion, but otherwise looks great!

benedmunds (approved):

lcobucci (approved):

cseufert (approved):

PR #1234, opened 2021-07-01
[PHPDoc] Clarify that @property/@var names are prefixed with a $

caugner:
## Problem

The examples for the `@property` tag all prefix the property name with a sigil ($), even though the specified syntax omits this.

## Solution

To avoid confusion, the sigil ($) should be added explicitly to the syntax specification.

## Context

- PhpStorm's `Add @property` quickfix does omit the sigil, but this behavior is tracked as a bug: https://youtrack.jetbrains.com/issue/WI-34398
- I brought this issue up in https://github.com/vimeo/psalm/issues/6028, because Psalm currently ignores `@property` annotations that omit the sigil.

### DIFF
--- a/proposed/phpdoc-tags.md
+++ b/proposed/phpdoc-tags.md
@@ -583,7 +583,7 @@ The `@property` tag is used to declare which "magic" properties are supported.
 
 #### Syntax
 
-    @property[<-read|-write>] ["Type"] [name] [<description>]
+    @property[<-read|-write>] ["Type"] $[name] [<description>]
 
 #### Description
 
--- a/proposed/phpdoc-tags.md
+++ b/proposed/phpdoc-tags.md
@@ -542,7 +542,7 @@ The @param tag is used to document a single parameter of a function or method.
 
 #### Syntax
 
-    @param ["Type"] [name] [<description>]
+    @param ["Type"] $[name] [<description>]
 
 #### Description
 


samdark (approved):

ashnazg:
As written, the presumption was that "name" equated to the variable name, `$` and all, e.g. `$foo`.

I see that `@method` does explicitly show its parens as literals in the syntax format, so that could hint that where variables are listed, the `$` literal should be shown explicitly in the syntax format.  If we are indeed going to presume that the "name" placeholder means "variable name _without_ `$`", and thus the literal `$` is required to be shown in the syntax format, then we need the same change on the `@param` and `@var` tag sections.

caugner:
@ashnazg 

> As written, the presumption was that "name" equated to the variable name, `$` and all, e.g. `$foo`.

If it was totally clear that `$` is part of the variable name, I guess PhpStorm would have never implemented their `Add @property` quickfix without the `$`. I have been using PHP and PHPDoc for many years and never realised that the `@property` name needs to include the `$` until I started using Psalm and noticed that it didn't pick up those annotations without a `$`. Since the `$` is mandatory for variables, I would regard it as part of the PHP variable syntax rather than part of the variable name itself.

Also, if you look at methods like [compact](https://www.php.net/compact) that accept a `var_name`, the `$` is never part of the variable name. The `$` is not considered part of the variable name for [variable variables](https://www.php.net/manual/en/language.variables.variable.php) either.

> If we are indeed going to presume that the "name" placeholder means "variable name _without_ `$`", and thus the literal `$` is required to be shown in the syntax format, then we need the same change on the `@param` and `@var` tag sections.

That makes sense. Would you like me to add those changes to this PR?

mbeccati:
From experience, I've been hit by this a while ago. I think phpStorm was ok-ish with `@property int foo` while some other tool did complain. Or vice-versa, but big +1 for clarity on my side ;-)

garak (approved):

shadowhand (approved):

caugner:
@vdelau Would you know who could have a look at this and potentially merge?

ashnazg (approved):

caugner:
Thank you so much @ashnazg! 🎉

PR #1251, opened 2021-08-18
Update RFC links for Accepted PSRs

mul91n:
At the time of this pull request, http://tools.ietf.org/html/* is returning 404 Not Found errors.

Considering this subdomain is an alias of levkowetz.com, let's use an official domain listed on the IETF's website: https://www.rfc-editor.org

This pull request composes of a regex find and replace

Search: http(.?)://tools.ietf.org/html/rfc(\d*)
Replace: https://www.rfc-editor.org/rfc/rfc$1.html

It also changes all links to `ietf.org` as https.

This PR does not change links to `ietf.org/rfc/rfcNNNN.txt`.

### DIFF
--- a/accepted/PSR-1-basic-coding-standard.md
+++ b/accepted/PSR-1-basic-coding-standard.md
@@ -8,7 +8,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
 "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
 interpreted as described in [RFC 2119].
 
-[RFC 2119]: http://www.ietf.org/rfc/rfc2119.txt
+[RFC 2119]: https://www.ietf.org/rfc/rfc2119.txt
 [PSR-0]: https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-0.md
 [PSR-4]: https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-4-autoloader.md
 
--- a/accepted/PSR-11-container.md
+++ b/accepted/PSR-11-container.md
@@ -13,7 +13,7 @@ The word `implementor` in this document is to be interpreted as someone
 implementing the `ContainerInterface` in a dependency injection-related library or framework.
 Users of dependency injection containers (DIC) are referred to as `user`.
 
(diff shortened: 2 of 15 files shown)


vdelau:
From my comment in https://github.com/php-fig/fig-standards/pull/1250/#pullrequestreview-732649299:
> Indeed www.rfc-editor.org seems the preferred source, given the referencing guidelines in the RFC style guide: https://www.rfc-editor.org/styleguide/part2/. However, the IETF also suggest direct links to https://www.ietf.org/rfc/rfcNNNN.txt. (see https://www.ietf.org/standards/rfcs/)
>
> I also noticed this change contains changes to source code within the PSR itself. These seem to be to be direct text links, as per IETF recommendation. I can imagine we leave those in place.
>
> Process wise, I believe secretaries can accept the changes for accepted PSR's. However, proposed PSR's should be changed by the respective editors. Could you remove the proposed PSR's from this PR, and split them of into individual PR's? Or just file issues against each respective PSR?

@weierophinney I'm inclined to continue with this PR, but I'm not so confident about the changes to the code fragments in PSR-7. I had the impression when writing my previous comment that hey linked to www.ietf.org, but it seems quite a couple link to tools.ietf.org. As editor of that PSR, what is your opinion on the matter?

shadowhand (approved):
I approve the changes to PSR-15, PSR-17 as Editor of those PSRs.

mul91n:
I rechecked http://tools.ietf.org/html/* and it appears to be redirecting to https://datatracker.ietf.org/doc/html/* now so this PR may not be necessary anymore.

weierophinney:
I've always linked to tools.ietf.org/html/ vs www.ietf.org/standards/rfcs/ as the former are HTML formatted, and thus offer accessibility (including linking of TOC items to the specific headers) that the plain text versions at the latter do not. Additionally, they allow linking to specific _sections_, which, again, the plain text versions do not.

My suspicion is if they are changing the servers from which they provide RFCs, they will ensure that the old URLs resolve to their new locations, as the IETF is very much aware that there are links _everywhere_ to their RFCs, in _every_ format they offer, and they will not want to invalidate those URLs. The last post from @mul91n suggests this is the case. I think we should be patient and see if this is a long-running issue before making sweeping changes here.

mul91n:
Pretty clear this is a non-issue as IETF updated their redirects, closing this off.

PR #1252, opened 2021-08-25
Errata: Type additions

navarr:
Errata for proposed PR php-fig/simple-cache#24

Proposed versions are:

* v1.1.0 - f101a27
* v1.2.0 - a8085a0
* v2.0.0 - 3fc2676

### DIFF
--- a/accepted/PSR-16-simple-cache-meta.md
+++ b/accepted/PSR-16-simple-cache-meta.md
@@ -68,3 +68,26 @@ For being an early reviewer
 * [Survey of existing cache implementations][1], by @dragoonis
 
 [1]: https://docs.google.com/spreadsheet/ccc?key=0Ak2JdGialLildEM2UjlOdnA4ekg3R1Bfeng5eGlZc1E#gid=0
+
+## 8. Errata
+### 8.1 Type additions
+
+The 1.1 and 1.2 releases of the `psr/simple-cache` package includs scalar parameter types and increases the minimum PHP version to 7.2 and 8.0 respectively.  This is considered a backwards compatible change as PHP 7.2 introduces covariance for parameters.  Any implementation of 1.0 is compatible with 1.1 and 1.2.
+
+The 2.0 release includes return types.  Return types break backwards compatibility as PHP does not support return type covariance.  Implementations of 2.0 remain compatible with clients designed for 1.0.
+
+Implementers MAY add return types to their own packages at their discretion, provided that:
+
+* the return types match those in the 2.0 package.
+* the implementation specifies a minimum PHP version of 8.0.0 or later
+* the implementation depends on `"psr/simple-cache": "^1.2 || ^2"` so as to exclude the untyped 1.0 version.
+
+Implementers MAY add parameter types to their own package in a new minor release, either at the same time as adding return types or in a subsequent release, provided that:
+
+* the parameter types match or widen those in the 1.2 package
+* the implementation specifies a minimum PHP version of 7.2.0 (8.0 if using mixed or union types) or later.
+* the implementation depends on `"psr/simple-cache": "^1.1 || ^2"` so as to exclude the untyped 1.0 version. (`"psr/simple-cache": "^1.2 || ^2"` if using mixed or union types)
+
+Implementers are encouraged, but not required to transition their packages toward the 2.0 version of the package at their earliest convenience.
+
+Consumers are encouraged to update their requirement to `"psr/simple-cache": "^1 || ^2"` at their earliest convenience.


Jean85 (changes requested):
On hold until this passes the required CC vote.

navarr:
@Crell Is it OK/appropriate to edit the PSR example code to match the interface as defined by the errata as I have done in 069fd7f?

> Errata clarifications may only be added to the meta document, not to the spec itself, as items in their own section. The spec itself may only be minimally edited to point readers toward a relevant errata if appropriate.

https://www.php-fig.org/bylaws/psr-workflow/

navarr:
@dragoonis: You're the editor of PSR-16, would you support this errata?

accepted/PSR-16-simple-cache-meta.md
```diff
@@ -68,3 +68,26 @@ For being an early reviewer
 * [Survey of existing cache implementations][1], by @dragoonis
 
 [1]: https://docs.google.com/spreadsheet/ccc?key=0Ak2JdGialLildEM2UjlOdnA4ekg3R1Bfeng5eGlZc1E#gid=0
+
+## 8. Errata
+### 8.1 Type additions
+
+The 2.0 release of the `psr/simple-cache` package includs scalar parameter types and increases the minimum PHP version to 8.0 respectively.  This is considered a backwards compatible change for implementing libraries as PHP 7.2 introduces covariance for parameters.  Any implementation of 1.0 is compatible with 2.0. For calling libraries, however, this reduces the types that they may pass (as previously any parameter that could be cast to string could be accepted) and as such requires incrementing the major version.
```
Crell, line 75:
```suggestion
The 2.0 release of the `psr/simple-cache` package includes scalar parameter types and increases the minimum PHP version to 8.0.  This is considered a backwards compatible change for implementing libraries as PHP 7.2 introduces covariance for parameters.  Any implementation of 1.0 is compatible with 2.0. For calling libraries, however, this reduces the types that they may pass (as previously any parameter that could be cast to string could be accepted) and as such requires incrementing the major version.
```

Some typos.

accepted/PSR-16-simple-cache-meta.md
```diff
@@ -68,3 +68,26 @@ For being an early reviewer
...
+
+The 3.0 release includes return types.  Return types break backwards compatibility for implementing libraries as PHP does not support return type covariance.
```
Crell, line 77: This is not true.  PHP supports return type contravariance (or whatever the opposite of what parameters do is. I can never remember which is which.)  A child can add a more restrictive return type than its parent, but cannot widen it.

accepted/PSR-16-simple-cache-meta.md
```diff
@@ -68,3 +68,26 @@ For being an early reviewer
...
+
+Implementing libraries **MAY** add parameter types to their own package in a new minor release, either at the same time as adding return types or in a subsequent release, provided that:
+
+* the parameter types match or widen those in the 2.0 package
+* the implementation specifies a minimum PHP version of 8.0 if using mixed or union types or later.
+* the implementation depends on `"psr/simple-cache": "^2 || ^3"` so as to exclude the untyped 1.0 version.
+
+Implementing libraries are encouraged, but not required to transition their packages toward the 3.0 version of the package at their earliest convenience.
+
+Calling libraries are encouraged to ensure they are sending the correct types and to update their requirement to `"psr/simple-cache": "^1 || ^2 || ^3"` at their earliest convenience.
```
Crell, line 93: It's not actually possible to be compatible with version 1 and 3 at the same time.  That's the reason for the intermediary v2 step.

I think the instructions here are incorrect.  Please see the language in the other recently updated specs.

accepted/PSR-16-simple-cache.md
```diff
@@ -149,13 +149,13 @@ interface CacheInterface
...
      */
-    public function get($key, $default = null);
+    public function get(string $key, mixed $default = null): mixed;
 
     /**
      * Persists data in the cache, uniquely referenced by a key with an optional expiration TTL time.
      *
      * @param string                 $key   The key of the item to store.
-     * @param mixed                  $value The value of the item to store. Must be serializable.
+     * @param mixed                  $value The value of the item to store, must be serializable.
```
Crell, line 158: This change is not relevant.

accepted/PSR-16-simple-cache.md
```diff
@@ -149,13 +149,13 @@ interface CacheInterface
      * @throws \Psr\SimpleCache\InvalidArgumentException
      *   MUST be thrown if the $key string is not a legal value.
      */
-    public function get($key, $default = null);
+    public function get(string $key, mixed $default = null): mixed;
```
Crell, line 152: Consistent with the other specs, the spec text itself *should not be changed* with types.  Just update the code package itself.

accepted/PSR-16-simple-cache-meta.md
navarr:
```suggestion
The 3.0 release includes return types.  Return types break backwards compatibility for implementing libraries as PHP does not support return type widening.
```

accepted/PSR-16-simple-cache-meta.md
navarr: A calling library can be compatible with versions 1 and 3, an implementing library can't.

Jean85: Maybe we should underline that concept at the first mention of `calling library` then, since the difference can be summarized as "uses but don't implement the interfaces".

navarr: @Jean85 Calling Library is defined [in the standard under "Definitions"](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-16-simple-cache.md)

Geolim4:
Hello @Jean85 any news about this ? When this MR will pass ?

garak (approved):

Jean85:
No vote has been called for the CC yet. We need @dragoonis to push it or someone else to take over.

Crell:
Per the new bylaw updates, the CC can appoint a new Maintainer at its leisure.  Or it can just make the updates on its own authority.  We don't need Paul as he's clearly uninvolved anymore.

Crell (approved):
I think this is OK now.

Crell:
I think we can take this to a CC vote now.  Everything seems good to me.  Thanks, @Geolim4!

Geolim4:
> I think we can take this to a CC vote now. Everything seems good to me. Thanks, @Geolim4!

Okay, nice. Poke me when you release the 3.0.0 so I can update my composer constraint asap.

Thanks again for your time boys.

Geolim4:
> Per the new bylaw updates, the CC can appoint a new Maintainer at its leisure. Or it can just make the updates on its own authority. We don't need Paul as he's clearly uninvolved anymore.

@Crell does Maintainer application is open to anyone or only to Fig member ?
If it is open to external people, I would propose myself to maintain this PSR on behalf of Phpfastcache project.

Crell:
> The Maintainer of a given artifact is appointed by the Core Committee by Approval Vote. Once a given artifact is created, the Maintainer may step down and name a replacement Maintainer at any time with Implicit Approval. If necessary, a Decision vote to appoint a new Maintainer may be held following a suitable nomination period.

Could be anyone.  I... think that's slightly confusing as written (and I JUST wrote it, damnit, where are my reviewers at???), but I think if you apply via the list and no one else is interested, we just have an Approval vote on you.  If someone else also steps forward, we have a Decision Vote to choose among you.

That doesn't block a vote on these changes, though.  Right now the CC itself is the maintainer, which means we can just call an Approval vote like we did on the others.

Geolim4:
@Crell @Jean85 when will the vote ends so I can plan an approximate date to update my dependencies. Thank you.

navarr:
@Jean85 The symfony issue was specific to versions prior to 7.4 right?  Are there any requirements from implementing libraries that come with this change?

I've currently got in my editor the diff:

```diff
- ### 8.1 Type additions
+ ### 8.1 Throwable
+ 
+ The 2.0 release of the `psr/simple-cache` package updates `Psr\SimpleCache\CacheException` to extend `Throwable`.
+ 
+ ### 8.2 Type additions
```

Jean85:
Yes, the issue was only for PHP versions older than 7.4. Diff LGTM :+1: maybe just add a leading `\` to `Throwable` to avoid confusion.

accepted/PSR-16-simple-cache-meta.md
```diff
@@ -68,3 +68,30 @@ For being an early reviewer
 * [Survey of existing cache implementations][1], by @dragoonis
 
 [1]: https://docs.google.com/spreadsheet/ccc?key=0Ak2JdGialLildEM2UjlOdnA4ekg3R1Bfeng5eGlZc1E#gid=0
+
+## 8. Errata
+### 8.1 Throwable
```
Crell, line 73:
```suggestion

### 8.1 Throwable
```

Minor style nit.

navarr:
This is now up for a vote with the Core Committee: https://groups.google.com/g/php-fig/c/DJ8ysB1JV4Y

PR #1257, opened 2021-10-23
Add answer to a recurring question re UTC

heiglandreas:
The question "why not to convert to UTC" comes up very frequently so
Larry asked to add the answer to the meta-docs for PSR-20.

### DIFF
--- a/proposed/clock-meta.md
+++ b/proposed/clock-meta.md
@@ -150,7 +150,7 @@ be enforced on the language level is, that the returned value is of type `DateTi
 currently is no way to enforce that this object has a certain timezone as its value. So from a language 
 level perspective there is no way to explicitly enforce a certain timezone in the returned object.
 
-On the other hand there is also a logical reason. The explicit contract should be usable in all situations
+On the other hand, there is also a logical reason. The explicit contract should be usable in all situations
 where one needs a way to retrieve the current time. And on the contract level we should not make an 
 assumption about what the caller might need. So would the contract define that only UTC is returned, then 
 use-cases that do require something else will have to explicitly work around that or find other ways of 
--- a/proposed/clock-meta.md
+++ b/proposed/clock-meta.md
@@ -142,10 +142,27 @@ _**Note:** Order descending chronologically._
 
 ### Why not simply convert all times to UTC?
 
-As timezones are based on political definitions. Therefore they change: regularily and unforseeably. And sometimes even
-for past times. So when you are converting a DateTime to UTC you are using the offset that is known today 
-to convert from the local time to UTC. When that offset changes due to political decisions then you have no chance to 
-adapt to that. 
+`now()` should be considered to be the equivalent of checking your wallclock at home or your wristwatch. 
+It usually gives you the current point in time at your current location. And from that point you can 
+then – if you need to – transpose to other timezones. That makes life much easier than setting your
+wallclock to UTC and having to transpose every time you want to communicate with someone.
+
+As this interface is a generic interface it can not impose a certain usecase and therefore needs to 
+be as open as possible. That means to allow people to use any DateTimeImmutable object possible.
+And that also means to make sure users understand the consequence of that. Which is that it is 
+the users requirement to make sure the returned DateTimeImmutable object conforms to their requirements.
+
+Much as you would tell someone calling you on the phone your local time when they ask you for the time. 
+And if they have some special requirements they need to make sure that they handle the information
+you gave them properly.
+
+Besides that it has shown that transposing times is usually error prone and should be avoided at 
+all cost and only done if absolutely necessary. 
+
+That is due to timezones being based on political definitions. Therefore they change: regularily and unforseeably.
+And sometimes even for past times. So when you are converting a DateTime to UTC you are using the 
+offset that is known today to convert from the local time to UTC. When that offset changes due
+to political decisions then you have no chance to adapt to that. 
 
 An Example: In  summer 2016 you scheduled an event in turkey for 14:00 on the 25th of December 2016. 
 With the information known in summer 2016 you would calculate that that is 12:00 UTC. What you couldn't imagine is, 
@@ -161,12 +178,6 @@ and using the existing timezone information. And as there have been requests to 
 concern your current conversion from local time to UTC. So to avoid confusion is to avoid transposing between timezones
 as much as possible and save it to the last possible moment.
 
-`now()` should be considered to be the equivalent of checking your wallclock at home or your wristwatch. It usually 
-gives you the current point in time at your current location. And from that point you can then – if you 
-need to – transpose to other timezones. That makes life much easier than setting your wallclock to UTC and 
-having to transpose every time you want to communicate with someone. 
 
-As this interface is a generic interface it can not impose a certain usecase and therefore needs to be as open
-as possible. That means to allow people to use any DateTimeImmutable object possible. And that also means to 
-make sure users understand the consequence of that. Which is that it is the users requirement to make sure the 
-returned DateTimeImmutable object conforms to their requirements. 
+
+
--- a/proposed/clock-meta.md
+++ b/proposed/clock-meta.md
@@ -159,7 +159,7 @@ from the issue of i.e., immutability which can also not be enforced on the langu
 necessary to adhere to other calls on the contract. IN the case of the `clock`-interface there will be no 
 other calls.
 
-And most important of all, the explicit contract provided by this interface does not reduce a users
+And most important of all, the explicit contract provided by this interface does not reduce a user’s
 possibilities to use an implicit contract within their application, that defines that the interface 
 will only return `DateTimeImmutable`s with a specific timezone set. And whether that is `UTC` or reduce
 `Antarctica/Troll` is completely up to the user. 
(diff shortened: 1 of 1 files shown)


proposed/clock-meta.md
```diff
@@ -137,3 +137,11 @@ This document stems from the work of many people in previous years, we recognize
...
+
+## 9. FAQ
+
+### Why not simply convert all times to UTC?
+
+Timezones change! Regularily! And unforseeably! So when you are converting a DateTime in the future to UTC you are using the offset that is known today to convert from the local time to UTC. When that offset changes between now and the datetime then you have no chance to adapt to that. 
+An Example: in  summer 2016 you scheduled an event in turkey for 14:00 on the 25th of December 2016. With the information known in summer 2016 you would calculate that that is 12:00 UTC. What you couldn't imagine is, that the Government decided later that year to stay on daylight saving time and not go back to standard time for the winter. So when the time comes you would show up at 12:00 UTC. With the then relevant timezone information that would be 15:00 local time. But your appointment was at 14:00 local time. Would you have *not* converted to UTC you would have stored `14:00 Asia/Istanbul` and you would have been on time.
+**Note:** While this might only seem relevant for future datetimes it illustrates that transposiing between timezones can cause much more confusion than just keeping and using the existing timezone information.
```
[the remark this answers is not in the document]
heiglandreas: The Term "DateTime" here refers to a defined point in Date and Time. Not an instance of `DateTime`.

proposed/clock-meta.md
```diff
@@ -137,3 +137,11 @@ This document stems from the work of many people in previous years, we recognize
 
  * 
 _**Note:** Order descending chronologically._
+
+## 9. FAQ
+
+### Why not simply convert all times to UTC?
+
+Timezones change! Regularily! And unforseeably! So when you are converting a DateTime in the future to UTC you are using the offset that is known today to convert from the local time to UTC. When that offset changes between now and the datetime then you have no chance to adapt to that.
```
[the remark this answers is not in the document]
heiglandreas: That is not an issue as that information is encoded into the timezone-information.

Remember: A timezone is a geographical region that is observing the same date and time by governmental definition. Such a timezone is identified by a timezone-identifier such as `Europe/Moskow`. 
Do not confuse timezones with offsets that are defining the difference to the time-standard (which can be either GMT **or** UTC) or with named offsets like `MSK` that cover some (but not all) timezones that are observing a certain offset. A timezone can be referenced via different named offsets at different times due to DST changes.

So when using a timezone, the DST information is available. When using an offset then you do not know whether that offset is standard-time or daylight saving time.

proposed/clock-meta.md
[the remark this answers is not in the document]
heiglandreas: Unless you know off mails hitting the timezone-database mailinlist reading "We changed from Offset X to Y 2 weeks ago" or " we will change from offset x to y tomorrow" (which means a new tzdb version will not be available to machines at the time the change actually hits.)

So unless your server is actually set to observing UTC and you are 100% sure that the result of the `now` function taps directly into that and doesn't do any conversions in between using the local time might be the more accurate variant.

And regardless of that: `now()` is the equivalent of checking your clock at home or your wristwatch. Which usually shows your local date and time. And from that you can then decide for yourself what you make up with that information.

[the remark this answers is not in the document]
samdark: Maybe it's better to write "date and time" then to avoid confusion?

Crell:
The analogy based argument isn't really convincing for me.  At this point I've essentially resigned myself to the fact that the timezone will be undefined, but the argument for it needs to be solid enough to avoid 10 years of people whining about it.  (See also: PSR-7.)  I don't feel like this is.

heiglandreas:
I've redone the argumentation completely away from advocating the use of local time over UTC (which is still a good idea, but kinda hard to convince people that it makes sense even for current date-times) over to trying to make it clear that the interface is as open as possible while at the same time being as strict as necessary. Not enforcing UTC on the explicit contract that the interface defines does not hinder an implicit contract on the implementation level which defines that the interface will always return objects with a certain timezone preset.

proposed/clock-meta.md
```diff
@@ -137,3 +137,37 @@ This document stems from the work of many people in previous years, we recognize
...
+
+There are different reasons why this interface does not enforce a specific timezone.
+
+For one thing there is a purely technical reason. The interface provides an explicit contract. Part of this 
+contract is the value that is returned by the `now`-function. The only thing that currently can explicitly 
+be enforced on the language level is, that the returned value is of type `DateTimeImmutable`. There 
+currently is no way to enforce that this object has a certain timezone as its value. So from a language 
+level perspective there is no way to explicitly enforce a certain timezone in the returned object.
+
+On the other hand there is also a logical reason. The explicit contract should be usable in all situations
```
samdark, line 153:
```suggestion
On the other hand, there is also a logical reason. The explicit contract should be usable in all situations
```

proposed/clock-meta.md
```diff
@@ -137,3 +137,37 @@ This document stems from the work of many people in previous years, we recognize
...
+where one needs a way to retrieve the current time. And on the contract level we should not make an 
+assumption about what the caller might need. So would the contract define that only UTC is returned, then 
+use-cases that do require something else will have to explicitly work around that or find other ways of 
+handling the problem the contract tries to solve as the contract does not fit the needs. This is different
```
samdark, line 157:
```suggestion
handling the problem, the contract tries to solve as the contract does not fit the needs. This is different 
```

proposed/clock-meta.md
```diff
@@ -137,3 +137,37 @@ This document stems from the work of many people in previous years, we recognize
...
+be enforced on the language level is, that the returned value is of type `DateTimeImmutable`. There 
+currently is no way to enforce that this object has a certain timezone as its value. So from a language 
+level perspective there is no way to explicitly enforce a certain timezone in the returned object.
+
+On the other hand there is also a logical reason. The explicit contract should be usable in all situations
+where one needs a way to retrieve the current time. And on the contract level we should not make an 
+assumption about what the caller might need. So would the contract define that only UTC is returned, then 
+use-cases that do require something else will have to explicitly work around that or find other ways of 
+handling the problem the contract tries to solve as the contract does not fit the needs. This is different 
+from the issue of i.e. immutability which can also not be enforced on the language level but which is
```
samdark, line 158:
```suggestion
from the issue of i.e., immutability which can also not be enforced on the language level but which is
```

proposed/clock-meta.md
```diff
@@ -137,3 +137,37 @@ This document stems from the work of many people in previous years, we recognize
...
+necessary to adhere to other calls on the contract. IN the case of the `clock`-interface there will be no 
+other calls.
+
+And most important of all, the explicit contract provided by this interface does not reduce a users
```
samdark, line 162:
```suggestion
And most important of all, the explicit contract provided by this interface does not reduce a user’s
```

proposed/clock-meta.md
```diff
@@ -137,3 +137,37 @@ This document stems from the work of many people in previous years, we recognize
...
+possibilities to use an implicit contract within their application, that defines that the interface 
+will only return `DateTimeImmutable`s with a specific timezone set. And whether that is `UTC` or reduce
+`Antarctica/Troll` is completely up to the user. 
+
+The explicit contract that is defined by the interface does not limit a user in what they are doing. It 
+tries to solve the problem of getting the current time in a reliable way. Which view on the current time 
+that is, is not part of the explicit contract.
+
+So all in all: This interface tries to be as open as possible while at the same time being as
```
samdark, line 171:
```suggestion
So, all in all: This interface tries to be as open as possible while at the same time being as 
```

samdark (approved):
The addition itself is fine. Minor issues are easy to be fixed.

localheinz:
What is the hold-up here - can we move this forward?

proposed/clock-meta.md
```diff
@@ -137,3 +137,37 @@ This document stems from the work of many people in previous years, we recognize
...
+currently is no way to enforce that this object has a certain timezone as its value. So from a language 
+level perspective there is no way to explicitly enforce a certain timezone in the returned object.
+
+On the other hand, there is also a logical reason. The explicit contract should be usable in all situations
+where one needs a way to retrieve the current time. And on the contract level we should not make an 
+assumption about what the caller might need. So would the contract define that only UTC is returned, then 
+use-cases that do require something else will have to explicitly work around that or find other ways of 
+handling the problem, the contract tries to solve as the contract does not fit the needs. This is different 
+from the issue of i.e., immutability which can also not be enforced on the language level but which is
+necessary to adhere to other calls on the contract. IN the case of the `clock`-interface there will be no
```
BafS, line 159:
```suggestion
necessary to adhere to other calls on the contract. In the case of the `clock`-interface there will be no 
```
Typo?

ashnazg (approved):
Some of the "English flow" could be tweaked, but content seems good to me.

ashnazg:
@cseufert ping... Github shows you still active in general, so reaching out to you here rather than just on Discord to confirm you're still interested in this PSR... ?

cseufert (approved):

ashnazg:
@cseufert, as Editor, you can merge these, right?

cseufert:
I cant merge it, can only approve it for merging

ashnazg:
> I cant merge it, can only approve it for merging

Ok, I'll find out who can, and get both open PRs merged.  Thanks for getting them both approved, and welcome back ;-)

Jean85:
Done, my pleasure! :smile:

PR #1198, opened 2019-11-05
Update PSR-4-autoloader-examples.md

WinterSilence:
### Please, don't close - this PR using in Github bug report, I'm will close him, when ticket done.

Fix directory separator, long/strange conditions. 
Optimize method `loadMappedFile`: move up `str_replace`.

PR #1266, opened 2022-01-25
Update `PSR-4-autoloader-examples.md`

WinterSilence:
Base on #1198
- PSR-12 in examples in PHPDoc descriptions
- `Psr4AutoloaderClass::register()`: add `$prepend` parameter, set `.php` extension directly.
- `Psr4AutoloaderClass::addNamespace()`: normalize Windows paths
- `Psr4AutoloaderClass::loadMappedFile()`: optimize code

### DIFF
--- a/accepted/PSR-4-autoloader-examples.md
+++ b/accepted/PSR-4-autoloader-examples.md
@@ -15,7 +15,8 @@ Closure Example
  * would cause the function to attempt to load the \Foo\Bar\Baz\Qux class
  * from /path/to/project/src/Baz/Qux.php:
  *
- *      new \Foo\Bar\Baz\Qux;
+ *      <?php
+ *      new \Foo\Bar\Baz\Qux();
  *
  * @param string $class The fully-qualified class name.
  * @return void
@@ -83,7 +84,7 @@ namespace Example;
  *
  *      <?php
  *      // instantiate the loader
- *      $loader = new \Example\Psr4AutoloaderClass;
+ *      $loader = new \Example\Psr4AutoloaderClass();
  *
  *      // register the autoloader
  *      $loader->register();
@@ -96,13 +97,13 @@ namespace Example;
  * \Foo\Bar\Qux\Quux class from /path/to/packages/foo-bar/src/Qux/Quux.php:
  *
  *      <?php
- *      new \Foo\Bar\Qux\Quux;
+ *      new \Foo\Bar\Qux\Quux();
  *
  * The following line would cause the autoloader to attempt to load the
  * \Foo\Bar\Qux\QuuxTest class from /path/to/packages/foo-bar/tests/Qux/QuuxTest.php:
  *
  *      <?php
- *      new \Foo\Bar\Qux\QuuxTest;
+ *      new \Foo\Bar\Qux\QuuxTest();
  */
 class Psr4AutoloaderClass
 {
@@ -117,11 +118,14 @@ class Psr4AutoloaderClass
     /**
      * Register loader with SPL autoloader stack.
      *
+     * @param bool $prepend If true, prepend the autoloader to the stack
+     * instead of appending it.
      * @return void
      */
-    public function register()
+    public function register($prepend = false)
     {
-        spl_autoload_register(array($this, 'loadClass'));
+        spl_autoload_extensions('.php');
+        spl_autoload_register(array($this, 'loadClass'), false, $prepend);
     }
 
     /**
@@ -141,7 +145,7 @@ class Psr4AutoloaderClass
         $prefix = trim($prefix, '\\') . '\\';
 
         // normalize the base directory with a trailing separator
-        $base_dir = rtrim($base_dir, DIRECTORY_SEPARATOR) . '/';
+        $base_dir = rtrim($base_dir, '\/') . '/';
 
         // initialize the namespace prefix array
         if (isset($this->prefixes[$prefix]) === false) {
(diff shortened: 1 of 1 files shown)


accepted/PSR-4-autoloader-examples.md
```diff
@@ -141,7 +145,7 @@ class Psr4AutoloaderClass
         $prefix = trim($prefix, '\\') . '\\';
 
         // normalize the base directory with a trailing separator
-        $base_dir = rtrim($base_dir, DIRECTORY_SEPARATOR) . '/';
+        $base_dir = rtrim($base_dir, '\/') . '/';
```
drupol, line 148: Why changing `DIRECTORY_SEPARATOR` into this?

accepted/PSR-4-autoloader-examples.md
```diff
@@ -208,15 +212,15 @@ class Psr4AutoloaderClass
             return false;
         }
 
+        // replace namespace separators with directory separators
+        // in the relative class name, append with '.php'
+        $filename = str_replace('\\', '/', $relative_class) . '.php';
```
drupol, line 217: Nice!

accepted/PSR-4-autoloader-examples.md
```diff
@@ -180,7 +184,7 @@ class Psr4AutoloaderClass
 
             // try to load a mapped file for the prefix and relative class
             $mapped_file = $this->loadMappedFile($prefix, $relative_class);
-            if ($mapped_file) {
+            if ($mapped_file !== false) {
```
drupol, line 187: I fully agree with this change.

accepted/PSR-4-autoloader-examples.md
```diff
@@ -117,11 +118,14 @@ class Psr4AutoloaderClass
...
+     * @param bool $prepend If true, prepend the autoloader to the stack
+     * instead of appending it.
      * @return void
      */
-    public function register()
+    public function register($prepend = false)
     {
-        spl_autoload_register(array($this, 'loadClass'));
+        spl_autoload_extensions('.php');
+        spl_autoload_register(array($this, 'loadClass'), false, $prepend);
```
WinterSilence, line 128: second parameter is `false` to omit `@throws` in PHPDoc, this added for static analyze tools

accepted/PSR-4-autoloader-examples.md
```diff
@@ -117,11 +118,14 @@ class Psr4AutoloaderClass
     /**
      * Register loader with SPL autoloader stack.
      *
+     * @param bool $prepend If true, prepend the autoloader to the stack
+     * instead of appending it.
      * @return void
      */
-    public function register()
+    public function register($prepend = false)
```
drupol, line 125: I guess we can type the argument to `bool` ?

accepted/PSR-4-autoloader-examples.md
WinterSilence: @drupol need use it everywhere or nowhere

drupol: But then using `DIRECTORY_SEPARATOR` is the way to go, no ?

accepted/PSR-4-autoloader-examples.md
WinterSilence: @drupol you mean `spl_autoload_register(array($this, 'loadClass'), false, (bool) $prepend);` ?

drupol: No, this:

```suggestion
    public function register(bool $prepend = false)
```

accepted/PSR-4-autoloader-examples.md
drupol: Thanks, I've learned something today :)

accepted/PSR-4-autoloader-examples.md
WinterSilence: @drupol it's not PHP5-compatible code

drupol: And it must be PHP5 compatible ? #shocked

WinterSilence: @drupol  lack of typing indicates it

drupol: Do we have the official confirmation that PHP5 needs to be supported?

If yes, please ignore my comment.

accepted/PSR-4-autoloader-examples.md
WinterSilence: @drupol for the same reason, I keeped useless check `false !==` in line `while (false !== $pos = strrpos($prefix, '\\')) {`

accepted/PSR-4-autoloader-examples.md
WinterSilence: @drupol I'm don't see reasons to not support PHP5. Anyway, it's not for me to decide.

drupol: Could be a long thread but basically, I don't see the reason to support something which is [no more supported officially](https://www.php.net/eol.php)... or else, what will be the limit?
I think adding types really help a lot of people when programming and avoid also many issues in advance.
But like you said, it's not up to us to decide, and it would be nice to have the point of view of someone who does.

accepted/PSR-4-autoloader-examples.md
WinterSilence: @drupol sorry, my fail, i'm gave invalid answer -  it's using to normalize Windows paths too

accepted/PSR-4-autoloader-examples.md
Crell: This is example code.  If we decide we can non-trivially touch it at all, we should go all in and update it to modern code.  Hell, I'd include union return types where appropriate, too, at that point.

That's if we decide we can touch it non-trivially at all to begin with; it's not obvious if this is part of the spec (which has vastly higher standards for what you can do) or not.

WinterSilence: @Crell @drupol without flame please

drupol: I shut up and listen.

PR #1271, opened 2022-03-13
Fix PHP snippet highlighting for PSR evolving page

andrew-demb:


drupol (approved):

andrew-demb:
If you think we need to preserve quoting code snippet (as before) - please let me know

Crell:
I defer to the secretaries on how we want to format the "Example" here.  Markdown is not great at this.

Jean85 (commented):
:-1: on this, this is intentional, we want to show what should be written MarkDown-wise, as you can see in the website:
image

PR #1273, opened 2022-03-21
PSR-Tracing - Pre-Draft Specification & Meta

AAllport:
Add Initial documents for Tracing PSR

Discussion: https://groups.google.com/g/php-fig/c/6Nm0ZQ1_hJA

### DIFF
--- a/proposed/tracing-meta.md
+++ b/proposed/tracing-meta.md
@@ -25,7 +25,7 @@ This standard aims to create an intentionally minimalist set of interfaces that 
 ## 3 Approaches
 
 To fulfil the requirement of most APM tools, this PSR will be loosely modeled on the [OpenTelemetry Tracing Specification][OTelTrace] and [Tracing API][OtelTraceApi].\
-We aim to allow the majority of APM tools to use the PSR interfaces without any backwards incompatible changes.
+We aim to allow the majority of APM tools to use the PSR interfaces a minimum backwards incompatible changes as to encourage adoption.
 
 [OtelTrace]: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/overview.md#tracing-signal
 [OtelTraceApi]: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md
--- a/proposed/tracing-meta.md
+++ b/proposed/tracing-meta.md
@@ -22,12 +22,9 @@ By taking transmission mechanisms [out of scope](#31-non-goals) for this PSR, we
 ### 3.1 Goals
 
 * To provide a set of interfaces for library and framework developers to add tracing signals to their codebase.  
-  This would in turn allow other libraries (Eg: [Clockwork], [Laravel-Debugbar]) to receive the same signals for further processing or analysis.
+  This would in turn allow other libraries to receive the same signals for further processing or analysis.
 * To allow traces collected by various providers to be reused by other providers.
 * This PSR may provide a minimal `TraceProvider`, etc. for other providers to extend, should they choose.
-
-[Clockwork]: https://underground.works/clockwork/
-[Laravel-Debugbar]: https://github.com/barryvdh/laravel-debugbar
 
 ### 3.1 Non-goals
 
--- a/proposed/tracing-meta.md
+++ b/proposed/tracing-meta.md
@@ -1,0 +1,48 @@
+# Tracing Meta Document
+
+## 1. Summary
+
+Application performance monitoring (APM) tools are becoming increasingly popular, however, the lack of formal interfaces is holding back the ecosystem.
+Whilst developers are inherently free to use whichever tools suit their use-cases, this is not the case for framework and library creators/maintainers.
+
+This standard aims to create an intentionally minimalist set of interfaces that can be used to provide tracing signals to 3rd party library's in a unified manner.
+
+## 3. Scope
+
+### 3.1 Goals
+
+* To provide a set of interfaces for library and framework developers to add tracing signals to their codebase.
+  This would in tern allow other APM providers to receive the same signals for further processing or analysis.
+* To allow traces collected by various APM libraries to be reused by other libraries/tools.
+* This PSR may provide a minimal `TraceProvider`, etc. for other APM tools to extend, should the choose.
+
+### 2.1 Non-goals
+
+* This PSR does not provide a comprehensive tracing client
+* This PSR does not define the mechanisms used for transmitting the data to 3rd party systems
+* This PSR does not cover collecting metrics within an codebase. Only traces
+
+## 3 Approaches
+
+To fulfil the requirement of most APM tools, this PSR will be modeled loosely on the OpenTelemetry (Trace).
+This will involve 
+We aim to allow the majority of APM tools to use the PSR interfaces without any backwards incompatible changes.
+
+
+## 4. People
+
+### 4.1 Editor
+* Adam Allport
+
+### 4.1 Sponsor
+* Alessandro Chitolina
+
+### Working Group Members
+* Alex Bouma
+* Ben.e
+* Brett McBride
+* Timo Michna
+
+## 6. Votes
+
+* [Entrance Vote](TBD)
--- a/proposed/tracing-meta.md
+++ b/proposed/tracing-meta.md
@@ -1,4 +1,4 @@
-# Tracing Meta Document
+# Application Tracing Meta Document
 
 ## 1. Summary
 
--- a/proposed/tracing-meta.md
+++ b/proposed/tracing-meta.md
@@ -19,12 +19,11 @@ The following definitions apply for this document:
 ## 2. Why Bother?
 
 The [OpenTelemetry][] team are in the process of releasing a set of SDK's that would allow frameworks and libraries to send signals to providers in a uniformed manner.
-However, there is a perceived expectation of a 1-way flow of responsibility, where Frameworks and Libraries are expected to accept Jagger traces[^1].
+However, there is a perceived expectation of a 1-way flow of responsibility, where Frameworks and Libraries are expected to accept Jagger-formatted traces.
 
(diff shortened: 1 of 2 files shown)


proposed/tracing-meta.md
```diff
@@ -0,0 +1,47 @@
...
+Whilst developers are inherently free to use whichever tools suit their use-cases, this is not the case for framework and library creators/maintainers.
+
+This standard aims to create an intentionally minimalist set of interfaces that can be used to provide tracing signals to 3rd party library's in a unified manner.
+
+## 3. Scope
+
+### 3.1 Goals
+
+* To provide a set of interfaces for library and framework developers to add tracing signals to their codebase.
+  This would in tern allow other APM providers to receive the same signals for further processing or analysis.
```
PowerKiKi, line 15:
```suggestion
  This would in turn allow other APM providers to receive the same signals for further processing or analysis.
```

proposed/tracing-meta.md
```diff
@@ -0,0 +1,47 @@
...
+* To allow traces collected by various APM libraries to be reused by other libraries/tools.
+* This PSR may provide a minimal `TraceProvider`, etc. for other APM tools to extend, should the choose.
```
PowerKiKi, line 17:
```suggestion
* This PSR may provide a minimal `TraceProvider`, etc. for other APM tools to extend, should they choose.
```

proposed/tracing-meta.md
```diff
@@ -0,0 +1,47 @@
...
+
+### 2.1 Non-goals
+
+* This PSR does not provide a comprehensive tracing client
+* This PSR does not define the mechanisms used for transmitting the data to 3rd party systems
+* This PSR does not cover collecting metrics within an codebase. Only traces
```
PowerKiKi, line 23:
```suggestion
* This PSR does not cover collecting metrics within a codebase, only traces
```

AAllport:
Thanks for those grammar checks @PowerKiKi, all applied!

Jean85:
Question: is it correct to say that this would be a PSR and not a PER? Can we count on OTel not changing in the future?

AAllport:
> Question: is it correct to say that this would be a PSR and not a PER? Can we count on OTel not changing in the future?

Good point. And it had certainly crossed my mind.
Whilst I will ultimately defer to @alekitto, this feels like it would be suitable as a PSR due to the limited scope of the document.

I think it's highly unlikely that OTel will move away from a Tracers->Spans->SubSpan layout.
Given this proposal is aimed at providing a common denominator. Only a drastic change to that methodology would cause this to become _outdated_

I do call on the OTel team to dispute anything I've said there :p
CC: @tidal @brettmc

alekitto:
> Question: is it correct to say that this would be a PSR and not a PER? Can we count on OTel not changing in the future?

Interesting question.

In my opinion yes, this should be a PSR: the goal should be providing a minimal stable set of interfaces the library authors can rely on. Also, OTel spec just reached 1.0, so can be considered stable enough for our goal. Ultimately, declaring that this specification will be _constantly_ changing, will sure cause a negative impact on its adoption.

I will be glad to hear any other opinion on this subject

tidal:
> OTel spec just reached 1.0

To clarify, The Trace "signal" API (which is relevant for us here) is marked as stable in the Otel specification.
The PHP implementation is not yet stable, however we are not planning on any drastic changes regarding 
the API (that's the component were all the relevant interfaces live).
However since all @open-telemetry/php-maintainers are on board with this initiative, we can still make 
adjustments if it makes sense.

> Can we count on OTel not changing in the future?

Yes and No. 
You can have a look at Otel's stability guarantess:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#api-stability
However, we as Otel PHP maintainers have a great interest in this PSR, as it will also help us with adoption.
So we likely don't want to break anything ragarding interfaces.

proposed/tracing.md
```diff
@@ -0,0 +1,15 @@
+# Common Interfaces and Functionality for Interoperability of Tracing for Application Performance Monitoring (APM)
```
Crell, line 1: Could stand to be shorter. :smile:

proposed/tracing-meta.md
```diff
@@ -0,0 +1,49 @@
...
+### 2.1 Non-goals
+
+* This PSR does not provide a comprehensive tracing client
+* This PSR does not define the mechanisms used for transmitting the data to 3rd party systems
+* This PSR does not cover collecting metrics within a codebase, only traces
+
+## 3 Approaches
+
+To fulfil the requirement of most APM tools, this PSR will be loosely modeled on the [OpenTelemetry Tracing Specification][OTelTrace] and [Tracing API][OtelTraceApi].\
+We aim to allow the majority of APM tools to use the PSR interfaces without any backwards incompatible changes.
```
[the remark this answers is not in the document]
AAllport: My original thought was "Aim" would create that flexibility.
But I agree the statement you have provided is certainly clearer. I will update 🔜 !

Crell:
Note that it's totally fair game for the spec to reference some other existing spec.  Eg, "this value may be any legal mime type as defined by IANA" or "this value may be any tag defined by the OpenTelemetry spec as listed at URL blah".  IMO, that's a fair "loophole" around spec's being immutable, as long as the thing they're linking out to is cleanly extensible and well maintained.  (That would preclude, for instance, using an enum to enforce a fixed list, but if the list is not truly fixed then that wouldn't make sense anyway."

Also, can I just note for the record that I love seeing the OT maintainers working here directly to ensure everything is as consistent and compatible as possible.  That's the ideal case. :grin:

proposed/tracing-meta.md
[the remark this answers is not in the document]
AAllport: Resolved: 3f318e9

proposed/tracing.md
AAllport: Resolved: 30401ee

proposed/tracing-meta.md
```diff
@@ -0,0 +1,49 @@
+# Application Tracing Meta Document
+
+## 1. Summary
+
+Application performance monitoring (APM) tools are becoming increasingly popular, however, the lack of formal interfaces is holding back the ecosystem.
+Whilst developers are inherently free to use whichever tools suit their use-cases, this is not the case for framework and library creators/maintainers.
+
+This standard aims to create an intentionally minimalist set of interfaces that can be used to provide tracing signals to 3rd party library's in a unified manner.
```
brettmc, line 8: *libraries

brettmc, line 8:
```suggestion
This standard aims to create an intentionally minimalist set of interfaces that can be used to provide tracing signals to 3rd party libraries in a unified manner.
```

proposed/tracing-meta.md
```diff
@@ -0,0 +1,49 @@
...
+### 2.1 Non-goals
+
+* This PSR does not provide a comprehensive tracing client
+* This PSR does not define the mechanisms used for transmitting the data to 3rd party systems
+* This PSR does not cover collecting metrics within a codebase, only traces
+
+## 3 Approaches
+
+To fulfil the requirement of most APM tools, this PSR will be loosely modeled on the [OpenTelemetry Tracing Specification][OTelTrace] and [Tracing API][OtelTraceApi].\
+We aim to allow the majority of APM tools to use the PSR interfaces a minimum backwards incompatible changes as to encourage adoption.
```
brettmc, line 28:
```suggestion
We aim to allow the majority of APM tools to use the PSR interfaces with minimum backwards incompatible changes to encourage adoption.
```
I think this is what is meant...

AAllport: 🤦

proposed/tracing-meta.md
```diff
@@ -0,0 +1,49 @@
...
+
+## 3. Scope
+
+### 3.1 Goals
+
+* To provide a set of interfaces for library and framework developers to add tracing signals to their codebase.
+  This would in turn allow other APM providers to receive the same signals for further processing or analysis.
+* To allow traces collected by various APM libraries to be reused by other libraries/tools.
+* This PSR may provide a minimal `TraceProvider`, etc. for other APM tools to extend, should they choose.
+
```
brettmc, line 18:
```suggestion
* To provide a set of interfaces for HTTP client implementations to enable sending outgoing distributed trace headers.
```

I think that an important goal of the spec should also be to provide interfaces to support distributed tracing. I see it as separate to providing tracing signals, and I think it's more widely useful. I believe most/all APM implementations support the [W3C traceparent+tracestate headers](https://www.w3.org/TR/trace-context/) (I checked elastic, datadog, and new relic), but they sometimes leave the injection of outgoing trace headers as an exercise for the reader. Having a PSR interface that was implemented by popular http client implementations would make tracing easier and more useful, IMO.

alekitto: I'm not sure it fits this proposal: distributed tracing is surely useful and we need to provide interfaces to support it, but the choice of decorating an http request with tracing header should be an implementation detail. 

Additionally this proposal is not an extension of PSR-18 and it shouldn't be from my point of view.

Surely http clients should be aware of the problem, but whether to decorate a request and how to do it is little bit out of scope at the moment.

brettmc: That makes sense - perhaps then something lighter like `To expose distributed tracing information to other libraries/tools`? I guess the important part is a consistent way to get those values from a tracing implementation. If we think that the existing wording sufficiently covers this, then I'm happy with the goals as they are.

alekitto: Yes, much better to me.
Maybe is already covered by `To allow traces collected by various APM libraries to be reused by other libraries/tools.`, but we can explicit it to make it crystal clear.

AAllport: I believe I have added clarification by way of the definitions in 916f0c2
Please do let me know if you disagree, etc

Jean85 (approved):
Entrance vote passed! https://groups.google.com/g/php-fig/c/GP8ThJsC8ac

PR #1279, opened 2022-05-06
Incorrect namespace prefix in example table

Netflyerr:


drupol (approved):

shadowhand (changes requested):
This change is not correct.

accepted/PSR-4-autoloader.md
```diff
@@ -62,7 +62,7 @@ class name, namespace prefix, and base directory.
 | Fully Qualified Class Name    | Namespace Prefix   | Base Directory           | Resulting File Path
 | ----------------------------- |--------------------|--------------------------|-------------------------------------------
 | \Acme\Log\Writer\File_Writer  | Acme\Log\Writer    | ./acme-log-writer/lib/   | ./acme-log-writer/lib/File_Writer.php
-| \Aura\Web\Response\Status     | Aura\Web           | /path/to/aura-web/src/   | /path/to/aura-web/src/Response/Status.php
+| \Aura\Web\Response\Status     | Aura\Web\Response  | /path/to/aura-web/src/   | /path/to/aura-web/src/Response/Status.php
```
shadowhand, line 65: If the prefix was set to `Aura\Web\Response` then the resulting file path for `Aura\Web\Response\Status` would be `src/Status.php`.

drupol (commented):
Indeed, sorry, the change is not correct.

accepted/PSR-4-autoloader.md
drealecs: Exactly. Existing prefix is inline with the Aura.Web configured composer psr-4 mapping: https://github.com/auraphp/Aura.Web/blob/2.x/composer.json#L22

PR #1281, opened 2022-06-10
Add PERs

Jean85:
This is due to the approval of the first PER release: https://github.com/php-fig/per-coding-style/releases/1.0.0

It will also lay the way to show this in the site.


PER.md
```diff
@@ -0,0 +1,25 @@
+# PHP Evolving Recommendations
+
+A PHP Evolving Recommendations is a meta document accompanied by one or more artifacts that are set to evolve over time with multiple releases.
+Following the [PER Workflow Bylaw][workflow], when a PER has passed the Entrance Vote it will be listed here as "In Development"; once the first release of an artifact is tagged, it will be listed as "Released".
+
+As also described in the [Mission and Structure Bylaw][structure], the Editor of a working group working on a PER is essentially the lead contributor and writer, and they are supported by at least two voting members; the Sponsor is instead a Project Representative or Core Committee member that acts as a steward and helps pushing the proposal forward.
```
Crell, line 6:
```suggestion
As also described in the [Mission and Structure Bylaw][structure], the Editor of a working group working on a PER is essentially the lead contributor and writer, and they are supported by at least two voting members; the Sponsor, if present, is a Core Committee member that assists the Editor in keeping the evolution process on track.
```

PER.md
```diff
@@ -0,0 +1,25 @@
...
+
+| Title                            | Editor        | Sponsor          |
+|----------------------------------|---------------|------------------|
+| [Coding Style][per-coding-style] | Korvin Szanto | Chris Tankersley |
+
+### In Development
+
+| Num | Title | Editor |
+|:---:|-------|--------|
+|  -  | -     | -      |
```
Crell, line 20: I don't know what "In Development" means in this case, since PERs have no end date.

Perhaps "Active" to indicate all WGs that are chartered and doing anything, and "Latest Releases" to link to the specific artifacts?

Jean85: As explained in the paragraph before, I wanted to describe PERs with no available releases (yet), so basically the same as the "Draft" phase for PSRs.

As for releases, I think that keeping the site up to date could be hard since it's a static site; on the "backend", this PR includes a single submodule of the PER, to which we can ask for a single checkout, and the latest release seems the best approach to me. As to releases, maybe a link to the GitHub Releases page is enough? Or to Packagist, if applicable.

PER.md
Jean85: I agree with the change, but Project Reps can be sponsors, you left that out: https://github.com/php-fig/fig-standards/blob/803f9f6d94a22a1624ace8af38db0aadb34ed915/bylaws/001-mission-and-structure.md?plain=1#L13

Crell: Project Reps can sponsor a project membership application.  They cannot sponsor a Working Group.

PER.md
Crell: I don't follow.  Latest release is what I'm talking about.  The caveat being that we cannot link to the `master` branch of a PER repo, as that is a development branch.  Only the latest tag is relevant.  If we link to the dev branch of the CS PER, we'll get whatever pre-release stuff hasn't been fully approved yet.

Jean85: Ok. Does [161d8a0](https://github.com/php-fig/fig-standards/pull/1281/commits/161d8a003ebf06dc387fe7e824be141881db5b9f) fix it?

PER.md
Jean85: Applied in [161d8a0](https://github.com/php-fig/fig-standards/pull/1281/commits/161d8a003ebf06dc387fe7e824be141881db5b9f).

PER.md
Crell: If we're OK with it linking to GitHub rather than being integrated into the site, then I believe so.

PER.md
```diff
@@ -0,0 +1,16 @@
+# PHP Evolving Recommendations
+
+A PHP Evolving Recommendations is a meta document accompanied by one or more artifacts that are set to evolve over time with multiple releases.
```
Crell, line 3:
```suggestion
A PHP Evolving Recommendation is a meta document accompanied by one or more artifacts that are set to evolve over time with multiple releases.
```

PER.md
Jean85: I don't know how we could integrate it. I'm up for suggestion, but that concerns php-fig/www.php-fig.org#307 (or we could still do a separate PR later)

samdark (approved):

PR #1282, opened 2022-06-16
Fix broken link to RFC2119

nicolus:


samdark:
But it isn't broken. It does redirect correctly.

nicolus:
@samdark : Oh ok, my bad, apparently my corporate proxy is blocking the redirection, as I'm getting this

image

But yeah, just tested while tethering from my phone and it redirects fine. Off to my IT's bug tracker then, sorry about that !

PR #1283, opened 2022-06-23
fix: Include the example of using the shortform of a closure

fruitl00p:
This explains the use of "spaces" around `fn ()`

### DIFF
--- a/accepted/PSR-12-extended-coding-style-guide.md
+++ b/accepted/PSR-12-extended-coding-style-guide.md
@@ -1054,24 +1054,27 @@ in the above section.
 ~~~php
 <?php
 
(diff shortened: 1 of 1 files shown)


Jean85:
I'm sorry, but we can't amend PSR-12 like this, it would have to go through an errata: https://www.php-fig.org/bylaws/psr-amendments/

At the same time, we're already working on https://github.com/php-fig/per-coding-style/, which is a natural successor of PSR-12. You should direct this suggestion there.

/cc @KorvinSzanto @Crell

Crell:
cf: https://github.com/php-fig/per-coding-style/pull/17

PR #1284, opened 2022-07-04
New PSR for data querying

alexandre-daubois:
A new PSR to standardize how data can be queried.

Here is the Google Groups thread: https://groups.google.com/g/php-fig/c/bKg4OvW-wcY

### DIFF
--- a/proposed/query.md
+++ b/proposed/query.md
@@ -1,0 +1,173 @@
+Common Interface for data querying
+==================================
+
+This document describes common interfaces to query data. These data can be from different sources, from in-memory data to databases, as well as filesystem files.
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in [RFC 2119][].
+
+The final implementations MAY decorate the objects with more
+functionality than the one proposed but they MUST implement the indicated
+interfaces/functionality first.
+
+[RFC 2119]: http://tools.ietf.org/html/rfc2119
+
+## 1. Specification
+
+### 1.1 Goal
+
+There are numerous ways of querying data from so many sources. Each system has its very own way of doing it. For instance, Doctrine ORM and Symfony Finder are both querying data with query builders (or alike), but there is no standardized interface to implement such data querying.
+
+The goal of this PSR it to define common interfaces to query data, no matter where they come from (arrays, iterators, etc.) or their format.
+
+### 1.2 Definitions
+
+*    **Query** - The object that will be used to query data. It receives a source as an `iterable`, which allows processing of arrays as well as other type of collections like generators.
+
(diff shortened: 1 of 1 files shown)


samdark:
Interesting. We have a different approach to similar problem at Yii: https://github.com/yiisoft/data

alexandre-daubois:
@samdark I **really** like how you managed filters! This is a great way to do this I didn't think of!

proposed/query.md
```diff
@@ -0,0 +1,173 @@
...
+     *
+     * @param iterable $source
+     *      The source to apply manipulations on.
+     *
+     * @param QueryContextInterface|null $context
+     *      The optional context that forward needed information to the Query for its execution.
+     *
+     * @return QueryInterface
+     */
+    public static function from(iterable $source, QueryContextInterface $context = null): QueryInterface;
```
Crell, line 65:
```suggestion
    public static function from(iterable $source, QueryContextInterface $context = null): static;
```

This is precisely the use case static returns were added for.

proposed/query.md
```diff
@@ -0,0 +1,173 @@
...
+
+    /**
+     * Applies a modifier to the Query.
+     *
+     * @param QueryModifierInterface $modifier
+     *      The actual modifier to apply.
+     *
+     * @return QueryInterface
+     */
+    public function applyModifier(QueryModifierInterface $modifier): QueryInterface;
```
Crell, line 89: Ibid.

proposed/query.md
alexandre-daubois: That's right, I updated it!

proposed/query.md
alexandre-daubois: Fixed as well. And also learnt what "ibid" stands for, thanks! :smile:

proposed/query.md
```diff
@@ -0,0 +1,173 @@
...
+     * @return QueryInterface
+     */
+    public static function from(iterable $source, QueryContextInterface $context = null): QueryInterface;
+
+    /**
+     * Get the context of the current Query, possibly modified by the latter.
+     *
+     * @return QueryContextInterface
+     */
+    public function getContext(): QueryContextInterface;
```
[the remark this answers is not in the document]
alexandre-daubois: I can understand this. Maybe `getContext`, `context` arguments and `QueryContextInterface` could simply be removed!

proposed/query.md
```diff
@@ -0,0 +1,173 @@
...
+use Psr\Query\QueryOperationInterface;
+
+/**
+ * Defines an object which will execute modifiers and operations to query data.
+ * A QueryInterface could receive any iterable depending on the implementation. This could go from the
+ * basic array of data to generators.
+ */
+interface QueryInterface
+{
+    /**namespace Psr\Query;
```
DerManoMann, line 54: This namespace looks like a cut&paste error?

alexandre-daubois: Yep, definitely. Thank you!

ashnazg (approved):
Looks interesting to me...

Crell:
This didn't go anywhere.

PR #1287, opened 2022-09-14
Fix ABNF in `phpdoc.md`

WinterSilence:
- Description cant exists without summary (`PHPDoc` rule)
- `eol` rule added to be compatible with PSR-12
- Summary can't be empty (`summary` rule)
- No spaces before  `tag-signature` (`tag-details` rule)
- Fix invalid definition of `keyword` rule
- Description can contain line tags and can't be empty (`tag-description` rule)

### DIFF
--- a/proposed/phpdoc.md
+++ b/proposed/phpdoc.md
@@ -205,7 +205,7 @@ interpreted as described in [RFC 2119][RFC2119].
       FQSEN    = fqnn / fqcn / constant / method / property  / function
       fqnn     = "\" [name] *("\" [name])
       fqcn     = fqnn "\" name
-      constant = (fqnn "\" / fqcn "::") name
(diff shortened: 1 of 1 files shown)


jaapio:
> Description can exists without summary (PHPDoc rule)

I'm not sure if I understand this rule correctly. In a docblock the first sentence is always recognized as the summary. The summary should be followed by a `.` (dot) and 2 newlines. The description is optional. But a description can never exist without a summary. 

> Description can contain line tags and can't be empty (tag-description rule)

the description on tags is optional. 

The rest of the changes look good to me.

WinterSilence:
@jaapio ups..i miss “t" in can't

jaapio (approved):

ashnazg (approved):

PR #1290, opened 2022-11-25
Proposal to dictate copyright ownership declaration

navarr:
Triggered by [php-fig/link-util#13](https://github.com/php-fig/link-util/issues/13), we need a policy for the declaration of copyright ownership

### DIFF
--- a/bylaws/006-licensing-policies.md
+++ b/bylaws/006-licensing-policies.md
@@ -1,7 +1,12 @@
 Licensing Policies
 ==================
 
-## 1. Code
+## 1. Copyright Ownership
+
+1. Anytime the copyright owner of code, documentation or otherwise is listed it
+   must be shown as "PHP Framework Interop Group and its contributors."
+
+## 2. Code
 
 1. All PHP FIG code must be licensed under the MIT license.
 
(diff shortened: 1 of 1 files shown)


ashnazg (approved):

vdelau:
Good proposal, but as it is a bylaw change this requires a vote.

ramsey:
I spoke with Pamela Chestek about this issue. She is a [board member at the OSI](https://opensource.org/board-member/pamela-chestek-2/) and has her own [private practice](https://www.chesteklegal.com).

She says:

1. This change is appropriate for the group to make.
2. The change may be retroactively added to all existing copyright notices on code owned by the project, provided it doesn't remove any existing copyright notices.

So, for example, if I had contributed code to PHP-FIG, and it had a notice like this:

```
Copyright (c) 2015 Ben Ramsey. All rights reserved.
```

PHP-FIG could update the notice to read:

```
Copyright (c) 2015 Ben Ramsey. All rights reserved.
Copyright (c) 2023 PHP Framework Interoperability Group and its contributors.
```

(FYI, "All rights reserved" is always implied, so you could add it or not.)

If the copyright notice already lists PHP-FIG, like [this one from the psr/log LICENSE file](https://github.com/php-fig/log/blob/fe5ea303b0887d5caefd3d431c3e61ad47037001/LICENSE#L1):

```
Copyright (c) 2012 PHP Framework Interoperability Group
```

PHP-FIG may update this to the following, without any legal problems:

```
Copyright (c) 2012 PHP Framework Interoperability Group and its contributors
```

ramsey (approved):

samdark (approved):

PR #1291, opened 2022-12-13
Update PSR-11-container.md

tspencer244:
Copy the wording given in the container meta document.

### DIFF
--- a/accepted/PSR-11-container.md
+++ b/accepted/PSR-11-container.md
@@ -48,9 +48,7 @@ A call to the `get` method with a non-existing id MUST throw a
 
 ### 1.3 Recommended usage
 
-Users SHOULD NOT pass a container into an object so that the object can retrieve *its own dependencies*.
-This means the container is used as a [Service Locator](https://en.wikipedia.org/wiki/Service_locator_pattern)
-which is a pattern that is generally discouraged.
+Users SHOULD NOT pass a container into an object, so the object can retrieve *its own dependencies*. Users doing so are using the container as a [Service Locator](https://en.wikipedia.org/wiki/Service_locator_pattern). [Service Locator](https://en.wikipedia.org/wiki/Service_locator_pattern) usage is generally discouraged.
 
 Please refer to section 4 of the META document for more details.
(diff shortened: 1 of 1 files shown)


tspencer244:
https://www.php-fig.org/psr/psr-11/meta/#4-recommended-usage-container-psr-and-the-service-locator

image

https://www.php-fig.org/psr/psr-11/

image

I found this wording confusing and wasn't sure whether it was recommending usage of service locator pattern or not recommending it :confused:

Crell:
Already approved PSRs are not subject to change.  They are frozen.

Using a service locator pattern is not-good, and should be avoided most of the time.  There's a few exceptions around lazy factories where it may be necessary, but those are fairly few.

tspencer244:
Sorry I probably confused the intent of this pull request mentioning my confusion around service locators.

The intent of this pull request was to correct a mistake in one of the documents.

If we look at the wording given in the main container document:

https://github.com/php-fig/fig-standards/blob/23d0b3e0e522b4c0276ad6ec1eced2884039a389/accepted/PSR-11-container.md?plain=1#L51

And then compare it to the wording the meta container document quotes the main container document as saying:

https://github.com/php-fig/fig-standards/blob/23d0b3e0e522b4c0276ad6ec1eced2884039a389/accepted/PSR-11-container-meta.md?plain=1#L76

They don't match and it just seems like it was a mistake for the main container document to have the wording as something that is very difficult to understand while the meta document states the information in a way that can be easily understood.

ralflang:
Actually I do not see the difference in meaning between the two wordings. Maybe I am too used to what general advice this section tries to convey.

tspencer244:
This is quite embarrassing but I actually have no idea what I was seeing wrong before. Looking at everything I have posted pictures of, pages linked to, code lines linked to - it all looks fine to me now :sweat_smile: 

I am very sorry for wasting all of your time on this!

Thank you very much for all the effort your team puts into establishing and maintaining these standards, they have really helped to improve the quality of life in my career :slightly_smiling_face:

PR #1292, opened 2023-02-03
Sync H1 and H2 Markdown headers

petk:
Hello, I was just browsing the PSRs and in the past there were some syncs done on the H1 and H2 Markdown headings. This fixes the rest. Issue is also that some titles are not displayed on the php-fig.org site if the headings don't follow some logic.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,7 +9,7 @@ discussion and formation stages involve everyone.
 
 See the [PHP-FIG FAQ](https://www.php-fig.org/faqs/) for more information.
 
-# Licensing
+## Licensing
 
 By contributing code you agree to license your contribution under the MIT
 license.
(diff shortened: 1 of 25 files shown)


drupol:
I think this could be automated with `prettier` (https://prettier.io/).

I quickly did a pull request to show how it can be done: https://github.com/drupol/fig-standards/pull/1

Let me know what you think all.

garak (approved):

Jean85 (approved):
Thank you!

Jean85:
> I think this could be automated with `prettier` (https://prettier.io/).

Well, not all unfortunately. This PR fixes some issues where the indentation level was wrong, and I don't think that can be automated...

drupol:
Oh ? Curious to see where you spotted that, can you add a comment where the issue is ? I will try to understand why and maybe provide a better solution.

CONTRIBUTING.md
```diff
@@ -63,7 +63,7 @@ Should you have any questions please contact the secretaries on info [at] php-fi
 [dot] org. Failure to comply with guidelines will result in revokation of merge
 access. Merge access is a privilege and not a right.
 
-# Tagging
+## Tagging
```
Jean85, line 66: All the fixes on this file where due to (logical) level errors. This cannot be automated, it's due to the content, not the form.

drupol: Oh right, i never though this would be the issue, this is actually a cool fix. It does that because there can only be one `# title` .

ashnazg:
Is this still open because of the discussion around automating it?

PR #1293, opened 2023-02-06
Fix psr/cache package name

petk:
There is psr/link mentioned in the PSR-6 instead of the psr/cache probably. And versions should be ^2.0 and ^3.0.

That is correct?


drupol (approved):

Crell (approved):
Oops.  This was a copy-pasta error on my part, I'm pretty sure.

Jean85 (approved):

PR #1296, opened 2023-02-27
Add errata for addition of parameter and return type declarations to PSR-7

weierophinney:
This patch adds section 7.2 to the PSR-7 metadocument, detailing the (proposed) changes to the specification to use explicit, language-level parameter and return type declarations. It documents the addition of parameter type declarations for the 1.1 release, the addition of return type declarations for 2.0, and how implementations may adopt one or both versions. It also documents and links to the (forthcoming) package releases within the primary document, detailing the changes in brief.

The format follows that of the already accepted PSR-13 evolution proposal.

See also:

- [Parameter type declaration additions patch](https://github.com/php-fig/http-message/pull/94)
- [Return type declaration additions patch](https://github.com/php-fig/http-message/pull/95)

### DIFF
--- a/accepted/PSR-7-http-message.md
+++ b/accepted/PSR-7-http-message.md
@@ -1903,3 +1903,6 @@ interface UploadedFileInterface
     public function getClientMediaType();
 }
 ```
+
+Since [psr/http-message version 1.1](https://packagist.org/packages/psr/http-message#1.1.0), the above interfaces have been updated to add argument type hints.
+Since [psr/http-message version 2.0](https://packagist.org/packages/psr/http-message#2.0.0), the above interfaces have been updated to add return type hints.
--- a/accepted/PSR-7-http-message.md
+++ b/accepted/PSR-7-http-message.md
@@ -1904,5 +1904,5 @@ interface UploadedFileInterface
 }
 ```
 
-Since [psr/http-message version 1.1](https://packagist.org/packages/psr/http-message#1.1.0), the above interfaces have been updated to add argument type hints.
-Since [psr/http-message version 2.0](https://packagist.org/packages/psr/http-message#2.0.0), the above interfaces have been updated to add return type hints.
+Since [psr/http-message version 1.1](https://packagist.org/packages/psr/http-message#1.1.0), the above interfaces have been updated to add argument type declarations.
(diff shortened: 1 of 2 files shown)


dbu (approved):
looks good to me. it is a documentation clarification to explain how to correctly implement the different versions of the PSR specification.

edit: just understood what you meant with "proposed". 1.1 and 2.0 of the code do not exist yet. but seems like a sound proposal to me. if we go forward with this, i will propose the same for PSR-18.

i wonder if the interfaces described in the document should match the latest version (aka have type declarations) instead of the oldest version? or is that explicitly wanted like this in the evolution process?

Crell:
For the other PSRs that have been updated, we've been leaving the spec text as-is, and just documenting the change in the metadocument.  So this is fine (or at least consistent) as is.

Crell (approved):

accepted/PSR-7-http-message.md
```diff
@@ -1903,3 +1903,6 @@ interface UploadedFileInterface
     public function getClientMediaType();
 }
 ```
+
+Since [psr/http-message version 1.1](https://packagist.org/packages/psr/http-message#1.1.0), the above interfaces have been updated to add argument type hints.
```
TimWolla, line 1907: It should not be called "type hint", because it's not just a "hint", the type is actually enforced.

navarr (approved):

accepted/PSR-7-http-message.md
TimWolla: To make sure it isn't missed: The same applies in the line below, for the return types.

ashnazg (approved):

accepted/PSR-7-http-message.md
weierophinney: Updated both to use the verbiage "type declarations" with 2b99fb2.

weierophinney:
Core Committee vote started 2023-03-13.

dbu:
> Core Committee vote started 2023-03-13.

can we add more changes to psr-7 1.1 and 2.0? i would love to add #1298 which brings no code changes, but would more precisely specify some behaviour.

vdelau (approved):

PR #1297, opened 2023-03-26
Change example class name

UVLabs:
The example class name for the file writer contradicts PSR-1 when it comes to the naming of classes.

PSR-1 states that:

**Class names MUST be declared in StudlyCaps.** but the example shows a class name that doesn't follow PSR-1.

### DIFF
--- a/accepted/PSR-4-autoloader.md
+++ b/accepted/PSR-4-autoloader.md
@@ -61,7 +61,7 @@ class name, namespace prefix, and base directory.
 
 | Fully Qualified Class Name    | Namespace Prefix   | Base Directory           | Resulting File Path
 | ----------------------------- |--------------------|--------------------------|-------------------------------------------
-| \Acme\Log\Writer\File_Writer  | Acme\Log\Writer    | ./acme-log-writer/lib/   | ./acme-log-writer/lib/File_Writer.php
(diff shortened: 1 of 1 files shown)


drupol (approved):

drupol:
I agree with the change, but I don't think validated PSRs document can be modified anymore.

rob006:
Isn't that intentional to clarify difference between PSR-4 and PSR-0?

samdark (approved):

UVLabs:
@rob006 PSR-0 is deprecated from since 2014, no need to show difference between something that you're trying to move away from

Also, PSR-4 states **"Underscores have no special meaning in any portion of the fully qualified class name."** anyone coming from PSR-0 can read that line make sense of it. But the example in PSR-4 is a contradiction, I say that because I got introduced to PSR-4 before PSR-1 because I simply wanted standardized autoloading...then I realized I've been breaking PSR-1 all along (don't ask how many products I need to update now to follow suite 😅).

So to help future devs trying to conform, I think this should be changed to avoid confusion.

rob006:
> no need to show difference between something that you're trying to move away from

I disagree - showing what has changed is crucial if you want to introduce new version of existing standard/API.

Also, PSR-4 does not depend on PSR-1, it is perfectly fine to use PSR-4 and have class named `File_Writer` at the same time. If people are interested in code style guidance, they should check respective PSR instead of relying on these examples.

PR #1298, opened 2023-03-31
Add errata for expected behaviour of UriInterface::withUserInfo to PSR-7

dbu:
This suggestion comes out of a discussion we had in https://github.com/php-http/discovery/issues/222

The "urlencode when necessary but don't double encode" behaviour is already implemented in laminas, guzzle and slim, and with https://github.com/Nyholm/psr7/pull/213/ was adopted in nyholm. It seems to be expected behviour, but the specification did not define it. Can we add this precision?

If i understand correctly, we are are not supposed to change the main document but only the meta document and the actual interface phpdoc in php-fig/http-message. If this suggestion seems acceptable, i can create the pull requests for http-message

### DIFF
--- a/accepted/PSR-7-http-message-meta.md
+++ b/accepted/PSR-7-http-message-meta.md
@@ -710,3 +710,15 @@ Implementers MAY add parameter types to their own packages in a new major releas
 Implementers are encouraged but not required to transition their packages toward the 2.0 version of the package at their earliest convenience.
 
 [1]: https://datatracker.ietf.org/doc/html/rfc7230#section-3.2
+
+### 7.3 Escaping User Info
+
+Some characters are reserved in the user info part of the authority section.
+According to (RFC3986 2.2 and 3.2.1)[https://www.rfc-editor.org/rfc/rfc3986], the reserved characters are `"/" / "?" / "#" / "[" / "]" / "@"`.
+Additionally, `:` must be encoded when in the username because it is used to separate username and password.
+
+`UriInterface::withUserInfo` MUST NOT double urlencode reserved characters.
+
+`UriInterface::getUserInfo` MUST urlencode the reserved characters when returning the authority.
+If there is a password, the `:` between username and password MUST NOT be urlencoded.
+
--- a/accepted/PSR-7-http-message-meta.md
+++ b/accepted/PSR-7-http-message-meta.md
@@ -717,8 +717,8 @@ Some characters are reserved in the user info part of the authority section.
 According to (RFC3986 2.2 and 3.2.1)[https://www.rfc-editor.org/rfc/rfc3986], the reserved characters are `"/" / "?" / "#" / "[" / "]" / "@"`.
 Additionally, `:` must be encoded when in the username because it is used to separate username and password.
 
-`UriInterface::withUserInfo` MUST NOT double encode reserved characters.
+`UriInterface::withUserInfo()` MUST NOT double encode reserved characters.
 
-`UriInterface::getUserInfo` MUST encode the reserved characters according to RFC3986 when returning the authority.
+`UriInterface::getUserInfo()` MUST encode the reserved characters according to RFC3986 when returning the authority.
 If there is a password, the `:` between username and password MUST NOT be encoded.
 
--- a/accepted/PSR-7-http-message-meta.md
+++ b/accepted/PSR-7-http-message-meta.md
@@ -717,8 +717,8 @@ Some characters are reserved in the user info part of the authority section.
 According to (RFC3986 2.2 and 3.2.1)[https://www.rfc-editor.org/rfc/rfc3986], the reserved characters are `"/" / "?" / "#" / "[" / "]" / "@"`.
 Additionally, `:` must be encoded when in the username because it is used to separate username and password.
 
-`UriInterface::withUserInfo` MUST NOT double urlencode reserved characters.
+`UriInterface::withUserInfo` MUST NOT double encode reserved characters.
 
-`UriInterface::getUserInfo` MUST urlencode the reserved characters when returning the authority.
-If there is a password, the `:` between username and password MUST NOT be urlencoded.
+`UriInterface::getUserInfo` MUST encode the reserved characters according to RFC3986 when returning the authority.
+If there is a password, the `:` between username and password MUST NOT be encoded.
 


accepted/PSR-7-http-message-meta.md
```diff
@@ -691,3 +691,14 @@ Further characters or sequences in header values should be rejected according
 to the HTTP specification.
 
 [1]: https://datatracker.ietf.org/doc/html/rfc7230#section-3.2
+
+### 7.2 urlencode User Info
```
dbu, line 695: the numbering will conflict with #1296 - i will adjust here once #1296 gets merged

accepted/PSR-7-http-message-meta.md
```diff
@@ -691,3 +691,14 @@ Further characters or sequences in header values should be rejected according
...
+
+In any release, `UriInterface::withUserInfo` MUST NOT double urlencode the username
+and password if they are already urlencoded.
```
TimWolla, line 698: I don't see how this can work. Is `h%65llo` already urlencoded or not?

rob006: What if username *is* `h%65llo` and it should be encoded to `h%2565llo`?

dbu: an actual username that looks like it is already urlencoded but is actually meant literal will need to be urlencoded by the caller already. 

this already is the case with all of the above mentioned implementations. we could mention that in the errata and maybe add it in the design decisions section.

in a totally strict world, we would have a value object for the username and password that knows if urlencoding already happened or not, but that would be a much bigger change and imho be overcomplicated. are you aware of psr7 implementations that behave differently, where for example passing `h%65llo` would be sent to the server as `h%2565llo` ?

TimWolla: > are you aware of psr7 implementations that behave differently?

I am not. I asked this question based on what I would expect as a user. And my first thought regarding your response was “this is bad” and Rob already explained why. Making *assumptions* based on whether some input data is already encoded or not is a common cause of security vulnerabilities. Sanitization / encoding is context dependent and thus must always happen at the output.

As a user I would expect the following:

1. I input some userinfo string in raw form.
2. If the userinfo string is embedded into the URI (i.e. in `__toString()`) it is appropriately encoded.
3. If I extract the userinfo string (e.g. because I want to provide it within an `Authorization` header), then I get the original raw form.

It's unfortunate that the `withUserInfo()` method takes a separate username and password, but the `getUserInfo()` method returns the full string. The RFC 3986 treats the userinfo as an opaque blob (which means that `getUserInfo()` technically behaves correctly).

dbu: your expection will not be met by any of the above implementations, because all of them urlencode directly in withUserInfo. given that, i think it will be difficult to change the behaviour, even though i agree that getUserInfo returning something else than was passed into withUserInfo sounds unexpected. 

we could add the clarification to getUserInfo that the username and password will be urlencoded, regardless of whether they have been passed to withUserInfo urlencoded or not.

apart from that amendment, i don't see anything else we can do about your input. changing this in all implementations is unrealistic and would make psr7 1.* and 2.* incompatible (on a small thing, but strictly incompatible). if somebody was already doing what you describe, they should right now do urldecode on the getUserInfo - if implementations would change it, the users must not decode anymore, or the `h%65llo` case would break.

TimWolla: > your expection will not be met by any of the above implementations, because all of them urlencode directly in withUserInfo.

Yes, I understand that.

> apart from that amendment, i don't see anything else we can do about your input.

My primary intention with the remark was that we *carefully* think about what to do here. I do not have a good answer for now, but I do not believe that this erratum as it is written is the right answer.

> would make psr7 1.* and 2.* incompatible 

Disregard versioning from your mental model. PSRs are immutable once written. The versioning is only about the technical implementation of the interfaces and must only implement what is already written in the PSR (i.e. proper types). See: https://www.php-fig.org/bylaws/psr-evolution/

rob006: > an actual username that looks like it is already urlencoded but is actually meant literal will need to be urlencoded by the caller already.

So the only reliable way of using this standard is to always encode username before passing it to `withUserInfo()`. In that case, what is the point of this "detect encoding" guidelines, if there is only a one safe way using this standard?

> are you aware of psr7 implementations that behave differently, where for example passing `h%65llo` would be sent to the server as `h%2565llo` ?

I'm not, but from my experience any this kind of conditional encoding solution is can of worms and sooner or later will lead to a problems and frustration. IMO standards like this should be strict - you either always encode, or never. "guess whether string is encoded or not" should not be a part of any standard :).

nicolas-grekas: Meanwhile I change the way nyholm/psr7 handles this, see https://github.com/Nyholm/psr7/pull/221, the reason why is discussed in https://github.com/Nyholm/psr7/issues/219, which gives some useful hint: because of the behavior guzzle/laminas have, they're not compatible with e.g. buzz NTLM auth. I argued there that maybe it's buzz that's not using the user info correctly, as it might be required to rawurldecode it before using it. But nothing in the PSR hints about this to this is a but dubious.

The point that triggered the discussion on nyholm/psr7 is the fact that passing a reserved char (in the sense of rfc3986) as user/password produces a broken URI.

If the PSR could be updated to clarify how this is supposed to behave, that'd close the topic.

One way is what other libs do:
- url-encode most chars. But this breaks use cases like buzz'.
- throw an exception when a reserved char is passed. This ship has sailed.
- specify that *reserved chars* MUST be encoded. This is the path taken by nyholm/psr7 as is preserves the use case of buzz.

TimWolla: > throw an exception when a reserved char is passed. This ship has sailed.

Even if the ship wouldn't have sailed, that never be correct: The `%` is a reserved character, because it is part of percent-encoding syntax. Thus if reserved characters would be rejected, then neither percent-encoded characters could be passed (because they would contain the reserved `%`), nor a raw `%`.

Given that PSR-7 is clear that `getQuery()` must return percent-encoded values, and that `withQuery()` must accept both encoded and non-encoded values, I tend to agree that this is also correct behavior for the userInfo. For users this means that they effectively always need to manually percent-encode first, because otherwise `h%65llo` would be ambiguous. Is it `hello` with an uselessly encoded `e` or is this `h`, `%`, `65llo`?

The UriInterface really is the “least good” part” of PSR-7, I remember some previous discussion on PHP-FIG discord that I participated in (May, 12 2022) where something else was unclear.

TimWolla: > I dont have to encode my query parameters either so why would I do it for the user Info?

@boesing You actually have to:

```
$uri = new Uri('http://127.0.0.1:8080');
$uri = $uri->withQuery('foo=b%61r');
```

Is this a query parameter `foo` with value `bar` or a query parameter foo with value `b%61r`? PSR-7 is clear it's the former, so if you want the latter, you have to encode it. Or ideally use `http_build_query()` which does the right thing.

nicolas-grekas: Just to prevent building on a wrong assumption: `%` is *not* a reserved character.

The [ANBF defines them as](https://www.rfc-editor.org/rfc/rfc3986#appendix-A):
```txt
   reserved      = gen-delims / sub-delims
   gen-delims    = ":" / "/" / "?" / "#" / "[" / "]" / "@"
   sub-delims    = "!" / "$" / "&" / "'" / "(" / ")"
                 / "*" / "+" / "," / ";" / "="
```

TimWolla: @nicolas-grekas Sorry, you are right of course. I should have checked this once more, instead of assuming. I've striked out that incorrect part of my comment. The remainder should be correct, though.

nicolas-grekas: Yep. So to me the PSR should define the behavior for reserved chars, because they are the ones that are really problematic for interop.

boesing: > > I dont have to encode my query parameters either so why would I do it for the user Info?
> 
> @boesing You actually have to:
> 
> ```
> $uri = new Uri('http://127.0.0.1:8080');
> $uri = $uri->withQuery('foo=b%61r');
> ```
> 
> Is this a query parameter `foo` with value `bar` or a query parameter `foo` with value `b%61r`? PSR-7 is clear it's the former, so if you want the latter, you have to encode it. Or ideally use `http_build_query()` which does the right thing.

Ha, dunno why I thought it accepts array, maybe good old non-PSR day memory. sorry.

TimWolla: > So to me the PSR should define the behavior for reserved chars

Except this also doesn't quite cut it, because the “setter” takes `username` and `password` separately, instead of a userinfo string. The `:` is a gen-delim and thus reserved, but specifically allowed to exist as a raw value within the userinfo. As a percent-encoded reserved character is not equivalent to the unencoded character.

Now RFC 3986 does not define that there the userinfo is a username and password separated with a colon. It defines it as a sequence of opaque bytes. Thus by separately encoding each of the username and password, I am unable to specify a userinfo that contains three colon separated parts. The previous implementation of nyholm/psr7 would have allowed for that (but then misconstructed the resulting URI).

nicolas-grekas: Actually the RFC does tell about the user colon password:)
But you're right, the missing spec is about the individual components.

TimWolla: > But you're right, the missing spec is about the individual components.

Yes, because the format is scheme-specific (as mentioned at the beginning of RFC 3986#3.2.1).

(This also made me realize a fun aspect of the `mailto:` scheme. The local-part of the email address is actually the `userinfo` part of the URI and that also nicely explains why the userinfo is part of the authority).

---------------

Regarding my confusion with `%` reserved or not:

Even if I was incorrect about the `%` being a reserved character, it nonetheless presents an issue with regard to specifying the behavior, as highlighted by my `getQuery()` example. Users have no other option than encoding the passed values themselves, if they want to ensure they are transmitted exactly how they need them to be transmitted. And at that point I'm wondering if this is solvable with an erratum or if the UriInterface should be fixed with a PSR that supersedes PSR-7.

shadowhand: As a user of PSR-7, the only thing that makes sense is for implementations of UriInterface to **always** encode what I pass into this method. It should not guess if it is encoded, it should not throw an exception, and it should not assume that is encoded.

The only logical thing that an implementation can do is assume raw input that MUST be encoded.

TimWolla: > the only thing that makes sense is for implementations of UriInterface to always encode what I pass into this method

This would prevent round-tripping of values. See the second paragraph in this comment:

https://github.com/zendframework/zend-diactoros/pull/248#issuecomment-322912952

shadowhand: It is highly unfortunate that `getUserInfo` does not decode. I have always found it incredibly annoying (and possibly short sighted) that URI parameters were not decoded on output.

However, I stand by argument because round tripping would become:

```php
$uri = $uri->withUserInfo(rawurldecode($previous->getUserInfo()));
```

accepted/PSR-7-http-message-meta.md
```diff
@@ -691,3 +691,14 @@ Further characters or sequences in header values should be rejected according
...
+
+In release 1.1, `UriInterface::withUserInfo` SHOULD urlencode username or password when
+they contain characters that are invalid in an URL and are not already urlencoded.
```
dbu, line 701: to sum up the above discussion: would it be correct to change this to something like: `SHOULD escape characters in in username and password that are defined as reserved in [RFC3986](https://www.rfc-editor.org/rfc/rfc3986#appendix-A)`

and accordingly below, and also in the rule about double-escaping?

nicolas-grekas: We might also say SHOULD NOT encode other characters, that would cover double encoding and a bit more, eg `\` which is encoded by guzzle (but not by nyholm's)

rob006: With encoding only reserved characters, if you pass `foo%3Abar:baz` as username, it will be encoded to `foo%3Abar%3Abaz`. How would you revert it back to `foo%3Abar:baz`?

dbu: @rob006 then the correct way to formulate would be: if it contains any reserved character, the whole value should be encoded? if you pass `foo%eAbar:baz`, the implementation has to assume that the string is not yet encoded, and should therefore also encode the `%` i think.

damn, i don't think there really is a good solution... because if we do that, the behaviour is different if the caller never encodes based on whether they pass `foo%3Abar` or `foo%3Abar:baz` (because the receiving server will urldecode and translate the `%3A`.

nonetheless, i think we should figure out the rule that makes for the least confusion and add an errata that clarifies how to do this. and in the interest of internal consistence and low surprises, i would go with the rules outlined on `getQuery`. (i am not sure what those rules mean for the `foo%3Abar:baz` example. but the current implementations i think will do a compromise that will not work, as they only encode the `:` but not the `%`)

rob006: There is simple and reliable solution: always encode. You build user info by encoding username and password, and implode them using `:`. If you want to retrieve username and password, you do the same backwards: explode results of `getUserInfo()` by `:`, and decode each part. Yes, it is inconvenient if you want to pass `getUserInfo()` result to `withUserInfo()`, but this is a result of asymmetry between these two methods, and can't be really fixed without serious BC break. But this use case could be simplified by adding `getUserName()` and `getPassword()` to interface: `$uri->withUserInfo($previous->getUserName(), $previous->getPassword()))`.

dbu: by `you`, do you mean the caller or the psr7 implementation?

given that all the largely used implementations discussed above do encode conditionally, i think clarifying that the arguments always must be encoded by the caller is not a viable option as it conflicts with existing implementations. if implementations were to change that behaviour, this would be a nasty hidden BC break for users.

my aim with this pull request is to try to clarify what implementations should do in a way that creates as little friction as possible.

i do think there would be merit for a new PSR with stricter and more robust rules, but that is out of scope for this pull request and discussion. (please feel free to cite this discussion as one motivation why there should be a new PSR for Uri, should you propose such a PSR)

rob006: > by `you`, do you mean the caller or the psr7 implementation?

I mean implementation. So `withUserInfo()` should always encode and `getUserInfo()` should always return this encoded value.  `getUserName()` and `getPassword()` would return **decoded** values (in the same form as passed to `withUserInfo()`).

dbu: that would be consistent and deterministic. we can't add new methods afaik, though if it is allowed for a version 2.0 it would make sense.

but i am unsure if that is what the current implementations do. the phpdoc on UriInterface::getQuery says `The value returned MUST be percent-encoded, but MUST NOT double-encode any characters.` i am not 100% sure i understand correctly, but i think when this was written, the `foo%3Abar:baz` case was not considered.

dbu:
@weierophinney do you have any input on this? how should `withUserInfo` (and also with query) handle a string like `h%65llo:bar`? if i understand correctly, current implementations would urlencode the `:` but not the `%` which is unlikely to be correct.

on getQuery there is a statement that double encoding must be avoided, which afaik made the implementors skip encoding of `%`.

RE the discussion above: i think we should not discuss how PSR-7 1.0 could have been done better, but find a pagmatic solution for moving forward with the current "state of the art" to avoid friction for the current users of PSR-7.

weierophinney:
@dbu — Essentially, we should have consistency of behavior between the various methods that modify different parts of the URI. We were explicit in the way the path, query parameters, and fragment were to handle encoding; we should mimic that for the user info as well, which means that (a) the values returned from `get*()` operations MUST be properly encoded, and (b) passing those values to a `with*()` method MUST be allowed, and MUST NOT double encode. (While I appreciate what @shadowhand notes about how it can be unintuitive to call a `get*()` method and not get back the raw value, the point of these methods, when we created the spec, was to return something _valid for inclusion in a URL or URL fragment_. We recognized that this particular interface was not ideal, but it was suited for the domain — HTTP messages — and we didn't want to block on having a full-blown spec for covering the entierty of RFC-3986. I always anticipated that we'd have a URI PSR at some point, and a replacement for PSR-7 that consumed it; it's just never happened.)

In this particular case, however, the characters that MUST be urlencoded differ from other parts of the URI, according to the ABNF. The relevant line of the ABNF reads as follows:

```text
userinfo      = *( unreserved / pct-encoded / sub-delims / ":" )
```

Note that `:` is explicitly allowed here, and should NOT be urlencoded. (Looking at Diactoros, we're doing this portion wrong!)

I would also argue that this would qualify for a new _minor_, but not a new _major_ release of the spec, because it corrects an _issue_ in the specification. Adopting the new minor might indicate that an existing implementation is broken currently, but fixing the implementation to conform would mean it still conforms to the _previous_ versions of the spec as well. This means that since we just released 1.1 and 2.0, we'd need to do 1.2 and 2.1 versions.

shadowhand:
> I always anticipated that we'd have a URI PSR at some point, and a replacement for PSR-7 that consumed it; it's just never happened.

As an aside, I think it is time for this process to begin. `UriInterface` is (probably) the weakest part of PSR-7, at least from a developer-user standpoint. (I don't know if it could be replaced by itself in a PER, or if we need a whole new PSR to replace PSR-7.)

Crell:
This sort of clarification belongs in Errata, not an update to the spec itself.

weierophinney:
> This sort of clarification belongs in Errata, not an update to the spec itself.

That's... what's being proposed, @Crell.

Crell:
I was referring to this:

> This means that since we just released 1.1 and 2.0, we'd need to do 1.2 and 2.1 versions.

Which I don't think is necessary.  Errata don't have tagged versions.

weierophinney:
@Crell — ah, good point. :)

dbu:
i updated the wording to an errata without new versions and to reflect what we discussed. wdyt?

i do not explicitly cover the "mixed case" `fo%65bar@baz` . an errata for these should also cover the other places like getQuery that have the same problem. 

> Note that : is explicitly allowed here, and should NOT be urlencoded.

@weierophinney `:` is used to separate username and password. in the username it must be encoded, otherwise the second part of the username would be considered part of the password.

weierophinney:
this looks good! I will bring it up for discussion this week, with an expected voting period starting two weeks from today.

weierophinney:
I've opened the discussion period:

- https://groups.google.com/g/php-fig/c/gX0zLyKe_xo/m/4z_m3qeRAgAJ

heiglandreas (commented):
Is URLencode the right tool? Or should this not rather use rawurlencode instead? The difference being whether spaces are encoded as `+` (urlencode) or `%20` (rawurlencode). The former can lead to disambiguity as a `+` can be a valid part of a user-info which would then be mistaken for a space

heiglandreas:
Perhaps just avoiding `urlencode` as word might be sufficient as that triggered me to think one could use the function of that name...

Perhaps something along the lines of "Encoding the URL according to RFC3986" might also be an option

accepted/PSR-7-http-message-meta.md
```diff
@@ -710,3 +710,15 @@ Implementers MAY add parameter types to their own packages in a new major releas
...
 
 [1]: https://datatracker.ietf.org/doc/html/rfc7230#section-3.2
+
+### 7.3 Escaping User Info
+
+Some characters are reserved in the user info part of the authority section.
+According to (RFC3986 2.2 and 3.2.1)[https://www.rfc-editor.org/rfc/rfc3986], the reserved characters are `"/" / "?" / "#" / "[" / "]" / "@"`.
+Additionally, `:` must be encoded when in the username because it is used to separate username and password.
+
+`UriInterface::withUserInfo` MUST NOT double urlencode reserved characters.
```
dbu, line 720:
```suggestion
`UriInterface::withUserInfo` MUST NOT double escape reserved characters.
```

accepted/PSR-7-http-message-meta.md
```diff
@@ -710,3 +710,15 @@ Implementers MAY add parameter types to their own packages in a new major releas
...
+
+`UriInterface::getUserInfo` MUST urlencode the reserved characters when returning the authority.
```
dbu, line 722:
```suggestion
`UriInterface::getUserInfo` MUST escape the reserved characters according to RFC3986 when returning the authority.
```

accepted/PSR-7-http-message-meta.md
```diff
@@ -710,3 +710,15 @@ Implementers MAY add parameter types to their own packages in a new major releas
...
+If there is a password, the `:` between username and password MUST NOT be urlencoded.
```
dbu, line 723:
```suggestion
If there is a password, the `:` between username and password MUST NOT be escaped.
```

accepted/PSR-7-http-message-meta.md
```diff
@@ -710,3 +710,15 @@ Implementers MAY add parameter types to their own packages in a new major releas
 Implementers are encouraged but not required to transition their packages toward the 2.0 version of the package at their earliest convenience.
 
 [1]: https://datatracker.ietf.org/doc/html/rfc7230#section-3.2
+
+### 7.3 Escaping User Info
+
+Some characters are reserved in the user info part of the authority section.
+According to (RFC3986 2.2 and 3.2.1)[https://www.rfc-editor.org/rfc/rfc3986], the reserved characters are `"/" / "?" / "#" / "[" / "]" / "@"`.
+Additionally, `:` must be encoded when in the username because it is used to separate username and password.
```
dbu, line 718:
```suggestion
Additionally, `:` must be escaped when in the username because it is used to separate username and password.
```

dbu (commented):
rewording suggestions to address @heiglandreas feedback.

accepted/PSR-7-http-message-meta.md
weierophinney: I would change the word escape to encode.

accepted/PSR-7-http-message-meta.md
weierophinney: Same comment.

accepted/PSR-7-http-message-meta.md
weierophinney: Same comment.

dbu: okay

dbu:
changed to "encode" / "encoding"

accepted/PSR-7-http-message-meta.md
```diff
@@ -710,3 +710,15 @@ Implementers MAY add parameter types to their own packages in a new major releas
...
+
+`UriInterface::withUserInfo` MUST NOT double encode reserved characters.
+
+`UriInterface::getUserInfo` MUST encode the reserved characters according to RFC3986 when returning the authority.
```
zonuexe, line 722: Is it possible to end the method name with `()`? The PSR-7 recommendation also writes method names in this style.

https://github.com/php-fig/fig-standards/blob/993c15f82058d9ec2504bb81e00f8b4061971ea5/accepted/PSR-7-http-message.md?plain=1#L272

accepted/PSR-7-http-message-meta.md
```diff
@@ -710,3 +710,15 @@ Implementers MAY add parameter types to their own packages in a new major releas
...
 
 [1]: https://datatracker.ietf.org/doc/html/rfc7230#section-3.2
+
+### 7.3 Escaping User Info
+
+Some characters are reserved in the user info part of the authority section.
+According to (RFC3986 2.2 and 3.2.1)[https://www.rfc-editor.org/rfc/rfc3986], the reserved characters are `"/" / "?" / "#" / "[" / "]" / "@"`.
+Additionally, `:` must be encoded when in the username because it is used to separate username and password.
+
+`UriInterface::withUserInfo` MUST NOT double encode reserved characters.
```
dbu, line 720: doing this as in PSR-7 itself, omitting the parameters of the method:

```suggestion
`UriInterface::withUserInfo()` MUST NOT double encode reserved characters.
```

accepted/PSR-7-http-message-meta.md
dbu, line 722:
```suggestion
`UriInterface::getUserInfo()` MUST encode the reserved characters according to RFC3986 when returning the authority.
```

dbu (commented):
@zonuexe thanks for the input, that seems consistent with the spec. i adjust the method names accordingly

mbniebergall (approved):
Voting passed.

PR #1301, opened 2023-05-13
Update PSR-22 - Application Tracing

AAllport:
After far too long, I've moved all the current work from the Google Doc to the Markdown file.
I have double-checked the meta-document, and I am still happy with its content

### DIFF
--- a/proposed/tracing.md
+++ b/proposed/tracing.md
@@ -16,7 +16,8 @@ interfaces and functionality first.
   Eg. Laravel, Symphony, CakePHP, Slim
 * **Library** - Any library that developers may include that adds additional functionality.
   Eg. Image Manipulation, HTTP Clients, 3rd Party SDKs
-* **Provider** - An organization that offers APM (Application Performance Monitoring) as a service. Typically, via a composer package
+* **Provider** - An organization that offers APM (Application Performance Monitoring) as a service. Typically, via a
+  composer package
 
 ## 2. Interfaces
 
@@ -26,6 +27,7 @@ See: [AAllport/psr-tracing - SpanInterface.php](https://github.com/AAllport/psr-
 
 A span represents a single operation within a trace. A trace is a collection of spans.
 A fully formed span will consist of the following data:
+
 - Name
 - Attributes, an associative array representing key-value pairs
 - Start Time
@@ -33,27 +35,30 @@ A fully formed span will consist of the following data:
 - Outcome
 - Exceptions
 
-Upon creation, a new span will not be activated or have any attributes set. This allows the user to populate the span with data.
-It is RECOMMENDED that users supply a span with all relevant attributes shortly after creation, notably before activation. However, Providers SHOULD make every effort to persist attribute data if provided after a span is activated.
+Upon creation, a new span will not be activated or have any attributes set.
+This allows the user to populate the span with data.
+It is RECOMMENDED that users supply a span with all relevant attributes shortly after creation, notably before
+activation.
+However, Providers SHOULD make every effort to persist attribute data if provided after a span is activated.
 
 To track the time a piece of work takes, a user will call `SpanInterface::activate()`.
 This will set the start time and push the current span into whichever context propagation system the provider chooses.
 Users MAY call `SpanInterface::start()` to populate the span with a start time, but not pushed into context propagation.
-Once complete, the user would call SpanInterface::finish() to set the end-time and “pop” the span from the current stack.
+Once complete, the user would call SpanInterface::finish() to set the end-time and “pop” the span from the current
+stack.
 
-Spans MUST support creating a child span via `SpanInterface::createChild(string $spanName): SpanInterface`. This SHOULD use `TracerInterface::createSpan()`, then set the subsequent span's parent appropriately.
+Spans MUST support creating a child span via `SpanInterface::createChild(string $spanName): SpanInterface`.
+This SHOULD use `TracerInterface::createSpan()`, then set the subsequent span's parent appropriately.
 
 A typical use would look like the following:
 
 ```php
-
 function imgResize($size=100) {
     $span = $this->tracer->createSpan('image.resize')
         ->setAttribute('size',$size)
         ->activate();
 
     try{
-   
       //Resize the image
       return $resizedImage;
     
@@ -65,11 +70,13 @@ function imgResize($size=100) {
         $span->finish();
     }    
 }
-
 ```
 
-This PSR does not dictate how a span’s internals should be represented, other than it SHOULD implement the SpanInterface. However, it is RECOMMENDED to conform to the [W3C TraceContext specification](https://www.w3.org/TR/trace-context/).
-Providers MUST implement the `SpanInterface::toTraceContextHeaders()` method. At a minimum, this should return an empty array. Most commonly, providers will return traceParent and baggage headers to pass on to child services.
+This PSR does not dictate how a span’s internals should be represented, other than it MUST implement the SpanInterface.
+However, it is RECOMMENDED to conform to the [W3C TraceContext specification](https://www.w3.org/TR/trace-context/).
+Providers MUST implement the `SpanInterface::toTraceContextHeaders()` method.
+At a minimum, this should return an empty array.
+Most commonly, providers will return traceParent and baggage headers to pass on to child services.
 
 Providers MUST add the following function signatures to allow data retrieval from their span:
 - `getAttribute(string $key): null|string|int|float|bool|Stringable;`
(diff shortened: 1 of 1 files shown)


proposed/tracing.md
```diff
@@ -10,14 +10,103 @@ interfaces and functionality first.
...
+
+A typical use would look like the following:
+
+```php
+function imgResize($size=100) {
+    $span = $this->tracer->createSpan('image.resize')
+        ->setAttribute('size',$size)
+        ->activate();
+
+    try{
```
samdark, line 61:
```suggestion
    try {
```

proposed/tracing.md
```diff
@@ -10,14 +10,103 @@ interfaces and functionality first.
...
+      //Resize the image
```
samdark, line 62:
```suggestion
      // Resize the image
```

proposed/tracing.md
```diff
@@ -10,14 +10,103 @@ interfaces and functionality first.
...
+        // Ideally, you would attach the exception to the span here
+        $span->setStatus(SpanInterface::STATUS_ERROR)
+             ->addException($e);
+    } finally {
+        $span->finish();
+    }    
+}
+```
+
+This PSR does not dictate how a span’s internals should be represented, other than it MUST implement the SpanInterface.
```
samdark, line 75:
```suggestion
This PSR does not dictate how a span’s internals should be represented, other than it MUST implement the `SpanInterface`.
```

proposed/tracing.md
```diff
@@ -10,14 +10,103 @@ interfaces and functionality first.
...
+However, it is RECOMMENDED to conform to the [W3C TraceContext specification](https://www.w3.org/TR/trace-context/).
+Providers MUST implement the `SpanInterface::toTraceContextHeaders()` method.
+At a minimum, this should return an empty array.
+Most commonly, providers will return traceParent and baggage headers to pass on to child services.
+
+Providers MUST add the following function signatures to allow data retrieval from their span:
```
samdark, line 81: It is better to provide an interface similar to what other PSRs do instead of defining a set of methods w/o it. See https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-7-http-message.md#3-interfaces

proposed/tracing.md
```diff
@@ -10,14 +10,103 @@ interfaces and functionality first.
...
+- `getAttribute(string $key): null|string|int|float|bool|Stringable;`
+- `getAttributes(): iterable;`
+- `getParent(): SpanInterface;`
+- `getChildren(): array;`
+
+The purpose of these functions is to allow data to be read from providers’ spans in a clear, uniform manner.
+`SpanInterface::getParent` MAY return null if no parent is present or the span has been instantiated outside the relevant adapter’s methods.
```
samdark, line 88:
```suggestion
`SpanInterface::getParent()` MAY return null if no parent is present or the span has been instantiated outside the relevant adapter’s methods.
```

proposed/tracing.md
```diff
@@ -10,14 +10,103 @@ interfaces and functionality first.
...
+
+Upon creation, a new span will not be activated or have any attributes set.
+This allows the user to populate the span with data.
+It is RECOMMENDED that users supply a span with all relevant attributes shortly after creation, notably before
+activation.
+However, Providers SHOULD make every effort to persist attribute data if provided after a span is activated.
+
+To track the time a piece of work takes, a user will call `SpanInterface::activate()`.
+This will set the start time and push the current span into whichever context propagation system the provider chooses.
+Users MAY call `SpanInterface::start()` to populate the span with a start time, but not pushed into context propagation.
```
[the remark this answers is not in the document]
alekitto: Neither `start` nor `activate` accept the `$time` parameter.
`start` sets the start time of the span, `activate` pushes the span into the context, marking it as the "currently active span" and sets the start time if not already set.

proposed/tracing.md
AAllport: Worth noting these methods do exist in the [interface](https://github.com/AAllport/psr-tracing/blob/main/src/SpanInterface.php).
I will leave this comment open until I rephrase it slightly to clarify that the paragraph is drawing attention as opposed to an additional requirement.


> The following methods are aimed at retrieving data from a providers span for programmatic use

proposed/tracing.md
```diff
@@ -10,14 +10,103 @@ interfaces and functionality first.
 
 [RFC 2119]: http://tools.ietf.org/html/rfc2119
 
-## Goal
+## 1. Definitions
 
-TBD
+* **Framework** - An application framework (or micro-framework) that runs a developers code.  
+  Eg. Laravel, Symphony, CakePHP, Slim
```
jkrzefski, line 16: ~Symphony~
Symfony

proposed/tracing.md
```diff
@@ -10,14 +10,103 @@ interfaces and functionality first.
...
+However, a reasonable attempt SHOULD be made to return a `SpanInterface` where possible.
+Similarly, `SpanInterface::getChildren()` MAY return an empty array if child spans have been created outside the relevant adapter's methods.
+For continual use of a secondary provider, it is RECOMMENDED to create a separate Tracer and use the `MultiTracer` described rather than relying on these methods.
+
+## 2.2 Tracer
+
+See: [AAllport/psr-tracing - TracerInterface.php](https://github.com/AAllport/psr-tracing/blob/main/src/TracerInterface.php)
+
+SDKs are expected to provide a concrete Tracer via whichever methods are most appropriate for the frameworks they support (eg, PSR-11 container).
+Libraries using this PSR SHOULD implement [TracerAwareTrait](#31-tracerawaretrait).
```
drealecs, line 98: I guess here we should mention `TracerAwareInterface` instead of the trait.
I believe that it might be better to avoid the AwareInterface pattern and have just a `TracerManager` class that can be global. My assumption is that providers are provided as a php extension or as a wrapper over that so exposing it as an global class might fit better. Or inject it in the container and obtain it where is needed.

proposed/tracing.md
```diff
@@ -10,14 +10,103 @@ interfaces and functionality first.
...
+- Exceptions
+
+Upon creation, a new span will not be activated or have any attributes set.
+This allows the user to populate the span with data.
+It is RECOMMENDED that users supply a span with all relevant attributes shortly after creation, notably before
+activation.
+However, Providers SHOULD make every effort to persist attribute data if provided after a span is activated.
+
+To track the time a piece of work takes, a user will call `SpanInterface::activate()`.
+This will set the start time and push the current span into whichever context propagation system the provider chooses.
```
drealecs, line 45: Is this related to `getCurrentSpan()` from `TracerInterface`? After "activation" the tracer should return it?

proposed/tracing.md
```diff
@@ -10,14 +10,103 @@ interfaces and functionality first.
...
+A fully formed span will consist of the following data:
+
+- Name
+- Attributes, an associative array representing key-value pairs
+- Start Time
+- End Time
+- Outcome
+- Exceptions
+
+Upon creation, a new span will not be activated or have any attributes set.
```
drealecs, line 38: I think it's good to keep mutability to a minimum on value objects.
How about?:
- activate the span and start time as constructor time
- have an ending method to fill in the outcome and eventual exceptions, end the time and deactivate the span.
  - I used something in a wrapper like `end()` and `endWithError($exception)` to accomplish this.

proposed/tracing.md
alekitto: I would personally avoid a `TracerManager`: DI containers can handle it well enough.

Maybe this can be rephrased as:
`It is RECOMMENDED for the libraries using this PSR implement TracerAwareInterface in order to ease the injection of a tracer` 

or something similar

proposed/tracing.md
alekitto: Activate and start are two different operations. In the majority of cases they are executed at the same time and immediately after the construction, but not always:
For example, you can just prepare an HTTP request via a client and construct the span but without firing it or processing it (no start nor activate should be called).
Then, when sending the request, you should call the `start` method, but you may not call `activate` because the request is in progress in background. Only when you need to wait and process the response you should call `activate` (which marks the span as the currently active span in the context).

The same applies for the outcome and exceptions: adding an exception does not mean that the outcome of the operation is not OK and does not imply the ending of the span; just like ending a span without adding any exception does not imply that the operation has been completed without errors.

Outcomes, exceptions and start/end times are completely unrelated concepts and, unfortunately, simplifying them leaves out some useful cases.

polishdeveloper:
@AAllport do you have any timeline on this PSR? I'm currently working on tracing and I got excited to learn that there is a PSR standard I could stick to. Is there anything I could help with to push it forward?

BafS:
@AAllport what is the status of this PR? Do you plan to work on it still?

BafS:
I'll take this as a "no" :)

PR #1302, opened 2023-06-04
Don't suggest deprecated `Serializable` interface for a PSR-14 Event

chalasr:


samdark (approved):

Crell (changes requested):
This is a change in the spec itself, technically, so would not be allowed.  I don't think that's an issue, though, since it doesn't say implementers SHOULD use Serializable, just that they MAY do so, along with various other options.  The intent is simply to say "whatev."  Removing Serializable is not needed for that, just like no one should be using `__sleep()` or `__wakeup()` these days but we're not removing those.

This is a won't-fix from me.  (The docblock newline is fine to make its own PR, though.)

chalasr:
Alright, makes sense. Maybe better give no hint at all and just remove that sentence altogether instead?

Crell:
No, that's a substantive change to the spec.  Those aren't allowed, even if it might make sense casually.

PR #1304, opened 2023-06-10
Update PSR-11-container-meta.md

MarwanAlsoltany:
Fix typos and punctuation.

### DIFF
--- a/accepted/PSR-11-container-meta.md
+++ b/accepted/PSR-11-container-meta.md
@@ -109,11 +109,11 @@ In the `BadExample` you should not inject the container because:
 - it is not directly clear from your code that the `BadExample` class will need
   the "db" service. Dependencies are hidden.
 
-Very often, the `ContainerInterface` will be used by other packages. As a end-user
+Very often, the `ContainerInterface` will be used by other packages. As an end-user
 PHP developer using a framework, it is unlikely you will ever need to use containers
 or type-hint on the `ContainerInterface` directly.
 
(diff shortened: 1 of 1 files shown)


samdark (approved):

accepted/PSR-11-container-meta.md
```diff
@@ -226,7 +226,7 @@ As a result, the `ContainerInterface` contains two methods:
 
 While `ContainerInterface` only defines one mandatory parameter in `get()`, it is not incompatible with
 existing containers that have additional optional parameters. PHP allows an implementation to offer more parameters
-as long as they are optional, because the implementation *does* satisfy the interface.
+as long as they are optional because the implementation *does* satisfy the interface.
```
Crell, line 229: I don't think this change is necessary.  It reads better to me with the comma.  (I think this is one of those cases where either way is technically language-legal, because English.)

Jean85: +1 on this from me.

MarwanAlsoltany: @Crell, @Jean85 I reverted this change back as I used to be. I still think it doesn't read that well though, a better solution would be to make it as follows:
```
While `ContainerInterface` only defines one mandatory parameter in `get()`, it is not incompatible 
with existing containers that have additional optional parameters. PHP allows an implementation to offer more parameters 
as long as they are optional (because the implementation *does* satisfy the interface).
```

Crell (approved):

PR #1306, opened 2023-08-03
Replace idiom for clarity

peterhauke:
Replace "roll" with "build" (alternatives considered were "create" and "develop")

"roll their own" is not a very common idiom and, in fact, I thought at first that it was a mistake and the intention was "roll out their own caching libraries"


Crell (approved):

Crell:
I believe this qualifies as non-substantive, especially since it's just in the preamble, nothing in the spec definition itself.

mbniebergall (approved):
Approving wording update

PR #1307, opened 2023-08-16
PSR-19: @selfdoc

Chi-teck:
**Rationale:**
1. This purpose of  `@selfdoc` is in a certain sense similar to `mixed` type hint in PHP. When a structural element has no docblock it is not clear whether the developer has forgotten about it or has explicitly chosen not to provide a documentation.

2. Another issue it can solve is nasty code formatting. Consider the following example.
```php
/**
 * @param Foo $foo
 */
protected $foo;

protected Bar $bar;

/**
 * @param Baz $baz
 */
protected $baz;
```

One of these properties has no docblock as it really not needed here. However the code formatting brings some visual dissonance as the vertical rhythm is broken.

With `@selfdoc` it looks much better.
```php
/**
 * @param Foo $foo
 */
protected $foo;

/**
 * @selfdoc
 */
protected Bar $bar;

/**
 * @param Bas $baz
 */
protected $baz;
``` 

3. PHP code sniffers and static analyzers can figure out if a structural element without a docblock is fully documented through PHP typehints. What they can't do is to determine if it still needs textual description of what it does. Only developer can decide whether the purpose of the structural element is clean and does not require any additional explanations. `@selfdoc` can instruct sniffers and analyzers to not complain on such elements.

### DIFF
--- a/proposed/phpdoc-tags.md
+++ b/proposed/phpdoc-tags.md
@@ -27,12 +27,13 @@ PSR-19: PHPDoc tags
   - [5.11. @property](#511-property)
   - [5.12. @return](#512-return)
   - [5.13. @see](#513-see)
-  - [5.14. @since](#514-since)
-  - [5.15. @throws](#515-throws)
-  - [5.16. @todo](#516-todo)
-  - [5.17. @uses](#517-uses)
-  - [5.18. @var](#518-var)
-  - [5.19. @version](#519-version)
+  - [5.14. @selfdoc](#514-selfdoc)
+  - [5.15. @since](#515-since)
+  - [5.16. @throws](#516-throws)
+  - [5.17. @todo](#517-todo)
+  - [5.18. @uses](#518-uses)
+  - [5.19. @var](#519-var)
+  - [5.20. @version](#520-version)
 
 ## 1. Introduction
 
@@ -769,7 +770,36 @@ function count()
 }
 ```
 
-### 5.14. @since
+### 5.14. @selfdoc
+
+The @selfdoc tag denotes self documented "Structural Elements".
+
+#### Syntax
+
+    @selfdoc
+
+#### Description
+
+The @selfdoc tag can be used to indicate that the documentation for the associated
+"Structural Element" was omitted intentionally as its signature provides all required
+information.
+
+#### Examples
+
+The below function is documented through PHP typehints and its name clearly states
+what it does.
+
+```php
+/**
+ * @selfdoc
+ */
+function count(array $items): int
+{
+    <...>
+}
+```
+
+### 5.15. @since
 
 The @since tag is used to denote _when_ an element was introduced or modified,
 using some description of "versioning" to that element.
@@ -819,7 +849,7 @@ class Foo
 }
 ```
 
-### 5.15. @throws
+### 5.16. @throws
 
 The @throws tag is used to indicate whether "Structural Elements" throw a
 specific type of Throwable (exception or error).
@@ -862,7 +892,7 @@ function count($items)
 }
 ```
 
-### 5.16. @todo
+### 5.17. @todo
 
 The @todo tag is used to indicate whether any development activities should
 still be executed on associated "Structural Elements".
@@ -894,7 +924,7 @@ function count()
 }
 ```
 
-### 5.17. @uses
+### 5.18. @uses
 
 Indicates whether the current "Structural Element" consumes the
 "Structural Element", or project file, that is provided as target.
@@ -947,7 +977,7 @@ function executeMyView()
 }
 ```
 
-### 5.18. @var
+### 5.19. @var
 
 You may use the @var tag to document the "Type" of the following
 "Structural Elements":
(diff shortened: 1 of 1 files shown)


jaapio:
I'm sorry I do not see the need for such annotation. For me it's more verbose than needed to use this. 

The purpose of this psr is not to introduce new types of tags but to standardize the tags used in the various open source projects. Unless I'm wrong this is a proposal for a new tag which is not part of our scope.

ashnazg:
You're describing the use case equivalent of a blank page in a document showing "THIS PAGE INTENTIONALLY LEFT BLANK".  This roughly describes the _organic_ growth of using a _standalone_ `@inheritDoc` tag (https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#42-using-the-inheritdoc-inline-tag-to-augment-a-description), which was not an original tag in the early **phpDocumentor** days... only the _inline_ `{@inheritDoc}` tag existed originally.

Over time, I saw several big projects start to use 
```
/**
 * @inheritDoc
 */
```
 docblocks simply to satisfy whatever static analyzers / doc builders they were using, rather than actually trying to say "this element inherits info from a parent".

Crell:
Regardless of the chicken-and-egg scoping question, I would also be against this tag.  It adds three extra lines of no-op text to tell you that there's nothing to tell you.  Actually it's worse than that, because as a docblock it would get parsed by the engine (by default), even if it's then never used.  So... I'm firmly -1 on this tag on its own merits, regardless of anything else.

Chi-teck:
> as a docblock it would get parsed by the engine

From my experience without  `{@selfdoc}` the engine will likely parse some dummy note instead. Like following.
```php
/**
 * The bar.
 */
protected Bar $bar;
 ```
 Developers have to add such docblocks to please PHP analyzers and sniffers.

**Comment by Chi-teck** [comment]
@ashnazg don't you think that using `@inheritDoc` is semantically incorrect for elements that are not inherited?

**Comment by Crell** [comment]
None of the SA tools I use have such a requirement.  That requirement is unique to Drupal and its own custom coding standards.

To be fair, Drupal's coding standards were mostly developed for PHP 4, and requiring docblocks on absolutely everything made sense.  There was no other way to get type information of any kind, and encouraging good documentation was one of the advantages of Drupal that I often argued for.  (I was one of the lead developers of Drupal for nearly a decade.)

But it's now 2023, almost all type information is available in the language itself (with the limited exception of generics and array types), and good variable names are a thing.

The issue here isn't that we need a new tag; it's that Drupal needs to get with the times and change its documentation requirements.

**Comment by Niklan** [comment]
It's not a Drupal requirement, either. It's just a preference that I've also like.

It's an optional doc block, not required. So if someone doesn't need it, it's fine.

Personally, I've also like that proposal. I'm working with Drupal for ~10 years now, and it's kind of a habit already to have all code have some doc block. I don't like how the code looks with random doc blocks where they only needed. I prefer all code to be consistent — all properties, methods and functions having a doc block. It is much easier for me to visually «parse» such code. Currently, it forces writing pointless comments like one that already mentioned with entity type manager.

As you said, PHP currently allows using type hints and other tools to make it clear what a variable is, but I still want to add some comment to it just for consistency.

For example, such code simply distracts me while I visually analyze the code:

```php
protected Foo $foo;
protected Bar $bar;

/**
 * Something important here.
 */
protected Baz $baz;
protected Quux $quux;
```

compared to:

```php
/**
 * {@selfdoc}
 */
protected Foo $foo;

/**
 * {@selfdoc}
 */
protected Bar $bar;

/**
 * Something important here.
 */
protected Baz $baz;

/**
 * {@selfdoc}
 */
protected Quux $quux;
```

I like the way that all code is consistent here, a comment is always followed by a property. I easily see generic `{@selfdoc}` and just skip, and property with some real comment is very noticeable, like in a first example.

Chi-teck:
> it's that Drupal needs to get with the times and change its documentation requirements.

Right, that's exactly what we are doing now. `@selfdoc` is for those who don't tolerate elements without docblocs. See point 2 in the issue summary.

It's up to developers to use this tag or simply omit the whole docblock.

ashnazg:
> @ashnazg don't you think that using `@inheritDoc` is semantically incorrect for elements that are not inherited?

Yes, I think it's incorrect to use for a _non-inherited_ element.  However, the spec's verbiage for usage of the non-inline `@inheritdoc` is **not** for non-inherited elements... the spec is explicit about saying it's only for inherited elements:

```
Because inheritance is implicit it may happen that 
it is not necessary to include a PHPDoc with a "Structural Element". 
This can cause confusion as it is now ambiguous whether the PHPDoc was omitted on purpose 
or whether the author of the code had forgotten to add documentation.

In order to resolve this ambiguity the @inheritDoc tag can be used to indicate 
that this element will inherit its information from a super-element.
```

If you are really really stuck in a documentation regime that is absolutely positively requiring a docblock on an element, it makes far more sense to me to _just go ahead and write some kind of Short Description for the element_.  I see no difference between this `@selfdoc` tag idea and choosing to add an `assertTrue()` to a unit test _just to make it pass_, and thus completely ignoring _the point_ of the unit test.

PR #1313, opened 2024-01-04
Update example code to latest version

kayw-geek:
This PR updated the example code as the `psr/simple-cache` package upgraded

### DIFF
--- a/accepted/PSR-16-simple-cache.md
+++ b/accepted/PSR-16-simple-cache.md
@@ -132,6 +132,7 @@ and is equivalent to a "Pool" in PSR-6. Different CacheInterface instances MAY b
 datastore, but MUST be logically independent.
 
 ```php
+
 <?php
 
 namespace Psr\SimpleCache;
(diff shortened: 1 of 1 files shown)


samdark (approved):

garak (approved):

Crell (changes requested):
We have not updated the main body of the spec for other specs that had type updates.  We've only updated the package, and left an Errata in the meta doc.  We should do the same here.

kayw-geek:
> We have not updated the main body of the spec for other specs that had type updates. We've only updated the package, and left an Errata in the meta doc. We should do the same here.

I see. So your suggestion is that the sample code doesn't need to be changed at all?

Crell:
It's not sample code.  It's the spec.  The spec doesn't change.  The interface packages are tweaked to move the already-defined type rules into the source.  That just needs a mention in the metadoc, which... I think already happened for PSR-16, didn't it?

PR #1330, opened 2025-06-11
New PSR for standartizing CAPTCHA (CaptchaInterface)

LeTraceurSnork:
Here's a proposal of new PSR for standartizing CAPTCHAs to one interface
Continuation of this thread: https://discord.com/channels/788815898948665366/788816084383694848/1379332512886689812

TTL:
Recently we've faced with a problem that government forbids to gather data abroad and we urgently need to switch Google ReCaptcha to other solutions, but the thing is that since those solutions doesn't have common interface yet, the codebase needs to be refactored, which isn't good - the task in a nutshell is just "switch vendors"


jaapio:
I do not really see why we need a psr for this. An interface can always be used by any developer in any application. 

The problem of recaptcha is a very specific implementation to validate humans interacting with your website. While there are other solutions.

Can you explain why you think this should be a standard?

LeTraceurSnork:
@jaapio 
> Can you explain why you think this should be a standard?
I'll try in examples

So, for example, I have a Google ReCaptcha on my website, lets say I installed it via composer from [corresponding repository](https://github.com/google/recaptcha). Let's omit frontend integration and look at the backend, we have smth like this:
```php
class FormSubmitService
{
    private $recaptcha;

    public function __construct(\ReCaptcha\ReCaptcha $recaptcha)
    {
        $this->recaptcha = $recaptcha;
    }

    public function submitForm(CustomRequest $request): CustomResponse
    {
        $token = $request->get('g-recaptcha-response');
        
        $response = $this->recaptcha->verify($token);
                                    
        if (!$response->isSuccess()) {
            return new CustomResponse('Recaptcha failed!');
        }
        
        // ... continue to submit fom
    }
}

//...

$recaptcha = new \ReCaptcha\ReCaptcha('my_secret_token');
$formSubmitService = new FormSubmitService($recaptcha);
$formSubmitService->submitForm($incomeCustomRequest);
```
So, what happens if we at some moment switch to another Captcha vendor (another SDK)? We immediately lose the `\ReCaptcha\ReCaptcha` class, `__construct()` fails with Type error, DI ruined. At this point we want to rely on some interface like this:
```php
public function __construct(CaptchaInterface $captcha)
{
    $this->captcha= $captcha;
}
```
because in this scenario we can pass:
```php
$recaptcha = new \ReCaptcha\ReCaptcha('my_secret_token');
$hcaptcha = new \HCaptcha\HCaptcha('my_secret_token');
$mcaptcha= new \MCaptcha\MCaptcha('my_secret_token');
$cloudflareTurnstile= new \Cloudflare\Turnstile('my_secret_token');
$smartCaptcha= new \Yandex\SmartCaptcha('my_secret_token');
$formSubmitService = new FormSubmitService($recaptcha);
$formSubmitService2 = new FormSubmitService($hcaptcha);
$formSubmitService3 = new FormSubmitService($mcaptcha);
$formSubmitService4 = new FormSubmitService($cloudflareTurnstile);
$formSubmitService5 = new FormSubmitService($smartCaptcha);
```
Yes, we will need an additional configuration when constructing the Captcha object itself, but inside the form handler there will be no changes, 'cause all of the above will be able to say true/false `$response->isSuccess()`, so the switching between providers will be much easier

KorvinSzanto:
Whoops, fat fingered the close button.

LeTraceurSnork:
@KorvinSzanto since all of the most popular Captchas (and, I guess, all of the external Captchas) requires secret token to connect to their validation routes, I guess it would benefit to add `InvalidArgumentException` to `CaptchaInterface::verify()` to handle `incorrect secret token` cases.
Not quite sure about `ValidationRequestErrorException` tho, but I guess smth like that can be added to `CaptchaInterface::verify()` as well, but I'm not quite sure to what scenario. Captcha service did not respond? Respond with not 2** (404, 500, etc.)? Respond with 200 but "Captcha not passed"? 🤔

LeTraceurSnork:
> Have you already gathered some consensus between the developers of the major captcha libraries in PHP?
Unfortunately, I did not, 'cause neither Google, hCaptcha, Yandex, nor their core developers in person did not respond to an invitation 😕
I will continue my attempts tho, now it is more convenient with a PR opened

captcha.md
```diff
@@ -0,0 +1,121 @@
...
+```php
+namespace Psr\Captcha;
+
+/**
+ * Interface of the object that CaptchaInterface obliged to return on ::verify() method.
+ * MUST contain enough information to consistently say whether user succesfully passed Captcha or not.
+ * SHOULD contain actual user's SCORING
+ * MAY contain additional information (e.g., gathered from it's captcha-vendor service's verification endpoint) (i.e. message, errors, etc.)
+ */
+interface CaptchaResponseInterface
```
samdark: Why is it an interface and not a bool?

captcha.md
```diff
@@ -0,0 +1,121 @@
...
+```php
+namespace Psr\Captcha;
+
+/**
+ * Interface of Captcha service itself.
+ * MUST decide whether user passed the Captcha or not and return corresponding response.
+ * SHOULD contain method to configure SCORING threshold (if applicable by PROVIDER)
+ * SHOULD throw a CaptchaException as soon as possible if appears any non-user related error that prevents correct Captcha solving (e.g. network problems, incorrect secret token, e.g.)
+ */
+interface CaptchaInterface
```
samdark: I think there might be a need for `challenge(string $token): string` method which, given a token, generates the challenge for the user to solve.

captcha.md
LeTraceurSnork: You mean, `why CaptchaInterface::verify()` returns that instead of a bool? If yes, then answer is - 'cause you might want to implement additional methods to specific `CaptchaResponse` classes, like `getScore()`, `getHost()` (for instance, those fields implemented by Google ReCaptcha, hCaptcha, SmartCaptcha) or similar - to extend response data

captcha.md
LeTraceurSnork: Idk what did you mean by this, probably `CaptchaInterface` needs to be renamed to `CaptchaVerifierInterface`, 'cause that is its purpose - to verify that passed token is valid. But idk what else can be retrieved from $token, especially in `string`. Do you mean `challenge` is frontend rendered `I am not a robot` checkbox with traffic lights? If so, i'm not sure that service should interact with it - it's just a verifier

captcha.md
samdark: Should be mentioned in meta-document.

captcha.md
samdark: Well, for example, given a token that is `24`, challenge might be `20+4` or something like that.

LeTraceurSnork: Well, that interface is not quite for that purpose, but rather to verify already solved captcha task by its token using some external (or even internal, it doesnt actually matter) captcha service. Roughly, it's an interface to build SDK on

samdark: I see. Makes sense.

JoshuaBehrens:
Here is Shopware 6 implementation https://github.com/shopware/shopware/tree/e4c3f565857f8e6c53d3c8e35f6b00dfd0eb3f73/src/Storefront/Framework/Captcha and Shopware 5 implementation https://github.com/shopware5/shopware/tree/8779bb0fc2cff04bb92dbba8ea5522263a71ab48/engine/Shopware/Components/Captcha

LeTraceurSnork:
> Here is Shopware 6 implementation https://github.com/shopware/shopware/tree/e4c3f565857f8e6c53d3c8e35f6b00dfd0eb3f73/src/Storefront/Framework/Captcha and Shopware 5 implementation https://github.com/shopware5/shopware/tree/8779bb0fc2cff04bb92dbba8ea5522263a71ab48/engine/Shopware/Components/Captcha

Do you believe that Shopware could use and get profit from implementation of this interfaces by achieving more information from responses?

captcha.md
```diff
@@ -0,0 +1,106 @@
...
+   - Provider API shutdowns
+   - Compliance requirement changes
+4. **Cost Containment**: Eliminate:
+   - System-wide code refactoring during migrations
+   - Parallel implementation maintenance
+   - Provider-specific testing overhead
+
+## 2. Interfaces
+
+### 2.1 CaptchaInterface
```
phpfui: This line should now read:

```
### 2.1 CaptchaVerifierInterface
```

Due to the code change.

captcha.md
LeTraceurSnork: mentioned it in `captcha-meta.md` (added meta-document)

captcha.md
LeTraceurSnork: done

LeTraceurSnork:
The administrators of the PHP-FIG community have created a Discord channel for the working group communications of this proposed PSR. I invite everyone interested in participating in the development/support (nominally, we need a working group) to join via the link: https://discord.com/channels/788815898948665366/1386802002599739443

VincentLanglet:
I feel like Captcha feature is something too specific to be worth a whole personal PSR.
And looking at the interface makes me believe this more.

You're looking for a PSR because all the catcha implementation have important differences ; but because of these differences, we end up with useless generic interfaces.

When looking at:
```
public function verify(string $token): CaptchaResponseInterface;
```
I feel like this is a really strict signature: you might need more informations to verify something

[Google accepts the IP](https://github.com/google/recaptcha/blob/master/src/ReCaptcha/ReCaptcha.php#L161)
If I rely on your interface but call `$catch->verify($token, $ip)` the benefit is null since I cannot be sure that when I'll change the lib I'll use, the signature will be the same. Using a signature `verifiy(string $token, array $context)` could help but same, it cannot guarantee the array keys will be the same. 

[Shopware lib use the request](https://github.com/shopware/shopware/blob/e4c3f565857f8e6c53d3c8e35f6b00dfd0eb3f73/src/Storefront/Framework/Captcha/GoogleReCaptchaV3.php#L28)
This does not seems compatible with your interface, and asking the user to parse the request in order to get the token doesn't seems right to me since you're coupling the implementations.

`CaptchaResponseInterface` is not better, even if the phpdoc is saying the implementation might have extra methods, if I want to be able to switch implementation easily in my project, I have to use only the one provided by the interface. So no score, no error message, and so on...

By trying to be too much generic, those PSR doesn't solve his goal IMHO

LeTraceurSnork:
> CaptchaResponseInterface is not better, even if the phpdoc is saying the implementation might have extra methods, if I want to be able to switch implementation easily in my project, I have to use only the one provided by the interface. So no score, no error message, and so on

@VincentLanglet on the second part of your message
Thank you for raising an important issue with `CaptchaResponseInterface` being too minimalistic.

While I believe that having just `isSuccess()` covers the most basic use case — telling whether the captcha has been passed or not, and keeps the code simple and swappable.
However, as you pointed out, some providers such as ReCaptcha v3 return extra details like 'score', error messages, etc. With the current contract, this data becomes inaccessible if using DI-bound code, unless adding hacks like downcasting or specific type checks, which are not ideal.
A few ways to extend:
- Add a generic `getData(): array` to expose arbitrary provider details; drawbacks: brittle, no typing. But that's just not good
- Let `verify()` return a more specific response interface for each implementation, but then consumer code is tied to implementation, so we made a circle
- Include optional getters in the base interface (`getScore()`, `getErrorCodes()`), but for many implementations these would need to return null or throw.

For now, the base interface solves the simple validation problem well, but for advanced response handling, it's worth discussing which extensibility pattern (if any) would be preferable — whether as a universal `getData()`, or a more formal structure.

Thanks for surfacing this; now we're talking and I believe this is what WG should discuss and eventually solve

phpfui:
I am going to go with what @LeTraceurSnork said. The whole point is to provide a minimal interface that all Capthca systems have to provide. It is minimal, but with some proper code around it, you can easily and quickly swap in another provider. With PHP, that could be configured with a ini file or database entry. The code will work the same. You can then fine tune the new provider if needed.

And as for minimal interfaces, I find the current Container interface also to minimal to be usefull, but that was agreed upon and in use.  So I don't think this interface should be rejected simply because it does not account for all use cases. It accounts for enough to be useful and will help standardize Captcha libraries to make them more generic and easy to swap out if needed.

Crell:
The de facto pattern of several past PSRs has been that the interfaces provide the "read" but not "configure" API.  Configure is left up to the particular implementation, and you configure it in advance.  The API is just for plugging it in and using it, assuming most of the configuration is already handled before the service gets injected.

(PSR-3 Logging, PSR-11 Containers, PSR-14 Events, PSR-20 Clocks, etc. all take that approach.)

phpfui:
The future of software is interoperable standards that allow things to be mixed and matched and not have vendor lock in. Vendors want lock in, standards prevent that. So yes, we need more standard interfaces, mail and translation would be two good ones to tackle.

We have standards in most industries. Metric bolts for example. JIS (Japanese Industrial Standard) comes to mind. The more we insist on standard interfaces the better off as a PHP community we are in the long run.

LeTraceurSnork:
> The last question I have in mind is "Which feature should have a PSR" and which shouldn't.

@VincentLanglet as I confessed to @samdark recently, I'm an inclusionist, as wikipedians would say 🙂 In my believe, all of your mentioned interfaces may actually be accepted if written properly, reviewed, agreed upon, etc. So, to `Where will be the limit ?` I would say 'the answer is none'. More PSRs to PSR god 🤩
p.s. isn't `TranslatorInterface` suggested in PSR-21? 🤔

> It's important to keep in mind that not all captchas are built the same. They won't always have a single value to validate, though I concede most do today

@KorvinSzanto yes, absolute most of the Captchas operates token as a string. I did some researches, I've got [two](https://www.aguko.com/cat/captchas) different [sources](https://www.wmtips.com/technologies/captchas/) that representeed different information about CAPTCHA providers usage percentage, so I've checked following: Google ReCaptcha, Cloudflare Turnstile, hCaptcha, Friendly Captcha, AWS WAF, Yandex SmartCaptcha, GeeTest, Tencent Captcha, Yidun Captcha, MTCaptcha, Altcha, VAPTCHA and KeyCAPTCHA. From all of those (that's more than 99,9% of market) only Altcha verifies something that is not a string - their SDK method accepts array of some data, but even that case is totally fits in presented interface - we with @VincentLanglet discussed it earlier - it can be achieved via mutable objects with `->setters()`. Now, after almost all of the known market being analyzed, I guess, even if there is some captcha provider that verifies, say, two tokens simultaneously (and both are required, so we can't go with `->setters()`) we can just agree that this one is just too exotic for an interface

> To me, captcha input is closer to a cache item than a simple log message. Like the event PSR, we should at the very least accept an object instead of a string. As it sits now, one must know the implementation to successfully use a provided CaptchaInterface

Oh, I see your point. Yes, while we could not have an opportunity to pump the verifier with additional data it needs beforehand (e.g. Verifier needs some data from HttpRequest that being validated right now and the one that actually contains the token), I guess, it is wise to give it an opportunity to do so. But what do you think of an idea to keep the `verify($token)` as it is, but to add `withData(object $data)` or `withData(array $data)` method (or smth like that)?

> In my opinion it'd be better to accept a full PSR-7 request in validate

With this being said, I guess, the question "what do we do with Response then" is much more acute for now 

@phpfui totally agree with you! More standarts -> less exotic packages that doesn't fits them -> less odd unsupportable legacy

LeTraceurSnork:
> > I invite everyone interested in participating in the development/support (nominally, we need a working group) to join via the link: https://discord.com/channels/788815898948665366/1386802002599739443
> 
> How to join?
> This is not an invite link, but just a channel link
> 

Pardon, link changed to https://discord.gg/uAfJDfY2

andrew-demb:
@LeTraceurSnork 

1. Let's not consider any mutating methods aka "setRequest"/"withRequest" to verifiers.

All data should be directly passed to the method

----

2. I don't think it will be necessary to parse request content. Most cases is about:
- IP (server request attributes, server request server params, reading headers)
- host (`RequestInterface::getUri()` - `UriInterface::getHost()`)

```php
<?php

// verifier that requires token, IP, host and some other data (optional)
class AcmeCaptchaVerifier implements CaptchaVerifierInterface
{
    public function __construct(private readonly string $requestIpAttributeName)
    {
    }

    public function validate(string $token, ContextInterface $context)
    {
        $someValue = null;
        if ($context instanceof AcmeContextInterface) {
            $ip = $context->getIpForAcmeProvider();
            $host = $context->getHostForAcmeProvider();
            $someValue = $context->getSomeValueForAcmeProvider();
        } elseif ($context instanceof RequestAwareContextInterface && $context->getRequest() instanceof \Psr\Http\Message\ServerRequestInterface) {
            /** @var \Psr\Http\Message\ServerRequestInterface $request */
            $request = $context->getRequest();
            $ip = $request->getAttribute($this->requestIpAttributeName, $request->getServerParams()['REMOTE_ADDR'] ?? null);
            $host = $request->getUri()->getHost();
            $someValue = null; // unable to determine
        } else {
            // OR fallback to some default values for $ip, $host, $someValue
            throw new \InvalidArgumentException('Unsupported context - should be AcmeContextInterface or RequestAwareContextInterface to determine IP and host');
        }

        $http_client = new HttpClient(); // in this example PSR-18 compatible

        // <=== START
        $http_request = new HttpRequest(
            [ // PSR-7 compatible
                'token' => $token,
                'ip' => $ip,
                'host' => $host,
                'some_value' => $someValue, // some other data
            ],
        );
        // <=== END
        $http_client->sendRequest();
    }
}

// verifier that uses ONLY token
class FooCaptchaVerifier implements CaptchaVerifierInterface
{
    public function __construct(private readonly string $requestIpAttributeName)
    {
    }

    public function validate(string $token, ContextInterface $context)
    {
        $http_client = new HttpClient(); // in this example PSR-18 compatible

        // <=== START
        $http_request = new HttpRequest(
            [ // PSR-7 compatible
                'token' => $token,
            ],
        );
        // <=== END
        $http_client->sendRequest();
    }
}

(new CaptchaVerifier())->verify($token, new AcmeContext('127.0.0.1', 'example.com', 'some_value'));
(new CaptchaVerifier())->verify($token, new Psr7Context(Psr7Request::createFromGlobals()));
```

andrew-demb:
... and why don't "withData" or "array $context" - it won't be typed. Very hard to decide unified keys what can be used by different implementations.

My proposal based on "empty interface" at a core and (at least) one additional extending interface that knows about PSR-7 request

LeTraceurSnork:
I think, I got your point
What if we add the second parameter to CaptchaVerifierInterface`::verify()`, but it will be optional?
Smth like this:
```php
public function verify(string $token, CaptchaContextInterface = new NoContext()): CaptchaResponseInterface;
```
And then add smth like this:
```php
interface CaptchaContextInterface
{
// need to think about its content and methods
}

class NoContext implements CaptchaContextInterface
{
// all of its methods will be stubs that do nothing/returns empty data
}
```

Then we can pass some context to verify() if we need it (creating it beforehand), but if we don't - then just call `->verify($token)` and context will be empty

LeTraceurSnork:
> I don't think it will be necessary to parse request content. Most cases is about:
IP (server request attributes, server request server params, reading headers)
host (RequestInterface::getUri() - UriInterface::getHost())

@andrew-demb My point was that if you rely on anything else rather than a token, you need to either pass it directly to `::verify()` method, but that could be a problem as discussed earlier (you don't know the implementation and what fields does it require) OR you need to pass them to `CaptchaVerifier` class itself during the configuration step - and only after it the verificator can be passed as a param to, say, `AbstractFormSubmitHandler` constructor. And after it you are no more able to set additional params, because they are behing the interface. So, with that said, I only see two options:
1. Methods `->withData()` anyway (or similar)
2. Verifier factory

@KorvinSzanto

VincentLanglet:
> The core purpose of the interface is to tell these solutions that, from the backend’s standpoint, it doesn’t matter how they operate on the frontend—as long as they pass a string-based token in the request.

Personally I'm not convinced about such incomplete interface since
- Not all captcha are string-token based
- Not all catcha are token based only

Asking to call some extra method `withIp()`, `withScore()` and so on which are not in the interface lose all the benefit.

What the purpose of writing
```
class FormService
{
      public function __construct(private readonly CaptchaVerifier $captcha) {}
      
      public function submitForm(CustomRequest $request): CustomResponse
    {
        $token = $request->get('g-recaptcha-response');
        
        $this->recaptacha->withIp($request->getIp()); // THIS BREAKS THE BENEFIT OF THE INTERFACE
        $response = $this->recaptcha->verify($token);
                                    
        if (!$response->isSuccess()) {
            return new CustomResponse('Recaptcha failed!');
        }
    }
```
and we could even argue it's odd to rely on the key `'g-recaptcha-response'` which might change from a CaptchaVerifier to another, passing the request directly feels better.

Anyway...

LeTraceurSnork:
Gentlemen, all who is interested in this PSR, or at least in discussing it - I believe, we need to at least duplicate this converstaion (maybe partially) to this official discussion in Google Groups: https://groups.google.com/g/php-fig/c/v8VymG3L9no

phphleb:
This is a pretty cool initiative and seems like it could be super handy. If you think about the main point of CAPTCHA, it’s to tell humans apart from bots (like a Voight-Kampff test but on the web). The ways people find to bypass CAPTCHAs are always evolving, and because of that, CAPTCHA implementations have to keep getting smarter too. If we limit CAPTCHA’s flexibility through a PSR, we might end up forcing developers to ditch unique, creative protection methods that don’t fit the PSR mold. 

Now, tying this to PSR-7 (handling all the request data), as suggested earlier, means that tons of sites that don’t use PSR-7 are going to have a tough time just implementing a regular CAPTCHA. Currently, without a set standard, it’s easier, and users are accustomed to how simple it is. Take WordPress for example; it doesn’t have PSR-7, and neither do lots of other sites; changing a CAPTCHA there is pretty straightforward.

From a corporate perspective, where tasks are split into backend and frontend, a standard might indeed speed things up on the backend, but it won’t really help the frontend guys, which means it won’t speed up CAPTCHA implementation or replacement overall. It might make life easier for backend developers, which is what we're aiming for, right? 

Despite this criticism, I think the idea has potential if it’s developed further. Though, I have to admit, I’m not an expert in CAPTCHA technology.

phpfui:
You can also see my implementation for Google's ReCaptcha here: https://github.com/phpfui/recaptcha/tree/PSR

LeTraceurSnork:
@andrew-demb @phphleb may I officially invite you to a Working Group? I really appreciate your effort, tho, and PHP-FIG rules stricly and explicitly requires 4+ people to form a Working Group for proposal at least become a PSR, so, for now it's only me and @samdark, now I need your participation as well :-)

LeTraceurSnork:
@phpfui @phphleb @andrew-demb @JoshuaBehrens still inspiring you to join (and preventing PR from stale) :-)

PR #1332, opened 2025-06-20
Update and rename PSR-7-https-message.md to PSR-7-https-message.md

stick720:
Yes updated


Jean85:
Closing this as it seems a massive edit with low effort; I could accept moving all links to HTTPS, but there are a lot more changes here, and some of them do not even make sense.

Crell:
Nothing in PSR-7 mandates HTTPS anyway, so this would be an incorrect change.

PR #1333, opened 2025-06-27
Update personnel.md to reflect May 2025 election results

Jean85:


mbniebergall (approved):
Updates look correct from recent voting, thanks.

PR #1337, opened 2025-07-16
New PSR for standardizing Validator (ValidatorInterface)

LeTraceurSnork:
Here's a proposal of new PSR for standardizing Validators to one interface
Continuation of long-lasting discussions on official Discord channel

## Goal

To unite ~~earth and water~~ many separated validator libraries under one interface so the switching and DI would be easier, protect from potential vendor-locks and reduce coupling.

### DIFF
--- a/proposed/validator-meta.md
+++ b/proposed/validator-meta.md
@@ -1,0 +1,85 @@
+Validator Meta Document
+=======================
+
+## 1. Summary
+
+The SimpleValidatorInterface and ExtendedValidatorInterface define universal, minimalistic contracts for value validation in PHP. They provide a unified way to perform type-agnostic validations with either a simple pass/fail response (suitable for most form or field inputs) or a detailed, structured validation report, supporting error codes for i18n and advanced cases. Inspired by Symfony's Validator and `Respect\Validation`, these interfaces allow maximum interoperability and rapid provider/implementation swap, while enabling framework- and library-level integration without tight coupling.
+
+Applications MAY depend only on these interfaces and swap implementations (or chains/compositions of validators) via DI, reducing coupling and simplifying maintenance, testing, and future migrations.
+
+## 2. Why Bother?
+
+Form and value validation is a ubiquitous problem across all PHP applications. Today, each framework (Symfony, Laravel, Yii, etc.) exposes its own interfaces, making it difficult to share validators, migrate between frameworks or libraries, or compose validation pipelines with userland tools.
+
+A universal validator contract allows:
+
+Rapid replacement of validation engines or libraries (vendor-agnostic interface).
+Interchangeable custom and vendor-specific validators (e.g., open-source and proprietary).
+Testability and integration with dependency injection containers.
+Consistent error format for user feedback, translation, or error mapping.
+Clean separation of simple (bool) and extended (structured, multi-error) validation.
+This pattern avoids having to deeply couple code to a specific validator ecosystem (as in Symfony) and dramatically reduces refactoring when switching stacks or updating validation strategies.
+
+#### Pros:
+
+- Plug-and-play with any validator implementations (no hard dependency on a specific package)
+- Extensible (vendors can add extra methods, but must comply with the common interface)
+- Clean and decoupled design 
+- Advanced: error codes, rich violation objects
+
+#### Cons:
+
+- Slight abstraction overhead (vs. using one concrete implementation)
+- Vendor- or application-specific error codes are not enforced by spec (but spec supports them)
+- Contextual or cross-field validation must be built upon (not in base interfaces)
+
+## 3. Design Decisions
+
+### 3.1. Simple vs. Extended
+
+To satisfy both lightweight and complex needs, two interfaces are provided:
+
+- `SimpleValidatorInterface` — exposes only `isValid(mixed $value): bool`. Suitable for basic needs (is it a valid phone? email? int in range?...) and maximizing performance. 
+- `ExtendedValidatorInterface` — exposes `validate(mixed $value): ValidatorResponseInterface`, for structured results (validation status, errors, error codes/messages/etc.).
+This separation enables applications to pick the level of detail they need. Many use cases will use only the simple interface.
+
+### 3.2. ValidatorResponseInterface
+
+A response from ExtendedValidatorInterface must expose:
+
+- `isValid(): bool` — Was validation successful? 
+- `getViolations(): array<ViolationInterface>` — Why did it fail, with detailed error objects.
+
+### 3.3. ViolationInterface
+
+Every violation contains:
+
+- Machine-readable code (for programmatic matching, i18n, etc.). 
+- Human-readable message (for fallback user feedback). 
+- Optionally: parameters for message templates, offending value, etc. 
+  Spec only mandates code and message; implementations may extend.
+
+### 3.4. Exception Handling
+
+Validation failures (value rejected) are NOT exceptional: response returned with errors.
+
+Exceptions are for implementation/configuration/usage errors — not for negative validation.
+
+### 3.5. Swappability & DI
+
+Core purpose — to allow hot-swapping validator engines and compositions via DI or configuration, zero code refactor.
+
+### 3.6. Scope
+
+Only single-value validation in scope; not object/collection/nested.
+Contextual validation may be layered by implementors using extensions.
+
+## 4. People
+
+### 4.1 Editor(s)
+
+- TBA
+
+### 4.2 Working Group members
+
+- TBA
--- a/proposed/validator-meta.md
+++ b/proposed/validator-meta.md
@@ -67,7 +67,9 @@ Exceptions are for implementation/configuration/usage errors — not for negativ
 
 ### 3.5. Swappability & DI
 
-Core purpose — to allow hot-swapping validator engines and compositions via DI or configuration, zero code refactor.
+Core purpose — to make validation logic universally pluggable and reusable, in both applications and stand-alone
+libraries, with no forced dependency on any particular framework or DI mechanism, and with zero need for code changes
+when switching validator implementations.
 
 ### 3.6. Scope
 
--- a/proposed/validator-meta.md
+++ b/proposed/validator-meta.md
@@ -40,7 +40,7 @@ This pattern avoids having to deeply couple code to a specific validator ecosyst
 To satisfy both lightweight and complex needs, two interfaces are provided:
 
 - `SimpleValidatorInterface` — exposes only `isValid(mixed $value): bool`. Suitable for basic needs (is it a valid phone? email? int in range?...) and maximizing performance. 
-- `ExtendedValidatorInterface` — exposes `validate(mixed $value): ValidatorResponseInterface`, for structured results (validation status, errors, error codes/messages/etc.).
+- `ExtendedValidatorInterface` — exposes `validate(mixed $value, array $context = []): ValidatorResponseInterface`, for structured results (validation status, errors, error codes/messages/etc.).
 This separation enables applications to pick the level of detail they need. Many use cases will use only the simple interface.
 
 ### 3.2. ValidatorResponseInterface
--- a/proposed/validator.md
+++ b/proposed/validator.md
@@ -63,12 +63,13 @@ interface ExtendedValidatorInterface
      * SHOULD throw ValidatorException if the Validator itself could not tell if the value passes or not (e.g. Validator misconfigured)
      *
      * @param mixed $value
+     * @param array $context
      *
      * @throws ValidatorException
      *
      * @return ValidatorResponseInterface
      */
-    public function validate(mixed $value): ValidatorResponseInterface;
+    public function validate(mixed $value, array $context = []): ValidatorResponseInterface;
 }
 ```
 
--- a/proposed/validator.md
+++ b/proposed/validator.md
@@ -1,0 +1,144 @@
+Validator PSR: Specification Draft
+==================================
+
+This document describes common interfaces to validate values in PHP, in both simple and structured modes.
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in [RFC 2119][].
+
+The final implementations MAY decorate the objects with more
+functionality than the one proposed but they MUST implement the indicated
+interfaces/functionality first.
+
+[RFC 2119]: http://tools.ietf.org/html/rfc2119
+
+## 1. Specification
+
+### 1.1. Definitions
+
+- Validation: Process of checking if input value satisfies certain constraints (type, format, range, etc.). 
+- Violation: Individual reason why a value fails validation (error). 
+- Error Code: Machine-readable string/code for a validation failure, suitable for mapping, i18n or programmatic branching.
+
+## 2. Interfaces
+
+### 2.1. SimpleValidatorInterface
+
+```php
+namespace Psr\Validator;
+
+/**
+ * Minimalistic validator interface.
+ */
+interface SimpleValidatorInterface
+{
+    /**
+     * Validates supplied value.
+     * MUST return true if $value passes.
+     * MUST return false if $value fails.
+     * MUST NOT throw exceptions if $value fails.
+     * SHOULD throw ValidatorException if the Validator itself could not tell if the value passes or not (e.g. Validator misconfigured)
+     *
+     * @param mixed $value
+     *                    
+     * @throws ValidatorException
+     *
+     * @return bool
+     */
+    public function isValid(mixed $value): bool;
+}
+```
+
+## 2.2. ExtendedValidatorInterface
+
+```php
+namespace Psr\Validator;
+
+/**
+ * Extended validator interface returning structured response.
+ */
+interface ExtendedValidatorInterface
+{
+    /**
+     * Validates supplied value and returns a ValidatorResponseInterface instance.
+     * MUST return ValidatorResponseInterface if $value was validated at all (with any result).
+     * MUST NOT throw exceptions if $value fails.
+     * SHOULD throw ValidatorException if the Validator itself could not tell if the value passes or not (e.g. Validator misconfigured)
+     *
+     * @param mixed $value
+     *
+     * @throws ValidatorException
+     *
+     * @return ValidatorResponseInterface
+     */
+    public function validate(mixed $value): ValidatorResponseInterface;
+}
+```
+
+## 2.3. ValidatorResponseInterface
+
+```php
+namespace Psr\Validator;
+
+/**
+ * Wraps the result of validation including status and violations list.
+ */
+interface ValidatorResponseInterface
+{
+    /**
+     * MUST return true if $value passed, and false otherwise.
+     * MUST be immutable.
+     *
+     * @return bool
+     */
+    public function isValid(): bool;
+
+    /**
+     * Returns all violations (validation errors) as ViolationInterface objects.
+     *
+     * @return ViolationInterface[]
+     */
+    public function getViolations(): array;
+}
+```
+
+## 2.4. ViolationInterface
+
+```php
+namespace Psr\Validator;
+
+/**
+ * A single validation violation.
+ */
+interface ViolationInterface
+{
+    /**
+     * Machine-readable error code (for mapping, i18n, client logic).
+     *
+     * @return string
+     */
+    public function getCode(): string;
+
+    /**
+     * Error message (human-readable, MAY be locale-dependent).
+     *
+     * @return string
+     */
+    public function getMessage(): string;
+}
+```
+
+## 2.5. ValidatorException
+
+```php
+namespace Psr\Validator;
+
+/**
+ * Thrown ONLY if validator is misconfigured or cannot process request.
+ * MUST NOT be thrown if $value was actually tested, no matter the result.
+ */
+class ValidatorException extends \InvalidArgumentException
+{
+}
+```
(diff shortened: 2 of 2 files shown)


proposed/validator-meta.md
```diff
@@ -0,0 +1,85 @@
...
+Form and value validation is a ubiquitous problem across all PHP applications. Today, each framework (Symfony, Laravel, Yii, etc.) exposes its own interfaces, making it difficult to share validators, migrate between frameworks or libraries, or compose validation pipelines with userland tools.
+
+A universal validator contract allows:
+
+Rapid replacement of validation engines or libraries (vendor-agnostic interface).
+Interchangeable custom and vendor-specific validators (e.g., open-source and proprietary).
+Testability and integration with dependency injection containers.
+Consistent error format for user feedback, translation, or error mapping.
+Clean separation of simple (bool) and extended (structured, multi-error) validation.
+This pattern avoids having to deeply couple code to a specific validator ecosystem (as in Symfony) and dramatically reduces refactoring when switching stacks or updating validation strategies.
```
Crell, line 21: It's not just about specific applications switching stacks.  It's about me being able to use a validation routine in a stand-alone library and not care if it's going to be used with Symfony, Slim, or Laravel.  Eg, in Serde I can just tell people "if you want fancier validation, go use PSR-Whatever and leave me alone."  (Or integrate a PSR-Whatever bridge into Serde if I feel like it.)

proposed/validator-meta.md
```diff
@@ -0,0 +1,85 @@
...
+A response from ExtendedValidatorInterface must expose:
+
+- `isValid(): bool` — Was validation successful? 
+- `getViolations(): array<ViolationInterface>` — Why did it fail, with detailed error objects.
+
+### 3.3. ViolationInterface
+
+Every violation contains:
+
+- Machine-readable code (for programmatic matching, i18n, etc.).
```
Crell, line 57: Disagree here.  The class name is the machine-readable code.  We don't need more opaque inconsistent strings.

proposed/validator-meta.md
```diff
@@ -0,0 +1,85 @@
...
+
+### 3.4. Exception Handling
+
+Validation failures (value rejected) are NOT exceptional: response returned with errors.
+
+Exceptions are for implementation/configuration/usage errors — not for negative validation.
+
+### 3.5. Swappability & DI
+
+Core purpose — to allow hot-swapping validator engines and compositions via DI or configuration, zero code refactor.
```
Crell: See above.

proposed/validator-meta.md
```diff
@@ -0,0 +1,85 @@
...
+
+### 3.6. Scope
+
+Only single-value validation in scope; not object/collection/nested.
+Contextual validation may be layered by implementors using extensions.
```
Crell, line 77: We should think through what compound validation would look like.  That should be included.

proposed/validator.md
```diff
@@ -0,0 +1,144 @@
...
+
+This document describes common interfaces to validate values in PHP, in both simple and structured modes.
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in [RFC 2119][].
+
+The final implementations MAY decorate the objects with more
+functionality than the one proposed but they MUST implement the indicated
+interfaces/functionality first.
```
Crell: This block is unnecessary.

proposed/validator.md
```diff
@@ -0,0 +1,144 @@
...
+     * MUST NOT throw exceptions if $value fails.
+     * SHOULD throw ValidatorException if the Validator itself could not tell if the value passes or not (e.g. Validator misconfigured)
+     *
+     * @param mixed $value
+     *
+     * @throws ValidatorException
+     *
+     * @return ValidatorResponseInterface
+     */
+    public function validate(mixed $value): ValidatorResponseInterface;
```
Crell: If the validation passed, we don't need detailed information.  We can likely short-circuit that to something like `true|ValidatorErrorInterface` or similar.

proposed/validator.md
```diff
@@ -0,0 +1,144 @@
...
+ */
+interface ValidatorResponseInterface
+{
+    /**
+     * MUST return true if $value passed, and false otherwise.
+     * MUST be immutable.
+     *
+     * @return bool
+     */
+    public function isValid(): bool;
```
Crell, line 92: This can be a property now, not a method. :smile:

proposed/validator.md
```diff
@@ -0,0 +1,144 @@
...
+## 2.5. ValidatorException
+
+```php
+namespace Psr\Validator;
+
+/**
+ * Thrown ONLY if validator is misconfigured or cannot process request.
+ * MUST NOT be thrown if $value was actually tested, no matter the result.
+ */
+class ValidatorException extends \InvalidArgumentException
```
Crell: I believe FIG convention is to us an interface here to mark exceptions from this package, not a base class.

proposed/validator.md
```diff
@@ -0,0 +1,144 @@
...
+     * @return string
+     */
+    public function getCode(): string;
+
+    /**
+     * Error message (human-readable, MAY be locale-dependent).
+     *
+     * @return string
+     */
+    public function getMessage(): string;
```
Crell: Could also be a property.

proposed/validator.md
```diff
@@ -0,0 +1,144 @@
...
+ * A single validation violation.
+ */
+interface ViolationInterface
+{
+    /**
+     * Machine-readable error code (for mapping, i18n, client logic).
+     *
+     * @return string
+     */
+    public function getCode(): string;
```
Crell, line 118: As specified, this is extraneous and unhelpful.

Crell:
As I have said many times, get the WG together first, design details second. :smile:

Crell:
Also: The current proposal doesn't seem to go far enough.  How do I link the validator to metadata?  Would, for instance, an attribute implement this interface?  If so, what does it do if it needs a service to do its validation?  (Eg, "does this record exist in the db?")  Then the validation metdata (attribute) and the validation logic (a service) need to be separate.

We don't have to mandate how that is handled, but we do need to ensure that it can be handled in a clean and portable fashion.

proposed/validator.md
Nyholm: There are zero benefits making it a property. Just drawbacks. 

Ie. If it is a property you are forcing the implementation to be a value object.

derrabus:
Hello and thank you for working on this proposal. I believe and common interface for validators can be useful, especially for interoperability with libraries and even external systems like JS frontends. A few remarks from my side. I'll might add some more later when I have more time.

#### Source of violation

When I validate a complex object, a violation with a human-readable message is often not enough. Imagine, the object represents a form submitted by the user. I want to validate the whole object and receive a list of violations. If the form is invalid, I want to render the message next to the field with the invalid input. In order to facilitate that, I need a machine readable representation of the property path. This is how Symfony does it.

https://github.com/symfony/symfony/blob/5015234dd9fe86dee561eb31a0de524f68a052a0/src/Symfony/Component/Validator/ConstraintViolationInterface.php#L86-L98

I'm not saying you should copy the way Symfony does it, but please take that use-case into consideration.

#### Translation

Regarding the error message, you're documenting it as…

> Error message (human-readable, MAY be locale-dependent).

This basically means that the validator has to be aware of the target locale when conducting the validation. This also means that I can only target a single locale which I have to determine before calling the validator. It would be great if the violation object would be translatable in some way. Unfortunately, we don't have a PSR standard for translators or translatable objects.

proposed/validator.md
derrabus: Those drawbacks should be gone with PHP 8.4 though. Not sure if we want to target PHP 8.4 for this standard just to gain access to hooked properties. That would as of today be rather strict and hurt the adoption of the standard.

Nyholm: But there are still no benefits... =)

proposed/validator-meta.md
```diff
@@ -0,0 +1,85 @@
...
+- Vendor- or application-specific error codes are not enforced by spec (but spec supports them)
+- Contextual or cross-field validation must be built upon (not in base interfaces)
+
+## 3. Design Decisions
+
+### 3.1. Simple vs. Extended
+
+To satisfy both lightweight and complex needs, two interfaces are provided:
+
+- `SimpleValidatorInterface` — exposes only `isValid(mixed $value): bool`. Suitable for basic needs (is it a valid phone? email? int in range?...) and maximizing performance.
```
samdark, line 42: Is that running a single validation rule against the value? If so, I'd name it differently... like `RuleInterface` or something.

proposed/validator-meta.md
```diff
@@ -0,0 +1,85 @@
...
+- Contextual or cross-field validation must be built upon (not in base interfaces)
+
+## 3. Design Decisions
+
+### 3.1. Simple vs. Extended
+
+To satisfy both lightweight and complex needs, two interfaces are provided:
+
+- `SimpleValidatorInterface` — exposes only `isValid(mixed $value): bool`. Suitable for basic needs (is it a valid phone? email? int in range?...) and maximizing performance. 
+- `ExtendedValidatorInterface` — exposes `validate(mixed $value): ValidatorResponseInterface`, for structured results (validation status, errors, error codes/messages/etc.).
```
samdark:
```suggestion
- `ValidatorInterface` — exposes `validate(mixed $value): ValidatorResponseInterface`, for structured results (validation status, errors, error codes/messages/etc.).
```

for the reason in https://github.com/php-fig/fig-standards/pull/1337/files#r2213153862

proposed/validator-meta.md
samdark: @Crell class name doesn't work well for exposing it in APIs i.e. JavaScript or Android/iOS clients should not be dealing with PHP fully qualified class names.

proposed/validator-meta.md
```diff
@@ -0,0 +1,85 @@
...
+
+Exceptions are for implementation/configuration/usage errors — not for negative validation.
+
+### 3.5. Swappability & DI
+
+Core purpose — to allow hot-swapping validator engines and compositions via DI or configuration, zero code refactor.
+
+### 3.6. Scope
+
+Only single-value validation in scope; not object/collection/nested.
```
samdark, line 76: I'd ensure that these fall into the given interfaces well.

proposed/validator.md
samdark: Usually there's a context that matters, such as in the case when you need a password to be repeated exactly.

proposed/validator-meta.md
samdark: Also, in reality, single value validation without context is not enough.

proposed/validator.md
```diff
@@ -0,0 +1,144 @@
...
+ * Minimalistic validator interface.
+ */
+interface SimpleValidatorInterface
+{
+    /**
+     * Validates supplied value.
+     * MUST return true if $value passes.
+     * MUST return false if $value fails.
+     * MUST NOT throw exceptions if $value fails.
+     * SHOULD throw ValidatorException if the Validator itself could not tell if the value passes or not (e.g. Validator misconfigured)
```
samdark, line 37: Should we use generic `\RuntimeException` for this case?

proposed/validator.md
samdark: @Crell that would create more code handling it. Having a single return value works better in this case.

proposed/validator.md
```diff
@@ -0,0 +1,144 @@
...
+
+## 2.4. ViolationInterface
+
+```php
+namespace Psr\Validator;
+
+/**
+ * A single validation violation.
+ */
+interface ViolationInterface
```
samdark, line 111: For intl, there is a need for:

```php
 /**
     * Returns parameters used for {@see $message} translation.
     *
     * @return array A mapping between parameter names and values.
     */
    public function getParameters(): array
```

samdark (commented):
Overall, I see it an interesting thing to standardize. Here's what we have in Yii3 about the topic:

https://github.com/yiisoft/validator

proposed/validator.md
Crell: 1. No, it does not force it to be a value object.  It just changes the spelling to be simpler and more concise.
2. The benefit is conceptual integrity.  See the talk I gave at PHPTek this year on the new world properties open up for us.
3. We should target 8.5, frankly.  This wouldn't be ready before 8.5 is released anyway, and we've learned our lesson in the past that targeting an older-than-latest release always backfires and just makes more work for all involved.

Crell:
@samdark Please don't use "Reviews".  They break threading badly. :smile: 

Having a single `ValidatorResponseInterface` on which you have to check `isValid` is also more code.  There will be code that determines which validators passed and which failed either way.  The rest is just spelling.  And for spelling, I (almost) always favor using the type system.

Re using the class name as the ID: The ID is an opaque string currently.  It wouldn't be useful to some other system anyway.  Honestly I'm not even sure what purpose it serves to exist at all, so unless that can be demonstrated I would object to having it.  (Not something that has to be resolved before a WG forms, of course.  The WG vote should likely include a lot *fewer* details than what is listed here.)

proposed/validator.md
derrabus: > We should target 8.5, frankly.

For Symfony this would mean that we cannot easily implement the standard before version 9 which will be released in November 2027. That's quite a stretch.

LeTraceurSnork: > We should target 8.5, frankly

Oh, no, please don't, not again 🌚 Not everybody use PHP 8.0+, let alone 8.5. Think about those poor millions that just cannot update their legacy from PHP 7.0+ (or even 5.6+) to newer versions - we can't leave them behind. We shall at least give them non-typed v1, and then, maybe, update to typed v2

Crell: Putting on my Core Committee hat for a moment, I will never vote for a spec that goes out of its way to support 10 year old versions of PHP.  A spec will always be used far more in the future than in the past.  Anyone on PHP 5 still (god help them) isn't going to be able to use any 3rd party validator libraries anyway, because those will all be written for more modern PHP versions.

We learned our lesson from PSR-6, and PSR-7 (if my memory of the timing is right).  And did it right in PSR-14 by using `object`.  Use whatever language features are current and useful.  If that makes the PHP version requirement high, so be it.  The future is longer than the past.

proposed/validator.md
LeTraceurSnork: Totally agree with @samdark here. I [faced this problem in WordPress once](https://github.com/WordPress/wordpress-develop/pull/4531) - a core function used to return `true|WP_Error` and if you checked that via `if ($response)` you always got true condition. So, eventually, they changed call from 
```php
if ( retrieve_password( $user->user_login ) )
```
to
```php
if ( true === retrieve_password( $user->user_login ) )
```

Regarding your thread - just... Why? Why do we need to complicate output instead of simplifying it?

p.s. I'm not even speaking about that we might want to read some details that Validator might provide on succesful validate

proposed/validator.md
derrabus: You are absolutely right @Crell about supporting ancient PHP releases. Most validation libraries that could adopt the new standard don't deliver new feature releases with PHP 7 support, so why bother about PHP 7.

But not supporting 10 year old PHP does not mean we have to exclude last year's PHP. As far as I can tell, requiring PHP 8.5 does not unlock any features that we could make use of in this contract, does it?

If hooked properties is what we want for this contract, let's target PHP 8.4? For Symfony at least, that would make adopting the standard way easier because we'll bump our PHP requirement to 8.4 in November.

proposed/validator-meta.md
LeTraceurSnork: Oh, yeah, I see your point. Tho, I'm struggling with exact words to put that point in them 😕
Feel free to rephrase that in a suggestion)))

proposed/validator-meta.md
LeTraceurSnork: Actually, I thought of some flag-like codes like those ones in `json_encode()` function (second argument)
But, yeah, class names are not good enough - for example, if you want to pass those violations to frontend

proposed/validator.md
Crell: If there's no 8.5 features that would be useful, then yes, a version target of 8.4 is fine.  But if we do find there are 8.5 features that would be useful, we should be very open to then requiring 8.5 to use them.

Off hand, I can't think of any new 8.5 features that would be useful here, but interface properties from 8.4, definitely are useful.

proposed/validator-meta.md
LeTraceurSnork: That will probably require some analogue of `Constraints` (like in Symfony), which may be just too much for this PSR. Not quite sure

proposed/validator.md
LeTraceurSnork: I would've understand if you'd say smth like `we can't support PHP 5.6 because we need to use object type as param` (which we don't...yet). But drop older versions support just because 8.5 is fancy? Why? It's not like we need to do something overbearing - just not use new syntax in an interface (that may be used, btw, in libraries that extends that interface). It's not like interface suggests usage of `money_format()` or smth 😁

proposed/validator.md
LeTraceurSnork: I suggest we decide that in one thread and then apply to all
https://github.com/php-fig/fig-standards/pull/1337#discussion_r2210895909

proposed/validator-meta.md
LeTraceurSnork: Well... That probably IS a `Constraint` (like Symfony's) 🤔
Shall it contain methods like `getCondition()` or smth like that then?

proposed/validator.md
LeTraceurSnork: No, because this one is more specific. RuntimeException is basically anything, while package exceptions is described specifically to show that error occures somewhere inside this package functionality

proposed/validator.md
LeTraceurSnork:
```suggestion
    public function validate(mixed $value, array $context = []): ValidatorResponseInterface;
```
Probably?

proposed/validator.md
LeTraceurSnork: fixed

proposed/validator.md
Nyholm: I think this long discussion proves my point. 

> There are zero benefits making it a property. Just drawbacks.

I understand we have property hooks and interface properties in PHP 8.4. But these two features did not bring any value to PHP other then "it is nicer to write". 

Since a PSR should be a standardized way to do something, I think we should consider adoption rather than try to move libraries to use language features that does not give any value. 

My strongest recommendation is to keep these as methods.

proposed/validator-meta.md
samdark: Maybe. That would be very useful if you want to replicate the same validation on the client side (mobile, SPA frontend) based on the rules used.

proposed/validator.md
samdark: Yes. Either that or a DTO.

proposed/validator.md
samdark: If you're not going to handle it specifically, it doesn't make sense to make it specific.

samdark:
@Crell That's what I've used to... and these give a very good context cause comments are attached to certain code.

proposed/validator-meta.md
LeTraceurSnork: clarified

proposed/validator.md
LeTraceurSnork: removed

proposed/validator.md
LeTraceurSnork: We should resolve this in https://github.com/php-fig/fig-standards/pull/1337#discussion_r2210882055 and then apply to all

proposed/validator.md
LeTraceurSnork: I guess, we could use PSR-21 benefits here. Not quite sure, what they looks like tho, probably better to ask PSR-21 WG. Ummm... @samdark 😁 ? @navarr ? What kind of interface is on its way to make i18n easier and how to use it, if it already exists/planned as draft?

proposed/validator.md
LeTraceurSnork: Personally, I have long suffered from the absence of a built-in exception in PHP that would semantically denote an "improperly configured entity/application. And I think this is exactly one of those cases.

proposed/validator.md
navarr: I don't think you'd need a `getParameters`, however it might be best for `getMessage` to return `string|Stringable`.  `Stringable` would allow it to be some sort of Phrase object that can be translated before it is output.

What do you think, @samdark?

samdark: Yes. That makes sense. Enough flexibility for i18n while not covering it explicitly.

LeTraceurSnork: `getMessage()` now returns `string|\Stringable`

proposed/validator.md
LeTraceurSnork: added `array $context = []` as second param

proposed/validator-meta.md
LeTraceurSnork: What about really simple Validators, like, `EmailValidator`, `PhoneValidator`, `NotEmptyValidator` ?

proposed/validator-meta.md
LeTraceurSnork: I suggest to resolve this thread in the very end - when we'll handle everything else and then we can concentrate everybody on this

LeTraceurSnork:
@Crell @samdark could you resolve all of the obsolete threads so we would be able to return to review this one again?

Crell:
Code review is premature when there's no working group.  Do that first.

PR #1339, opened 2025-07-26
[PSR-5] enhancements

jaapio:
This pull request introduces enhancements to the PHPDoc specification (`proposed/phpdoc.md`) by refining tag handling and adding support for inline tags. The most important changes include updates to the ABNF definition for tags, the introduction of inline tags, and the addition of tag specialization for improved metadata organization.

### Enhancements to tag handling:

* Updated the ABNF definition to modify the syntax for `tag-specialization` and `tag`:
  - `tag-specialization` now excludes the `-` character, and `tag` syntax includes an optional specialization prefix followed by a hyphen.

* Added the concept of tag specialization (`5.3.1 Tag Specialization`) to provide a scoped context for tags. Specialized tags can differ from the standard catalog, and parsers may ignore unsupported specialized tags.

### Support for inline tags:

* Introduced inline tags (`5.2.1 Inline tags`) within descriptions. Inline tags must start with `{@` and end with `}`. Escaping is supported for closing braces using `{}`. Inline tags are limited to regular tags and do not apply to annotation tags.

### Structural refinements:

* Updated the table of contents to reflect the addition of inline tags and the reorganization of tag-related sections.

### DIFF
--- a/proposed/phpdoc.md
+++ b/proposed/phpdoc.md
@@ -10,6 +10,7 @@ PSR-5: PHPDoc
 - [5. The PHPDoc Format](#5-the-phpdoc-format)
   - [5.1. Summary](#51-summary)
   - [5.2. Description](#52-description)
+    - [5.2.1 Inline tags](#521-inline-tags)
   - [5.3. Tags](#53-tags)
     - [5.3.1. Tag Name](#531-tag-name)
   - [5.4. Examples](#54-examples)
@@ -237,6 +238,33 @@ Common uses for the Description:
 * To provide a set of common use cases or scenarios in which the "Structural
   Element" may be applied
 
+#### 5.2.1 Inline tags
+
+A description MAY contain inline tags. Inline tags MUST follow the specification of
+tags with the same syntax. An inline tag MUST start with `{@` and MUST with `}`
+
+```php
+  /** 
+   * Summary.
+   * 
+   * This is a description with an inline tag {@tag rest of the tag format}
+   */
+```
+
+Inline tags are only covering regular tags and SHALL NOT cover _Annotation_ tags.
+
+As inline tags are always closed with a `}` developers cannot use this char in a 
+description. To overcome this issue parsers MUST support escape char `{` followed
+by `}` to be interpreted as a single `}` 
+
+```php
+   /**
+    * Summary.
+    *
+    * This is a description with an inline tag {@tag show case {} escape of ending char}
+    */
+```
+
 ### 5.3. Tags
 
 Tags supply concise metadata for a "Structural Element". Each tag starts on a
@@ -253,8 +281,8 @@ format, as dictated by the specific tag.
 > _type_ (`string`), variable name (`$argument1`),  and description (`This is a
 > parameter.`).
 
-The description MUST support Markdown as a formatting language.  The
-description of the tag MAY start on the same line or next line.  The following
+The description MUST support Markdown as a formatting language. The
+description of the tag MAY start on the same line or next line. The following
 tags are semantically identical:
 
 ```php
--- a/proposed/phpdoc.md
+++ b/proposed/phpdoc.md
@@ -12,7 +12,8 @@ PSR-5: PHPDoc
   - [5.2. Description](#52-description)
     - [5.2.1 Inline tags](#521-inline-tags)
   - [5.3. Tags](#53-tags)
-    - [5.3.1. Tag Name](#531-tag-name)
+    - [5.3.1. Tag Specialization](#531-tag-specialization) 
+    - [5.3.2. Tag Name](#532-tag-name)
   - [5.4. Examples](#54-examples)
 - [Appendix A. Types](#appendix-a-types)
   - [ABNF](#abnf)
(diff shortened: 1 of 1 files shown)


proposed/phpdoc.md
```diff
@@ -43,14 +41,15 @@ interpreted as described in [RFC 2119][RFC2119].
   preceded by a DocBlock. The collection contains the following constructs:
 
   * `require` / `include` (and their `\_once` variants)
-  * `class` / `interface` / `trait`
+  * `class` / `interface` / `trait` / `enum`
```
Crell: There's a non-zero chance there may be more "class likes" in the future.  Rather than a fixed list, perhaps say "class-like constructs, including class, interface, trait, enum, and others later added."  Or something like that.

proposed/phpdoc.md
```diff
@@ -191,15 +156,15 @@ interpreted as described in [RFC 2119][RFC2119].
 The PHPDoc format has the following [ABNF][RFC5234] definition:
 
     PHPDoc             = [summary [description]] [tags]
-    eol                = [CR] LF ; to compatible with PSR-12
+    eol                = [CR] LF ; to be compatible with PSR-12
```
Crell: PER-CS?

proposed/phpdoc.md
```diff
@@ -208,7 +173,7 @@ Examples of use are included in chapter 5.4.
 
 ### 5.1. Summary
 
-A Summary MUST contain an abstract of the "Structural Element" defining the
+A Summary MUST provide an abstract of the "Structural Element" defining the
 purpose. It is RECOMMENDED for Summaries to span a single line or two, but not
```
Crell: "A single line or two" is very squishy language.  Should we just require a single line?  PER-CS gives you up to 120 characters...

proposed/phpdoc.md
```diff
@@ -237,11 +201,36 @@ Common uses for the Description:
...
+   * Summary.
+   * 
+   * This is a description with an inline tag {@tag rest of the tag format}
+   */
+```
+
+Inline tags MAY only represent regular tags and SHALL NOT represent _Annotation_ tags.
+
+As inline tags are always closed with a `}` developers cannot use this char in a 
+description. To overcome this issue parsers MUST support escape sequence of `{` followed by `}` MUST be interpreted as a literal closing brace.
```
Crell, line 242: That's a pretty weird escape character, no?  Why not the more common `\`?

proposed/phpdoc.md
```diff
@@ -253,8 +242,8 @@ format, as dictated by the specific tag.
 > _type_ (`string`), variable name (`$argument1`),  and description (`This is a
 > parameter.`).
 
-The description MUST support Markdown as a formatting language.  The
-description of the tag MAY start on the same line or next line.  The following
+The description MUST support Markdown as a formatting language. The
+description of the tag MAY start on the same line or next line. The following
```
Crell, line 268: Does this implicitly mean supporting HTML, too?

Crell: Also... which Markdown? :-)  There are many flavors.

proposed/phpdoc.md
jaapio: it is, but this is how the current tooling implements this. I think we should not change that right now. Unless people have a different point of view on this.

ashnazg: That's purely historical... that character usage goes back to the early days of phpDocumentor 1.x, so it's probably too entrenched to arbitrarily change at this point.

proposed/phpdoc.md
jaapio: I had the same thought.  Doesn't that make the standard loosely defined? Because we do not know what will be added in the future. So I struggled with this. And I thought it would better to be explicit rather than allowing all. Adding more when the language evolves is possible with a change on the PSR as long it's not breaking? 

I'm all ok with using class-like constructs. As php allows docblocks on any level.

proposed/phpdoc.md
jaapio: This is again based on the existing implementations of the tools reading docblocks. I took phpDocumentor as a standard here, since all tools are following. I'm not trying to set a new standard here, as the docblocks are already like this for many years. 

But I get your point. We need to sharpen the definition here. 

The full definition of a summary is complex. :-( 

https://github.com/phpDocumentor/ReflectionDocBlock/blob/5.x/src/DocBlockFactory.php#L226C1-L238C14
```
      # 2. Extract the summary
      (?:
        (?! @\pL ) # The summary may not start with an @
        (
          [^\n.]+
          (?:
            (?! \. \n | \n{2} )     # End summary upon a dot followed by newline or two newlines
            [\n.]* (?! [ \t]* @\pL ) # End summary when an @ is found as first character on a new line
            [^\n.]+                 # Include anything else
          )*
          \.?
        )?
      )
```

proposed/phpdoc.md
jaapio: Maybe we just have to pick CommonMark as that's the most common standard I know in markdown. The set is limited, but as far as I know, most tools support the syntax of CommonMark. 

I need to do some more research on how other php based tools are doing this, but mostly it's just phpDocumentor doing this. I didn't find any other libraries that handle the description of docblocks in the wild. Apart from the IDE's. We would need some people from those sides to have a look at this.

proposed/phpdoc.md
```diff
@@ -237,11 +201,36 @@ Common uses for the Description:
 * To provide a set of common use cases or scenarios in which the "Structural
   Element" may be applied
 
+#### 5.2.1 Inline tags
+
+A description MAY contain inline tags. Inline tags MUST follow the specification of
+tags with the same syntax. An inline tag MUST start with `{@` and MUST with `}`
```
drupol:
```suggestion
tags with the same syntax. An inline tag MUST start with `{@` and MUST end with `}`
```

proposed/phpdoc.md
samdark: https://github.com/yiisoft/yii2-apidoc, https://www.yiiframework.com/doc/api/2.0 are generated using it.

samdark (approved):

akramwahid:
<script>alert('ggg')</script>

proposed/phpdoc.md
jaapio: I updated the wording a bit. To make it more explicit. 

> A Summary MUST be a concise abstract of the Structural Element’s purpose. It is RECOMMENDED to keep the Summary to a single line, though two lines MAY be used if needed.

PR #1346, opened 2026-01-14
Add errata clarifying `Exception` vs. `Throwable` usage in PSR-3

Aliance:
PSR-3 was written before PHP introduced the `Throwable` interface. At that time, `Exception` was the only error-related base type available.

In modern PHP versions, both `Exception` and `Error` implement `Throwable`, and it is common practice for logging libraries to accept any `Throwable` when logging failures.

This PR adds an Errata section to PSR-3 clarifying that references to `Exception` in the specification - specifically the `exception` context key - should be interpreted as allowing any `Throwable` instance.

The main specification text is intentionally left unchanged. This errata is meant to:

- reflect modern PHP language semantics;
- improve clarity for implementors;
- increase awareness without introducing breaking changes.

Per FIG process, this change is intended to go through the errata vote.

@Seldaek please, take a look once again if possible

Related pr: https://github.com/php-fig/fig-standards/pull/1345


samdark (approved):

Seldaek (approved):
LGTM, thanks

mbniebergall (approved):
Voting passed, merging PR.

